PR #5, opened 2018-12-10
[no ticket] Add Prettier configuration

fcoronelmakingsense:
- Add Prettier to format code formatter (https://prettier.io/)
- Add Prettier settings https://prettier.io/docs/en/configuration.html
- Add Prettier in build to correct files

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,16 +5,25 @@
   "dependencies": {
     "eclint": "^2.8.1",
     "g": "^2.0.1",
+    "prettier-check": "^2.0.0",
     "react": "^16.6.3",
     "react-dom": "^16.6.3",
     "react-scripts": "2.1.1"
   },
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -11,10 +11,17 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build && yarn eclint && yarn test",
+    "build": "yarn prettier && react-scripts build && yarn eclint && yarn test",
```
andresmoschini: I think that build should not modify files, it should verify that the files are pretty already.

`prettier-check` package is useful for that. See https://github.com/MakingSense/doppler-feature-toggle-js/blob/master/package.json#L24

cbernat:
Great!

PR #7, opened 2018-12-12
[no ticket] Add service method to test cors in doppler

fcoronelmakingsense:
- Add service test method to test cors in doppler app
- Obtain user data from doppler and show in screen

### DIFF
--- a/src/components/Header/Nav.js
+++ b/src/components/Header/Nav.js
@@ -3,11 +3,31 @@ import React, { Component } from 'react';
 import './Nav.css';
 
 export default class HeaderNav extends Component {
+  constructor() {
+    super();
+    this.state = {
+      data: {},
+    };
+  }
+
+  componentWillMount() {
+    fetch('http://localhost:52191/Reports/Reports/GetUserData', {
+      mode: 'cors',
+      credentials: 'include',
+    })
+      .then((response) => {
+        return response.json();
+      })
+      .then((data) => {
+        this.setState({ data: data.user });
+      });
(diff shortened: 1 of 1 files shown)


cbernat:
👍

src/components/Header/Nav.js
```diff
@@ -3,11 +3,31 @@ import React, { Component } from 'react';
...
 export default class HeaderNav extends Component {
+  constructor() {
+    super();
+    this.state = {
+      data: {},
+    };
+  }
+
+  componentWillMount() {
+    fetch('http://localhost:52191/Reports/Reports/GetUserData', {
```
andresmoschini: Should it be in the following way?

```suggestion
    return fetch('http://localhost:52191/Reports/Reports/GetUserData', {
```

andresmoschini: I think that the related URL should not be _Reports_ but _WebApp_

src/components/Header/Nav.js
```diff
@@ -3,11 +3,31 @@ import React, { Component } from 'react';
...
+      mode: 'cors',
+      credentials: 'include',
+    })
+      .then((response) => {
+        return response.json();
+      })
+      .then((data) => {
+        this.setState({ data: data.user });
+      });
```
andresmoschini: Only curiosity, why did you use two _then_ in place of only one?

```js
      .then((response) => {
        const data = response.json();
        this.setState({ data: data.user });
      });
```

PR #9, opened 2019-01-03
[INTERNATIONALIZATION] Add react-translated

lucianaDilizio:
- Add react-translated package.

This can be changed. I want to know your opinion. This is the easiest package for translations that I found. If you know better options let me know please 😉 

Source:
https://www.npmjs.com/package/react-translated#templated-text

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import logo from './logo.svg';
 import './App.css';
+import {Translate} from 'react-translated';
 
 import HeaderNav from './components/Header/Nav';
 
@@ -12,7 +13,7 @@ class App extends Component {
           <HeaderNav />
           <img src={logo} className="App-logo" alt="logo" />
           <p>
-            Edit <code>src/App.js</code> and save to reload.
+            <Translate text='Edit' /> <code>src/App.js</code> <Translate text='and save to reload.' />
           </p>
           <a
             className="App-link"
@@ -20,7 +21,7 @@ class App extends Component {
             target="_blank"
             rel="noopener noreferrer"
           >
-            Learn React
+            <Translate text='Learn React' />
           </a>
         </header>
       </div>
--- a/src/locales/translation.js
+++ b/src/locales/translation.js
@@ -1,0 +1,14 @@
+export default {
+    'Edit': {
+     	en: 'Edit',
+     	es: 'Editar'
+    },
+    'and save to reload.': {
+    	en: 'and save to reload.',
+     	es: 'y guarda para recargar el sitio.'
+    },
+    'Learn React': {
+        en: 'Learn React',
+        es: 'Aprende React'
+      }
+  }
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "prettier-check": "^2.0.0",
     "react": "^16.6.3",
     "react-dom": "^16.6.3",
-    "react-scripts": "2.1.1"
+    "react-scripts": "2.1.1",
+    "react-translated": "^2.2.4"
   },
   "scripts": {
     "start": "react-scripts start",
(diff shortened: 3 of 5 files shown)


src/locales/translation.js
```diff
@@ -0,0 +1,14 @@
+export default {
+    'Edit': {
+     	en: 'Edit',
+     	es: 'Editar'
+    },
+    'and save to reload.': {
```
fcoronelmakingsense: maybe is better use keys without spaces like `And_Save_To_Reload` or `and_save_to_reload`

lucianaDilizio: If you check the link in the source, the examples are like this. We can use it in this way if is better, of course.

src/App.js
```diff
@@ -12,15 +13,15 @@ class App extends Component {
           <HeaderNav />
           <img src={logo} className="App-logo" alt="logo" />
           <p>
-            Edit <code>src/App.js</code> and save to reload.
+            <Translate text='Edit' /> <code>src/App.js</code> <Translate text='and save to reload.' />
```
andresmoschini: There should be a better way to translate text with variable content, something like:

<Translate text='Edit {content}'><code>src/App.js</code></Translate>

src/locales/translation.js
andresmoschini: I understand your point @fcoronelmakingsense but maybe not, I would use the suggested way proposed by library owners.

andresmoschini:
Great initiative @lucianaDilizio !!

> This can be changed. I want to know your opinion. This is the easiest package for translations that I found. If you know better options let me know please 😉

I think that it is an important think to take into account, and also how popular is it, did you research about that? Do you have some numbers to compare this library with other options? Also, in MakingSense we are already using React, did you ask what library are we using here? (maybe we would decide for someone different than they, but at least, with more information)

src/App.js
lucianaDilizio: 🤔

andresmoschini: Probably it is the right way:

https://www.npmjs.com/package/react-translated#styled-text

```
<Translate text='Edit `src/App.js` and save to reload.' />
```

and if you want to make it customizable:

```
<Translate
    text='Edit `{filename}` and save to reload.'
    data={{ filename: 'src/App.js' }}
/>
```

andresmoschini:
(I am only asking, not expressing an opinion)

Should we test it in any way?

lucianaDilizio:
Thanks @andresmoschini! I was looking some options and this was the most simple that I found. Maybe is not too popular but it works fine. 
I will ask to my coworkers which library we are using here, I forgot that in servicies there are some projects that are using React. (Good point 😉)
I didn't create this PR to merge specifically. I just want to know your opinions and suggestions to find the best option.

andresmoschini:
> I didn't create this PR to merge specifically. I just want to know your opinions and suggestions to find the best option.

Sure :), I am not criticizing, only asking for a little research, [npm](https://www.npmjs.com) is very useful for that. 

My concern is mainly because it could work great now but it could be not enough for us in the future, or it could even lose maintenance.

lucianaDilizio:
Good point @andresmoschini! You're right. I will looking for something better so. Thanks!

cbernat:
We've talked this with @lucianaDilizio and she will be looking into other alternatives before merging this.

andresmoschini:
> We've talked this with @lucianaDilizio and she will be looking into other alternatives before merging this.

Maybe she could create another PR to compare.

cbernat:
> > We've talked this with @lucianaDilizio and she will be looking into other alternatives before merging this.
> 
> Maybe she could create another PR to compare.

Yes, that would be great too. I've seen a lot of people use `react-intl` maybe try that one.

andresmoschini:
Closed in favor of #13

PR #10, opened 2019-01-03
fix(docker): add hot reload to watch

cbernat:
Add hot reload to change the files and see it reflected in localhost. For local development using docker.


andresmoschini:
:+1:

PR #13, opened 2019-01-04
feat: add react-intl package

lucianaDilizio:
At the request of the audience 🥁 
- Add React Intl package

Documentation:
https://github.com/yahoo/react-intl/wiki

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,8 +1,14 @@
 import React, { Component } from 'react';
 import logo from './logo.svg';
 import './App.css';
+import { FormattedHTMLMessage } from 'react-intl';
+import { addLocaleData } from 'react-intl';
+import en from 'react-intl/locale-data/en';
+import es from 'react-intl/locale-data/es';
 
 import HeaderNav from './components/Header/Nav';
+
+addLocaleData([...en, ...es]);
 
 class App extends Component {
   render() {
@@ -12,7 +18,7 @@ class App extends Component {
           <HeaderNav />
           <img src={logo} className="App-logo" alt="logo" />
           <p>
-            Edit <code>src/App.js</code> and save to reload.
+            <FormattedHTMLMessage id="app.title" />
           </p>
           <a
             className="App-link"
@@ -20,7 +26,7 @@ class App extends Component {
             target="_blank"
             rel="noopener noreferrer"
           >
-            Learn React
+            <FormattedHTMLMessage id="app.link" />
           </a>
         </header>
       </div>
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -1,9 +1,24 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
+import { IntlProvider } from 'react-intl';
+import messages_es from './i18n/es.json';
+import messages_en from './i18n/en.json';
+
+const messages = {
+  es: messages_es,
+  en: messages_en,
+};
+
+const language = 'en';
 
 it('renders without crashing', () => {
   const div = document.createElement('div');
-  ReactDOM.render(<App />, div);
+  ReactDOM.render(
+    <IntlProvider locale={language} messages={messages[language]}>
+      <App />
+    </IntlProvider>,
+    div,
+  );
   ReactDOM.unmountComponentAtNode(div);
 });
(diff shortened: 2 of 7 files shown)


lucianaDilizio:
test this please

lucianaDilizio:
test this please

lucianaDilizio:
test this please

cbernat:
Test this please.

lucianaDilizio:
Test this please

cbernat:
Test this please.

cbernat:
Test this please

cbernat:
Test this please

cbernat:
Test this please

src/index.js
```diff
@@ -3,8 +3,23 @@ import ReactDOM from 'react-dom';
...
+import messages_es from './i18n/es.json';
+import messages_en from './i18n/en.json';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+const messages = {
+  es: messages_es,
+  en: messages_en,
+};
+
+const language = 'es'; //TODO: this hardcoded data will depend by the app language
```
cbernat: Why don't we make it browser dependable for now? :) What do you think?

lucianaDilizio: Yes, I'm agree with you. I will apply this.

src/index.js
```diff
@@ -3,8 +3,30 @@ import ReactDOM from 'react-dom';
...
 
-ReactDOM.render(<App />, document.getElementById('root'));
+const messages = {
+  es: messages_es,
+  en: messages_en,
+};
+
+//TODO: this hardcoded data will depend by the app language
+const language = (
+  (navigator.languages && navigator.languages[0]) ||
```
cbernat: I don't think you need all this options, I've check here https://www.w3schools.com/jsref/prop_nav_language.asp and with `navigator.language` and `navigator.browserLanguage` we are covered.

src/App.js
```diff
@@ -1,9 +1,15 @@
 import React, { Component } from 'react';
 import logo from './logo.svg';
 import './App.css';
+import { FormattedHTMLMessage } from 'react-intl';
+import { addLocaleData } from 'react-intl';
```
andresmoschini: Why not using these both lineas in only one?

```js
import { FormattedHTMLMessage, addLocaleData } from 'react-intl';
```

Is it a desired coding style? or it is only something casual?

src/App.test.js
```diff
@@ -1,9 +1,24 @@
...
+  en: messages_en,
+};
+
+const language = 'en';
 
 it('renders without crashing', () => {
   const div = document.createElement('div');
-  ReactDOM.render(<App />, div);
+  ReactDOM.render(
+    <IntlProvider locale={language} messages={messages[language]}>
```
andresmoschini: What is the meaning of this element?

src/App.test.js
```diff
@@ -1,9 +1,24 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
+import { IntlProvider } from 'react-intl';
+import messages_es from './i18n/es.json';
+import messages_en from './i18n/en.json';
+
+const messages = {
+  es: messages_es,
+  en: messages_en,
```
andresmoschini: Why are you importing all languages if then there is a constant to use only `en`?

src/App.js
lucianaDilizio: Casual 😛

src/App.test.js
lucianaDilizio: Yes, you're right, in the test we can keep the one that we use.

src/index.js
```diff
@@ -3,8 +3,24 @@ import ReactDOM from 'react-dom';
...
+import messages_en from './i18n/en.json';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+const messages = {
+  es: messages_es,
+  en: messages_en,
+};
+
+//TODO: this hardcoded data will depend by the app language
+const language = navigator.language.toLowerCase().split(/[_-]+/)[0] || 'en';
```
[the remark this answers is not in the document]
lucianaDilizio: Yes,  it could be a possibility. I didn't think of that. Maybe we can also use the default messages that the library provides when it doesn't find the language.

PR #15, opened 2019-01-10
[No ticket] Add environments variables

fcoronelmakingsense:
- Add `.env` for build variables
- Add `.env.development` for develop variables

based on https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables

Then maybe need for int or qa.

- Add redirect to doppler login page when can't get user data

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+REACT_APP_API_URL=https://app2.fromdoppler.com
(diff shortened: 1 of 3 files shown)


fcoronelmakingsense:
test this please

.env
```diff
@@ -0,0 +1 @@
+REACT_APP_API_URL=https://app2.fromdoppler.com
```
andresmoschini: Why not calling it with a more descriptive name, like _Doppler_MVC_URL_ or something like that?

PR #20, opened 2019-01-23
keep session alive

fcoronelmakingsense:
- Add  setinterval every 60 seconds to call getUserdata method to get user information data and keep session alive in doppler

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -22,6 +22,13 @@ class App extends Component {
 
   componentWillMount() {
     this.getUserData();
+    this.interval = setInterval(() => {
+      this.getUserData();
+    }, 60000);
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.interval);
   }
 
   manageJwtToken() {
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -21,7 +21,13 @@ class App extends Component {
   }
 
   componentWillMount() {
-    this.getUserData();
+    this.interval = setInterval(() => {
+      this.getUserData();
```
andresmoschini: Should you verify if the user returned by doppler is the same that the user of the JWT token?

fcoronelmakingsense:
> Is it possible adding any test?

I need research for test in general for react

andresmoschini:
> > Is it possible adding any test?
> 
> I need research for test in general for react

The sooner the better.

src/App.js
fcoronelmakingsense: i added now

andresmoschini: What to do if it is different? we should make honor to _Doppler MVC_ response or _JWT_ token?

PR #24, opened 2019-02-05
[No ticket] Add more functionality in testing

fcoronelmakingsense:
- Add local storage mocks to test happy path
- Add jwt decode mock to works decode jwt token functionality
- Change fetch request to axios token

### DIFF
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -8,6 +8,10 @@ import App from './App';
 import { IntlProvider } from 'react-intl';
 import messages_es from './i18n/es.json';
 import messages_en from './i18n/en.json';
+import jwt_decode from 'jwt-decode';
+
+//Add mock to decode jwt token and not fail
+jest.mock('jwt-decode');
 
 const messages = {
   es: messages_es,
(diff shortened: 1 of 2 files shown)


src/App.test.js
```diff
@@ -22,11 +26,31 @@ const response = {
...
+};
+global.localStorage = localStorageMock;
+
 describe('App component', () => {
   afterEach(cleanup);
 
   beforeEach(() => {
-    axios.get = jest.fn(() => Promise.resolve(response));
+    axios.get = jest.fn((url) => {
+      if(url===(process.env.REACT_APP_API_URL + '/Reports/Reports/GetUserData')) {
```
cbernat: Could you fix the spacing between `===` and other symbols here?

src/App.test.js
```diff
@@ -50,5 +83,10 @@ describe('App component', () => {
     const userEmail = getByText(response.data.user.Email);
 
     expect(userEmail).toBeDefined();
+    expect(axios.get).toHaveBeenCalledTimes(1);
+    expect(axios.get).toHaveBeenCalledWith(process.env.REACT_APP_API_URL + '/Reports/Reports/GetUserData', {
```
cbernat: Maybe this `process.env.REACT_APP_API_URL + '/Reports/Reports/GetUserData' ` can be defined in a variable to reuse.

src/App.test.js
andresmoschini: Does not Prettify or JSLint detect it?

cbernat: Seems he did not run it I suspect...

src/App.test.js
```diff
@@ -8,6 +8,10 @@ import App from './App';
 import { IntlProvider } from 'react-intl';
 import messages_es from './i18n/es.json';
 import messages_en from './i18n/en.json';
+import jwt_decode from 'jwt-decode';
+
+//Add mock to decode jwt token and not fail
```
andresmoschini: Very arguably, but I think that these comments are not necessary. What do you think?

fcoronelmakingsense: yes could be

src/App.test.js
fcoronelmakingsense: i forgot run. now i run and fixed

andresmoschini: There are Prettify plugins for most text editors, I do not like them, but maybe you do.

PR #26, opened 2019-02-08
Fix index.html behavior in any CDN folder

andresmoschini:
Issue:

issue

Maybe this approach is not the best since it limits us from some routing features, see https://facebook.github.io/create-react-app/docs/deployment#serving-the-same-build-from-different-paths

> If you are not using the HTML5 pushState history API or not using client-side routing at all, it is unnecessary to specify the URL from which your app will be served.

But, by the moment, it is fine. In the future, we could update _homepage_ property as part of the build process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "react-scripts": "2.1.1",
     "react-testing-library": "^5.4.4"
   },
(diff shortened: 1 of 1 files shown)


andresmoschini:
@GusBaamonde reviewed it and ask me for merging.

PR #29, opened 2019-02-13
[WIP] Header nav initial structure

fcoronelmakingsense:
- Create this pr to discuss is correctly implemented the navigation. In doppler the navigation its obtain from BE. in this case maybe it's not necessary and only need account type to show or not menu options. Now not implemented this yet.

- Add validation in App.js is user logged in or not to not render the elements until check user logged correctly. I put `Loading...` temporaly until defined what show here

image

- That's what it looks like now without styles


src/components/Header/Header.js
```diff
@@ -1,12 +1,31 @@
...
+
+  constructor() {
+    super();
+    this.state = {
+      user: {},
+    };
+  }
+
+  componentWillMount() {
+    this.setState({ user: this.props.user })
```
andresmoschini: If I am not wrong, when something is not changed by the component, it should be dealt as _props_ and not as _state_, what do you think?

src/components/Header/Header.js
```diff
@@ -1,12 +1,31 @@
...
+        <HeaderNav />
+        <HeaderUserMenu user={this.state.user} />
+      </header>
+    );
+  }
 
-const Header = (props) => {
-  return (
-    <header>
-      <HeaderNav user={props.user} />
```
andresmoschini: I think that this line was ok.

src/components/Header/Header.js
fcoronelmakingsense: Yes! this it's not final. Need the real data and pass only what child needs. I need change this, i was trying things

src/components/Header/Header.js
```diff
@@ -1,12 +1,31 @@
-import React from 'react';
+import React, { Component } from 'react';
 import HeaderNav from './HeaderNav/HeaderNav';
+import HeaderUserMenu from './HeaderUserMenu/HeaderUserMenu';
+import { ReactComponent as DopplerLogo } from '../../doppler-logo.svg';
+
+class Header extends Component {
+
+  constructor() {
+    super();
```
andresmoschini: if I am not wrong, it is a good practice using _props_ parameter, something like:

```
  constructor(props) {
    super(props);
```

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
```diff
@@ -0,0 +1,18 @@
...
+      <div>
+        <span>user avatar text</span>
+        <div>
+          <span>{this.props.user.Email}</span>
+        </div>
+      </div>
+    );
+  }
+
+};
```
andresmoschini:
```suggestion
}
```

This curly brace is related to the class definition, right?

src/components/Header/Header.js
fcoronelmakingsense: I have to research what user properties I need for the headernav yet

src/components/Header/Header.js
fcoronelmakingsense: i didn't know. good!

src/components/Header/Header.js
```diff
@@ -1,12 +1,31 @@
-import React from 'react';
+import React, { Component } from 'react';
 import HeaderNav from './HeaderNav/HeaderNav';
+import HeaderUserMenu from './HeaderUserMenu/HeaderUserMenu';
+import { ReactComponent as DopplerLogo } from '../../doppler-logo.svg';
```
andresmoschini: Nice!

src/components/Header/Header.js
```diff
@@ -1,12 +1,31 @@
...
+
+class Header extends Component {
```
andresmoschini: If you are already importing global object (_React_) why not using this in place importing another name:

```suggestion
// import React from 'react';
class Header extends React.Component {
```

Another option is not importing global React, for example:


```
import { Component, AnotherThing, AnotherMore } from 'react';
```

src/components/Header/Header.js
andresmoschini: Great!

src/components/Header/Header.js
fcoronelmakingsense: i didn't know that. thanks! good catch! Which you prefeer?

andresmoschini: Mmm, I think that _Nicholas Zakas_ prefers the last one, let me research...

andresmoschini: Well he talks about exports, but, I think that the same applies for imports: https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/

So, what do you think about **avoiding import defaults when it is possible**?

src/components/Header/HeaderNav/HeaderNav.js
```diff
@@ -1,10 +1,85 @@
...
+import { FormattedMessage } from 'react-intl';
 
-const HeaderNav = (props) => {
+
+const HeaderNav = () => {
   return (
     <nav>
       <ul>
-        <li>{props.user.Email}</li>
+        <li>
```
lucianaDilizio: Maybe this could be an iteration, what do you think? To avoid repeat html code. Maybe this information is neccesary to be obtained by the BE so in this case I was thinking that may be we could use a map or somethink like that.

andresmoschini: Se puede hacer una iteración con react, incluso hardcodeando los datos de entrada, igual no se si vale la pena.

fcoronelmakingsense: i created this pr to discuss that. i talked with @cbernat and maybe is better this is static and not dinamyc. but can changed. What do you think @andresmoschini ?

src/components/Header/Header.js
andresmoschini: 👁 I am not sure, please research before paying attention to me opinion.

src/components/Header/HeaderNav/HeaderNav.js
cbernat: I didn't get exactly what you were talking about, but in first instance maybe is better to consume data as it is in doppler. And then we will see, how can we improve from there.

andresmoschini: > Maybe this information is neccesary to be obtained by the BE

I did not know that you had so rich object in Doppler, _mainMenuData_, I understand now, and it makes sense absolutely.

Thanks @lucianaDilizio!

src/components/Header/Header.js
cbernat: Why not make this component as a function? For les overhead `class Header extends React.Component {`

fcoronelmakingsense:
I update this pr

src/i18n/en.json
```diff
@@ -1,4 +1,20 @@
 {
+  "header_nav": {
```
cbernat: I think we don't need to have the menu translation here, right? Since we have it in the json model

fcoronelmakingsense: yes good catch! thanks!

src/App.js
```diff
@@ -92,13 +92,21 @@ class App extends Component {
...
 
   render() {
-    return (
-      <div>
-        <Header user={this.state.user} />
-        <img src={logo} alt="logo" />
-        <Footer />
-      </div>
-    );
+    const isLoggedIn = Object.keys(this.state.user).length;
```
andresmoschini: I prefer something like this:

```suggestion
    const isLoggedIn = !!this.state.user;
```

In my opinion, having user = null when the user is not authenticated, is semantically more correct.

fcoronelmakingsense: thanks!

src/components/Header/Header.js
```diff
@@ -1,12 +1,30 @@
...
-    <header>
-      <HeaderNav user={props.user} />
-    </header>
-  );
-};
+class Header extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      user: {},
```
andresmoschini: I think that it should be

```suggestion
      user: null,
```

I mean, when the _user_ is not defined, _null_ is a better representation than an empty object.

Also, empty object will not work with `const isLoggedIn = !!this.state.user;`

src/components/Header/HeaderNav/HeaderNav.js
```diff
@@ -1,10 +1,28 @@
...
+        {navItems.map((item, index) => (
+          <li key={index}>
+            <a className={item.isSelected ? 'active' : ''} href={item.url}>
+              {item.title}
+            </a>
+            {item.subNav ? (
+              <ul>
+                {item.subNav.map((itemSubNav, index) => (
+                  <li key={index}>
+                    <a className={itemSubNav.isSelected ? 'active' : ''} href={itemSubNav.url}>
```
andresmoschini: Las comillas hay que ponerlas o JSX las pone solo?

```suggestion
                    <a className={itemSubNav.isSelected ? 'active' : ''} href='{itemSubNav.url}'>
```

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
```diff
@@ -0,0 +1,16 @@
+import React, { Component } from 'react';
+
+class HeaderUserMenu extends Component {
```
andresmoschini: What do you think about simply using a function in place of a component?

src/headerData.json
```diff
@@ -0,0 +1,9 @@
...
+    {"title":"Automation","url":"https://app2.fromdoppler.com/Automation/Automation/AutomationApp/","isEnabled":false,"isSelected":false,"subNav":[],"idHTML":"automationMenu","$$hashKey":"object:8"},
+    {"title":"Reportes","url":"https://app2.fromdoppler.com/Campaigns/Reports/","isEnabled":false,"isSelected":false,"subNav":[],"idHTML":"reportMenu","$$hashKey":"object:9"},
+    {"title":"Plantillas","url":"https://app2.fromdoppler.com/Templates/Main/","isEnabled":false,"isSelected":true,"idHTML":"templateMenu","$$hashKey":"object:10"}
+  ]
+}
```
andresmoschini: What do you think about prettifying this json?

src/components/Header/HeaderNav/HeaderNav.js
fcoronelmakingsense: si pones las comillas te lo toma como un string

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
fcoronelmakingsense: it's depends if need or not state. I not implemented yet this 💃

src/headerData.json
fcoronelmakingsense: it's posible?

src/components/Header/Header.js
fcoronelmakingsense: yes you right

src/components/Header/HeaderNav/HeaderNav.js
andresmoschini: Joya, gracias por la aclaración, tenemos que aprender mucho de esta notación.

src/headerData.json
andresmoschini: Sure! I just did it using [this tool](https://jsonformatter.curiousconcept.com/), there are other tools, or you also could do it manually.

```json
{
  "nav":[
    {
      "title":"Campañas",
      "url":"https://app2.fromdoppler.com/Campaigns/Draft/",
      "isEnabled":false,
      "isSelected":false,
      "subNav":[
        {
          "title":"Borradores",
          "url":"https://app2.fromdoppler.com/Campaigns/Draft/",
          "isEnabled":false,
          "isSelected":false,
          "idHTML":"campaignDraftMenu",
          "$$hashKey":"object:32"
        },
        {
          "title":"Programadas",
          "url":"https://app2.fromdoppler.com/Campaigns/Scheduled/",
          "isEnabled":false,
          "isSelected":false,
          "idHTML":"campaignScheduledMenu",
          "$$hashKey":"object:33"
        },
        {
          "title":"Enviadas",
          "url":"https://app2.fromdoppler.com/Campaigns/Sent/",
          "isEnabled":false,
          "isSelected":false,
          "idHTML":"campaignSentMenu",
          "$$hashKey":"object:34"
        },
        {
          "title":"Test A/B",
          "url":"https://app2.fromdoppler.com/Campaigns/TestAB/",
          "isEnabled":false,
          "isSelected":false,
          "idHTML":"campaignTestABMenu",
          "$$hashKey":"object:35"
        }
      ],
      "idHTML":"campaignMenu",
      "$$hashKey":"object:6"
    },
    {
      "title":"Listas",
      "url":"https://app2.fromdoppler.com/Lists/SubscribersList",
      "isEnabled":false,
      "isSelected":false,
      "subNav":[
        {
          "title":"Listas Principales",
          "url":"https://app2.fromdoppler.com/Lists/SubscribersList/",
          "isEnabled":false,
          "isSelected":false,
          "idHTML":"listMainMenu",
          "$$hashKey":"object:41"
        },
        {
          "title":"Segmentos",
          "url":"https://app2.fromdoppler.com/Lists/Segment/",
          "isEnabled":false,
          "isSelected":false,
          "idHTML":"listSegmentMenu",
          "$$hashKey":"object:42"
        },
        {
          "title":"Suscriptores",
          "url":"https://app2.fromdoppler.com/Lists/MasterSubscriber/",
          "isEnabled":false,
          "isSelected":false,
          "idHTML":"listMasterSubscriberMenu",
          "$$hashKey":"object:43"
        },
        {
          "title":"Campos Personalizados",
          "url":"https://app2.fromdoppler.com/Lists/MasterCustomFields/",
          "isEnabled":false,
          "isSelected":false,
          "idHTML":"listCustomFieldMenu",
          "$$hashKey":"object:44"
        },
        {
          "title":"Formularios",
          "url":"https://app2.fromdoppler.com/Lists/Form/",
          "isEnabled":false,
          "isSelected":false,
          "idHTML":"listFormMenu",
          "$$hashKey":"object:45"
        }
      ],
      "idHTML":"listMenu",
      "$$hashKey":"object:7"
    },
    {
      "title":"Automation",
      "url":"https://app2.fromdoppler.com/Automation/Automation/AutomationApp/",
      "isEnabled":false,
      "isSelected":false,
      "subNav":[

      ],
      "idHTML":"automationMenu",
      "$$hashKey":"object:8"
    },
    {
      "title":"Reportes",
      "url":"https://app2.fromdoppler.com/Campaigns/Reports/",
      "isEnabled":false,
      "isSelected":false,
      "subNav":[

      ],
      "idHTML":"reportMenu",
      "$$hashKey":"object:9"
    },
    {
      "title":"Plantillas",
      "url":"https://app2.fromdoppler.com/Templates/Main/",
      "isEnabled":false,
      "isSelected":true,
      "idHTML":"templateMenu",
      "$$hashKey":"object:10"
    }
  ]
}
```

fcoronelmakingsense:
I update this pr:

- Remove user from header properties
- Add user menu component

cbernat:
@fcoronelmakingsense good work! Next time you make corrections into a pr, it would be nice you made them into a new commit, to make it more readable :)

src/components/Header/HeaderNav/HeaderNav.js
```diff
@@ -1,10 +1,28 @@
 import React from 'react';
 
 const HeaderNav = (props) => {
+  const navItems = props.nav;
   return (
     <nav>
       <ul>
-        <li>{props.user.Email}</li>
+        {navItems.map((item, index) => (
```
lucianaDilizio: ❤️

src/components/Header/Header.js
```diff
@@ -1,12 +1,20 @@
 import React from 'react';
 import HeaderNav from './HeaderNav/HeaderNav';
+import HeaderUserMenu from './HeaderUserMenu/HeaderUserMenu';
+import headerData from '../../headerData.json';
```
andresmoschini: ❤️

src/headerData.json
andresmoschini: Does it make sense?

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
```diff
@@ -0,0 +1,46 @@
...
+          {(user.plan.isSubscribers === 'true' || user.plan.isMonthlyByEmail === 'true') ?(
+           <p>
+            {user.plan.planName} | <strong>{user.plan.maxSubscribers} {user.plan.itemDescription}</strong>
+           </p>)
+          : ''}
+          <p>
+            <strong>{user.plan.remainingCredits}</strong>{user.plan.description}
+          </p>
+        </div>
+         {(user.clientManager && user.plan.buttonUrl && user.plan.pendingFreeUpgrade !== 'true') ? (
```
andresmoschini: Esto es de Doppler, no de la WebApp, pero que lástima que manden strings para esto :(

src/App.js
```diff
@@ -92,13 +92,18 @@ class App extends Component {
...
+    if (isLoggedIn) {
+      return (
+        <div>
+          <Header />
+          <img src={logo} alt="logo" />
+          <Footer />
+        </div>
+      );
+    } else {
+      return <div>Loading...</div>;
```
lucianaDilizio: Will this change in a future? If it won't, we should use the react-intl lib here.

andresmoschini:
Hi @fcoronelmakingsense, I know that you are still working on it, but maybe there are already some steps that could be applied in a less ambitious PR.

What do you think about extracting some changes and create a new independent PR or reduce the scope of this one?

In that way, we could merge it soon and allow me to apply some refactorings over your changes.

Another option is that I ignore this PR and apply the refactorings over _master_, but you will have a lot of conflicts after.

src/App.js
fcoronelmakingsense: Yes will change. Need Mariana define something..

andresmoschini: By the moment, I think that `Loading...` is enough acceptable, or you could choose another option, like a spinner. 

So, define an option for ourselves, @MarianaMS will have the opportunity of change it in the future. In the meantime let's make it polished and following all of our requirements, for example using i18n.

Quiero decir, aunque algo no esté definido, no lo haganos a medias. Tomemos una decisión, aunque sea simple y hagamosla bien pulida.

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
andresmoschini: 💤 💤 💤

src/components/Header/Header.js
andresmoschini: 👍

src/App.js
```diff
@@ -92,13 +92,18 @@ class App extends Component {
...
 
   render() {
-    return (
-      <div>
-        <Header user={this.state.user} />
-        <img src={logo} alt="logo" />
-        <Footer />
-      </div>
-    );
+    const isLoggedIn = !!this.state.user;
```
andresmoschini: did you change the related code where `state.user` is set as _empty object_ in place of _null_ value?

fcoronelmakingsense: Now yes!

andresmoschini: 👍  Thanks!

PR #34, opened 2019-02-21
[feat] Integrate doppler ui system header

fcoronelmakingsense:
- Change markup to adapt doppler ui library header
- Add logic to open and close user menu
- Add logic in hover on main menu navigation

image

### DIFF
--- a/src/components/Header/HeaderNav/HeaderNav.js
+++ b/src/components/Header/HeaderNav/HeaderNav.js
@@ -1,31 +1,54 @@
 import React from 'react';
 
-const HeaderNav = (props) => {
-  const navItems = props.nav;
-  return (
-    <nav>
-      <ul>
-        {navItems.map((item, index) => (
-          <li key={index}>
-            <a className={item.isSelected ? 'active' : ''} href={item.url}>
-              {item.title}
-            </a>
-            {item.subNav ? (
-              <ul>
-                {item.subNav.map((itemSubNav, index) => (
-                  <li key={index}>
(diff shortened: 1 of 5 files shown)


src/components/Header/HeaderNav/HeaderNav.js
```diff
@@ -1,31 +1,54 @@
...
+    return (
+      <nav className="nav-left-main">
+        <div className="menu-main--container">
+          <ul className="menu-main">
+            {navItems.map((item, index) => (
+              <li
+                className="submenu-item"
+                key={index}
+                onMouseEnter={this.hoverMenu.bind(this, item.subNav)}
+                onMouseLeave={this.hoverMenu.bind(this)}
```
andresmoschini: Is it necessary to bind here? you already did that in the constructor.

src/components/Header/HeaderNav/HeaderNav.js
```diff
@@ -1,31 +1,54 @@
...
-};
+class HeaderNav extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.hoverMenu = this.hoverMenu.bind(this);
+  }
+
+  hoverMenu(hasSubNav) {
+    if (Array.isArray(hasSubNav) && hasSubNav.length) {
```
andresmoschini: If I understand it fine, here you are overloading the meaning of _hasSubNav_ with toggle status, what do you think about simply pass a boolean parameter?

Also, maybe we could break the components in a different way to avoid to verify _hasSubNav_ here.

PR #37, opened 2019-02-22
[fix] add datahub in header

fcoronelmakingsense:
- Add datahub in header and selected option
- Add last doppler ui system build to fix header navigation

### DIFF
--- a/src/components/Header/HeaderUserMenu/HeaderUserMenu.js
+++ b/src/components/Header/HeaderUserMenu/HeaderUserMenu.js
@@ -9,7 +9,7 @@ const HeaderUserMenu = (props) => {
           {user.avatar.text}
         </span>
       </span>
-      <div class="user-menu">
+      <div className="user-menu">
         <header>
           <span className="user-avatar--menu" style={{ background: user.avatar.color }}>
             {user.avatar.text}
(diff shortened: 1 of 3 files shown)


andresmoschini:
👍

src/headerData.json
```diff
@@ -108,19 +105,23 @@
...
       "isEnabled":false,
-      "isSelected":true,
+      "isSelected":false,
       "idHTML":"templateMenu",
       "$$hashKey":"object:10"
+    },
+    {
+      "title":"Doppler Data Hub",
+      "url":"https://cdn.fromdoppler.com/doppler-webapp/int-latest/",
+      "isEnabled":false,
```
cbernat: What's the meaning of this isEnabled always in false for all items?

fcoronelmakingsense: BE things haha

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
```diff
@@ -9,7 +9,7 @@ const HeaderUserMenu = (props) => {
           {user.avatar.text}
         </span>
       </span>
-      <div class="user-menu">
+      <div className="user-menu">
```
cbernat: 👍

cbernat:
👍

src/headerData.json
cbernat: If you do not use it, why mapping it here? :)

PR #41, opened 2019-02-25
Upgrade dependencies to the last ones

andresmoschini:
Hi Team, 

I have upgraded dependencies to the last ones, and also moved prettier as default dependency, fix line endings and update yarn.lock.

It will allow us to [use Hooks](https://reactjs.org/docs/hooks-intro.html)!


cbernat:
👍

PR #42, opened 2019-02-27
Move i18n inside app

andresmoschini:
Hi team, this PR is only to move localization behavior inside our app, if you change _componentWillMount_ in the following way you will be able to verify it

```js
componentWillMount() {
  this.interval = setInterval(() => {
    const locale = 'es';
    this.setState({
      i18n: {
        locale: locale,
        messages: flattenMessages(messages[locale]),
      },
    });
}, 3000);
```

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -114,11 +114,11 @@ class App extends Component {
     return (
       <IntlProvider locale={i18n.locale} messages={i18n.messages}>
         {isLoggedIn ? (
-          <div>
+          <>
             <Header />
             <img src={logo} alt="logo" />
             <Footer />
-          </div>
+          </>
         ) : (
           <div>
             <FormattedMessage id="loading" />
--- a/src/App.js
+++ b/src/App.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import logo from './logo.svg';
 import './App.css';
-import { addLocaleData, IntlProvider } from 'react-intl';
+import { addLocaleData, FormattedMessage, IntlProvider } from 'react-intl';
 import en from 'react-intl/locale-data/en';
 import es from 'react-intl/locale-data/es';
 import messages_es from './i18n/es.json';
@@ -120,7 +120,9 @@ class App extends Component {
             <Footer />
           </div>
         ) : (
-          <div>Loading...</div>
+          <div>
+            <FormattedMessage id="loading" />
+          </div>
         )}
       </IntlProvider>
     );
--- a/src/App.js
+++ b/src/App.js
@@ -1,23 +1,37 @@
 import React, { Component } from 'react';
 import logo from './logo.svg';
 import './App.css';
-import { addLocaleData } from 'react-intl';
+import { addLocaleData, IntlProvider } from 'react-intl';
 import en from 'react-intl/locale-data/en';
 import es from 'react-intl/locale-data/es';
+import messages_es from './i18n/es.json';
+import messages_en from './i18n/en.json';
+import { flattenMessages } from './utils';
 import jwt_decode from 'jwt-decode';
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -1,23 +1,38 @@
...
 
 class App extends Component {
   constructor() {
     super();
+    //TODO: this hardcoded data will depend by the app language
+    const locale = navigator.language.toLowerCase().split(/[_-]+/)[0] || 'en';
     this.state = {
       user: null,
       loginSession: {},
+      i18n: {
```
fcoronelmakingsense: need this in state?

src/App.js
```diff
@@ -96,17 +111,20 @@ class App extends Component {
...
-        </div>
-      );
-    } else {
-      return <div>Loading...</div>;
-    }
+    const i18n = this.state.i18n;
+    return (
+      <IntlProvider locale={i18n.locale} messages={i18n.messages}>
+        {isLoggedIn ? (
+          <div>
```
fcoronelmakingsense: maybe can change for `<>`, what do you think?

src/App.js
cbernat: What about using context for language localization? https://reactjs.org/docs/context.html

cbernat: @fcoronelmakingsense I think if the user wants to change the language it should be able to do so, app state seems to be a logical choice for this scenario.

src/App.js
andresmoschini: I am fine, I did not want to change your layout, but if `<>` is fine, great.

src/App.js
fcoronelmakingsense: yes maybe you are right! 💯

andresmoschini: Mmm, I heard that it is not a good idea using context API directly, but maybe that info is old.

In this case, we need the state only here, so I think that we do not need _context_ and neither other option.

We could make _messages_ a global variable and simply keep _locale_ in the state. It seemed to me to be more clear and isolated.

Alternative: 

```js
import { flattenMessages } from './utils';
import messages_es from './i18n/es.json';
import messages_en from './i18n/en.json';

// . . .

const messages = {
  es: messages_es,
  en: messages_en,
};

// . . .

class App extends Component {
  constructor() {
    // . . .
    this.state = {
      user: null,
      loginSession: {},
      locale: locale, // 'es' / 'en'
    }
  }

    render() {
      // . . .
      const locale = this.state.locale;
      return (
        <IntlProvider locale={locale} messages={flattenMessages(messages[locale])}>
```

What do you think?

cbernat: But messages will have both languages messages in context right? So we can change, because some time in the near future we will need this feature.

andresmoschini: @fcoronelmakingsense @cbernat @mvillaseco do you have any opinion?

andresmoschini: @lucianaDilizio

andresmoschini: > But messages will have both languages messages in context right? So we can change, because sometime in the near future we will need this feature.

En el ejemplo que puse, _messages_ es una variable global que se usa en la función de _render_. En el código del PR, en la función de _render_ solo se utilizan datos del _state_ (que fueron obtenidos de la variable global)

¿Tienen opinión?

cbernat: Capaz va mejor por fuera del render, no pondría en el state algo que no sea dinámico.

andresmoschini: Hablamos con @cbernat y coincidimos que la forma en que lo estamos haciendo en el PR parece más limpia.

Respecto al _Context_, no necesitamos usarlo explicitamente, si no me equivoco la librería lo usa internamente, en cualquier caso funciona sin problemas.

PR #45, opened 2019-03-01
Test i18n

andresmoschini:
Hi team (mainly @fcoronelmakingsense and @lucianaDilizio)

I have moved the default language configuration to index.js again to allow testing.

I am having an issue with a not related test, but I am PRing anyway to allow you to review my changes, could you kindly review?

### DIFF
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -1,23 +1,12 @@
 import React from 'react';
 import axios from 'axios';
-import { flattenMessages } from './utils';
 import { render, cleanup, wait } from 'react-testing-library';
 import 'jest-dom/extend-expect';
-
 import App from './App';
-
-import { IntlProvider } from 'react-intl';
-import messages_es from './i18n/es.json';
-import messages_en from './i18n/en.json';
 import jwt_decode from 'jwt-decode';
 
 //Add mock to decode jwt token and not fail
 jest.mock('jwt-decode');
-
-const messages = {
-  es: messages_es,
-  en: messages_en,
-};
 
 const response = {
   data: {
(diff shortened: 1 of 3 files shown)


andresmoschini:
@fcoronelmakingsense I am not sure if you have been working this XL weekend, or you could review it on Wednesday morning, I would really appreciate your review.

src/App.test.js
```diff
@@ -1,93 +1,77 @@
...
+        {
+          withCredentials: 'include',
+        },
+      );
+      expect(axios.get).toHaveBeenCalledWith(
+        process.env.REACT_APP_API_URL + '/Reports/Reports/GetJwtToken',
+        {
+          withCredentials: 'include',
+        },
+      );
```
cbernat: Shouldn't you check email value here? To detect if something went wrong with the data you are showing.

src/index.js
```diff
@@ -4,7 +4,10 @@ import './index.css';
 import App from './App';
 import * as serviceWorker from './serviceWorker';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+// TODO: this hardcoded data will depend by the app language
+const locale = navigator.language.toLowerCase().split(/[_-]+/)[0] || 'en';
```
cbernat: Did you think how will you get this language from user data, since it's outside app component?

cbernat:
I like your tests refactoring 💯 They are easier to follow.

src/App.test.js
andresmoschini: Mmm, in this particular case I do not think so. I mean, we are not sending the email to the backend, and the backend is _moked_.

Also, I am cleaning JwtToken related code to reintroduce it late, at that moment, we should verify different scenarios, for example when the stored token does not match Doppler's cookie, there will be a lot of tests.

src/index.js
cbernat: This does not avoid the fact that maybe you get `loading...` message and then the app starts working in spanish, if it were the user language. But then the loading message is only temporary, so I guess is fine.

andresmoschini:
@fcoronelmakingsense I need your blessing 🙏

PR #47, opened 2019-03-06
WIP - More refactoring

andresmoschini:
I AM ONLY CREATING THIS PR TO FORCE A DEPLOY, DO NOT PAY ATTENTION BY THE MOMENT.

### DIFF
--- a/src/services/doppler-mvc-client.ts
+++ b/src/services/doppler-mvc-client.ts
@@ -1,14 +1,98 @@
 import { AxiosInstance, AxiosStatic } from 'axios';
+import { Color } from 'csstype';
+import headerDataJson from '../headerData.json';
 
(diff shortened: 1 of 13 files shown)


src/services/doppler-mvc-client.ts
```diff
@@ -0,0 +1,124 @@
+import { AxiosInstance, AxiosStatic } from 'axios';
+import { Color } from 'csstype';
+import headerDataJson from '../headerData.json';
+
+// TODO: Consider better the name, DopplerMvc tries to make reference to Doppler API implemented
+// inside MVC App
+
+interface NavEntry {
```
fcoronelmakingsense: it's ok all interfaces in the same file?

andresmoschini: Mmm, I only them here, what do you think?

PR #49, opened 2019-03-07
Very minor and small: Use deconstructing to simplify code

andresmoschini:
What do you think?

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -56,11 +56,13 @@ class App extends Component {
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -56,11 +56,13 @@ class App extends Component {
   }
 
   render() {
-    const isLoggedIn = this.state.dopplerSession.status === 'authenticated';
-    const i18n = this.state.i18n;
+    const {
```
cbernat: Very nice!

PR #50, opened 2019-03-08
Include header data in session object

andresmoschini:
Hi team,

I am moving headers data inside session object and using it in header components.

I have also refined some header components in order to use a sanitized version of _DopplerLegacy's HeaderData_ object. 

In the way, I have detected that we are not paying attention to all _HeaderData_ properties, so, I changed the mappers to map only used properties so we could detect when we need something else.

Could you review it, please?

Next steps:

* Add tests.
* [DBR-94](http://jira.makingsense.com/browse/DBR-94) - Pay attention to _HeaderData_ language to translate the App.
* [DBR-81](http://jira.makingsense.com/browse/DBR-81) - Read _HeaderData_ from backend.
* [DBR-95](http://jira.makingsense.com/browse/DBR-95) - Move more AJAX calls to doppler-legacy-client
* Add tests (more yet).

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -2,19 +2,18 @@ import React from 'react';
 import HeaderNav from './HeaderNav/HeaderNav';
 import HeaderMessages from './HeaderMessages/HeaderMessages';
 import HeaderUserMenu from './HeaderUserMenu/HeaderUserMenu';
-import headerData from '../../headerData.json';
 import { FormattedMessage } from 'react-intl';
 
-const Header = () => {
+const Header = ({ userData: { user, nav, alert } }) => {
   return (
     <div>
-      {headerData.alert ? <HeaderMessages alert={headerData.alert} /> : null}
+      {alert ? <HeaderMessages alert={alert} /> : null}
       <header className="header-main">
         <div className="header-wrapper">
           <div className="logo">
             <span className="ms-icon icon-doppler-logo" />
           </div>
-          <HeaderNav nav={headerData.nav} />
+          <HeaderNav nav={nav} />
           <nav className="nav-right-main">
             <ul className="nav-right-main--list">
               <li>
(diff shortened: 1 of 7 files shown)


andresmoschini:
Next steps:
* Add tests.
* Pay attention to _HeaderData_ language to translate the App.
* Read _HeaderData_ from backend.
* Add tests (more yet).

src/components/Header/Header.js
```diff
@@ -2,19 +2,18 @@ import React from 'react';
 import HeaderNav from './HeaderNav/HeaderNav';
 import HeaderMessages from './HeaderMessages/HeaderMessages';
 import HeaderUserMenu from './HeaderUserMenu/HeaderUserMenu';
-import headerData from '../../headerData.json';
 import { FormattedMessage } from 'react-intl';
 
-const Header = () => {
+const Header = ({ userData: { user, nav, alert } }) => {
```
fcoronelmakingsense: its better than use props?

andresmoschini: In fact, we are using `props` here.

`props` object is being deconstructed. 

Given:

```js
const props = {
  userData: {
    user: {
        // . . . 
    }, 
    nav: {
        // . . . 
    },  
    alert: {
        // . . . 
    }, 
    another_property_1: : {
        // . . . 
    }, 
  },
  another_property_2: : {
        // . . . 
    }, 
};
```

This:

```js
const { userData: { user, nav, alert } } = props;
```

Is the same than:

```js
const user = props.userData.user;
const nav = props.userData.nav;
const alert = props.userData.alert;
```

So, it is shorter. But **the main advantage is** that it is clear what data we are using: we cannot use other data from _props_ without making it explicit. Does it make sense?

CC: @cbernat who was interested in deconstructing.

mvillaseco:
👍

src/components/Header/Header.js
fcoronelmakingsense: Yes, i understand now

PR #53, opened 2019-03-08
refactor(localization): encapsulate IntlProvider configuration and dependencies into a component

andresmoschini:
Hi Team, I have another little refactoring...

When I tried to introduce a test for `UpgradePlanForm.js`, I realized that I need _IntlProvider_ scope.

Talking with @mvillaseco, first, we thought about having a helper for the test, but then we understood that the same helper could be useful for all the application.

It is not a nice module, it works as a singleton, with a state, but I think that for the moment it is fine, and in the future, we could refactor it.

When we need to test independent components, we could do something like this:

```js-jsx
describe('UpgradePlanForm', () => {
  it('should . . . ', () => {
    const { getByText } = render(
      <DopplerIntlProvider>
        <UpgradePlanForm isSubscriber={true} />
      </IntlProvider>);
    // . . .
```

Could you review it?

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,24 +1,13 @@
 import React, { Component } from 'react';
 import logo from './logo.svg';
 import './App.css';
-import { addLocaleData, FormattedMessage, IntlProvider } from 'react-intl';
-import en from 'react-intl/locale-data/en';
-import es from 'react-intl/locale-data/es';
-import messages_es from './i18n/es.json';
-import messages_en from './i18n/en.json';
-import { flattenMessages } from './utils';
+import DopplerIntlProvider from './DopplerIntlProvider';
+import { FormattedMessage } from 'react-intl';
 import Header from './components/Header/Header';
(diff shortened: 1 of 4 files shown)


cbernat:
👍

andresmoschini:
Hit @fcoronelmakingsense, @cbernat and team, sorry, I have added a new commit to this PR.

It is a double to allow omit real translations in some tests, for example

```js-jsx
// . . .
import DopplerIntlProvider from '../../DopplerIntlProvider.double-with-ids-as-values';

describe('UpgradePlanForm', () => {
  it('should . . . ', () => {
    const { getByText, debug } = render(
      <DopplerIntlProvider>
        <UpgradePlanForm isSubscriber={true} />
      </IntlProvider>);
    debug();
    // . . .
```

Will render something like this:

```
      <body>
        <div>
          <div>
            <span>
              loading
            </span>
          </div>
        </div>
      </body>
```

Because `loading` is the intl id.

Does it make sense?

PR #56, opened 2019-03-12
Add di container

andresmoschini:
Hi team,

I know that it could be difficult to understand at the beginning, but I am sure that it will be very useful. 

> "Dependency injection is a tough problem. Especially in JavaScript. Lots of people didn't realize that but putting a proper dependency management is a key process of every development cycle. JavaScript ecosystem offers different tools and we as developers should pick the one that fits in our needs."
> _from [React in patterns](https://krasimir.gitbooks.io/react-in-patterns/chapter-10/#final-thoughts)_

I still need to add a _new helper component_ or _component wrapper_ to allow using it easily in any part of the system, in a way similar that we are doing with DopplerIntlProvider and FormattedMessage. I will do that in future PR to keep it small.

I am adding some comments about weird usages of ES6 and TypeScript.

See also: https://blog.ploeh.dk/2012/11/06/WhentouseaDIContainer/

Could you please review it?

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -4,28 +4,20 @@ import DopplerIntlProvider from './DopplerIntlProvider';
 import { FormattedMessage } from 'react-intl';
 import Header from './components/Header/Header';
 import Footer from './components/Footer/Footer';
-import axios from 'axios';
-import { HttpDopplerLegacyClient } from './services/doppler-legacy-client';
-import { OnlineSessionManager } from './services/session-manager';
 import Reports from './components/Reports/Reports';
 
 class App extends Component {
-  constructor(props) {
-    super(props);
+  constructor({ locale, dependencies: { sessionManager } }) {
+    super();
 
     this.updateSession = this.updateSession.bind(this);
 
-    this.sessionManager =
-      (props.dependencies && props.dependencies.sessionManager) ||
-      new OnlineSessionManager(
-        (props.dependencies && props.dependencies.dopplerLegacyClient) ||
-          new HttpDopplerLegacyClient(axios, process.env.REACT_APP_API_URL),
-        process.env.REACT_APP_DOPPLER_LEGACY_KEEP_ALIVE_MS,
-      );
+    /** @type { import('./services/session-manager').SessionManager } */
+    this.sessionManager = sessionManager;
 
     this.state = {
       dopplerSession: this.sessionManager.session,
-      i18nLocale: props.locale,
+      i18nLocale: locale,
     };
   }
 
--- a/src/services/pure-di.ts
+++ b/src/services/pure-di.ts
@@ -1,0 +1,63 @@
+import axios, { AxiosStatic } from 'axios';
+import { HttpDopplerLegacyClient, DopplerLegacyClient } from './doppler-legacy-client';
+import { OnlineSessionManager, SessionManager } from './session-manager';
+
+interface AppConfiguration {
+  dopplerLegacyUrl: string;
+  dopplerLegacyKeepAliveMilliseconds: number;
+}
+
+export interface AppServices {
+  axiosStatic: AxiosStatic;
+  appConfiguration: AppConfiguration;
+  dopplerLegacyClient: DopplerLegacyClient;
+  sessionManager: SessionManager;
+}
+
+export class AppCompositionRoot implements AppServices {
+  private readonly instances: Partial<AppServices>;
+
+  constructor(predefinedInstances: Partial<AppServices> = {}) {
+    this.instances = { ...predefinedInstances };
+  }
+
+  private singleton<N extends keyof AppServices, T extends AppServices[N]>(
+    name: N,
+    factory: () => T,
+  ): T {
+    if (!this.instances[name]) {
+      this.instances[name] = factory();
+    }
+    return this.instances[name] as T;
+  }
+
+  get axiosStatic() {
+    return this.singleton('axiosStatic', () => axios);
+  }
+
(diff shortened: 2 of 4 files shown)


src/App.js
```diff
@@ -5,27 +5,19 @@ import DopplerIntlProvider from './DopplerIntlProvider';
...
 import Footer from './components/Footer/Footer';
-import axios from 'axios';
-import { HttpDopplerLegacyClient } from './services/doppler-legacy-client';
-import { OnlineSessionManager } from './services/session-manager';
 
 class App extends Component {
-  constructor(props) {
-    super(props);
+  constructor({ locale, dependencies: { sessionManager } }) {
+    super();
```
andresmoschini: Here I am not calling _super(props)_, only _super()_. It is discouraged in React documentation, but they do not say the reason. 

I imagine that it is because, in this way, _this.props_ is not available inside the constructor, but it is after that.

In my opinion, using deconstructing in constructor parameter make it more explicit and clear, so, I prefer this way.

src/App.js
```diff
@@ -5,27 +5,19 @@ import DopplerIntlProvider from './DopplerIntlProvider';
...
     this.updateSession = this.updateSession.bind(this);
 
-    this.sessionManager =
-      (props.dependencies && props.dependencies.sessionManager) ||
-      new OnlineSessionManager(
-        (props.dependencies && props.dependencies.dopplerLegacyClient) ||
-          new HttpDopplerLegacyClient(axios, process.env.REACT_APP_API_URL),
-        process.env.REACT_APP_DOPPLER_LEGACY_KEEP_ALIVE_MS,
-      );
+    /** @type { import('./services/session-manager').SessionManager } */
```
andresmoschini: 🚧

src/services/pure-di.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export class AppCompositionRoot implements AppContainer {
+  private readonly instances: Partial<AppContainer>;
+
+  constructor(predefinedInstances: Partial<AppContainer> = {}) {
+    this.instances = { ...predefinedInstances };
+  }
+
+  private singleton<N extends keyof AppContainer, T extends AppContainer[N]>(
```
andresmoschini: 🚧

src/services/pure-di.ts
```diff
@@ -0,0 +1,63 @@
...
+  axiosStatic: AxiosStatic;
+  appConfiguration: AppConfiguration;
+  dopplerLegacyClient: DopplerLegacyClient;
+  sessionManager: SessionManager;
+}
+
+export class AppCompositionRoot implements AppContainer {
+  private readonly instances: Partial<AppContainer>;
+
+  constructor(predefinedInstances: Partial<AppContainer> = {}) {
```
andresmoschini: `Partial<T>` is a type that has some (or none) of the properties of `T`, it allows us to accept a subset of _AppContainer_ properties, to override default _AppCompositionRoot_ services.

See more details in https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types

src/services/pure-di.ts
```diff
@@ -0,0 +1,63 @@
...
+    name: N,
+    factory: () => T,
+  ): T {
+    if (!this.instances[name]) {
+      this.instances[name] = factory();
+    }
+    return this.instances[name] as T;
+  }
+
+  get axiosStatic() {
```
andresmoschini: I am using properties in order to have lazy initialization.

Another option could be using factory methods, for example: `getAxiosStatic()`, but it does not match _AppContainer_ interface:

```js
export interface AppContainer {
  axiosStatic: AxiosStatic;
  appConfiguration: AppConfiguration;
  dopplerLegacyClient: DopplerLegacyClient;
  sessionManager: SessionManager;
}
```

Another option could be using all as factories, the same for the interface: 


```js
export interface AppContainer {
  getAxiosStatic: () => AxiosStatic;
  getAppConfiguration: () => AppConfiguration;
  getDopplerLegacyClient: () => DopplerLegacyClient;
  getSessionManager: () => SessionManager;
}
```

src/services/pure-di.ts
andresmoschini: This type is not so easy to read, but it is great!:

It only allows call `Singleton` function with the right key and type pair, for example: 

```
this.singleton('appConfiguration', X);
```

_X_ could be only a object that implements the class _AppConfiguration_

src/App.js
andresmoschini: It looks a little weird but it is the way that I found to allow intellisense for this injected property.

fcoronelmakingsense: What does this line?

andresmoschini: Oh, sorry, I thought that I have added a clarificatory comment, but it seems that something went wrong. 

As you have seen, _sessionManager_ instance is being injected using props, so IDE cannot determine the type of it and provide us a nice intellisense experience.

Since it is a JavaScript file and not a TypeScript one, we need using JSDoc to specify the type. See more information here: https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#import-types

andresmoschini: Does it make sense?

src/App.js
cbernat: I agree it's more readable this way. `constructor({ locale, dependencies: { sessionManager } }) {`

andresmoschini:
We have discussed these changes in a meeting with @cbernat, @fcoronelmakingsense, and @GusBaamonde.

PR #57, opened 2019-03-14
Add DI to context

andresmoschini:
This continues the working started in #56 

Now, based on the meeting that we have with @cbernat, @fcoronelmakingsense and @GusBaamonde. We noticed that it was not so nice having to update parent component to inject dependencies into child components.

I found a better option using [Higher-order component pattern](https://krasimir.gitbooks.io/react-in-patterns/chapter-04/#higher-order-component). In this way, with a simple decorator, we make our components ready to get dependencies from the context.

The tests are a little weird, but they really test our DI related code.

The next step is using this approach to inject _DopplerLegacyClient_ into the components that consume Doppler Legacy backend.

Could you review? I think that reading it commit by commit could be a good idea. Let me know if you need a meeting to understand it.


andresmoschini:
Here you have a working example:

image

PR #59, opened 2019-03-14
chore(header): add alert into mocked data and update library

cbernat:
image

### DIFF
--- a/src/headerData.json
+++ b/src/headerData.json
@@ -175,5 +175,17 @@
         "$$hashKey":"object:26"
       }
     ]
+  },
+  "alert":{
+    "type":"blocker",
+    "message":"Para que puedas enviar tu próxima Campaña, antes necesitamos verificar el origen de tus Suscriptores.",
+    "button":{
+      "action":"validateSubscribersPopup",
+      "text":"Verificar Ahora"
+    },
+    "nextAlert":{
+      "type":"warning",
+      "message":"Estamos verificando el origen de tus Suscriptores. Te contactaremos para que puedas continuar con tu Campaña."
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


src/headerData.json
```diff
@@ -175,5 +175,17 @@
...
       }
     ]
+  },
+  "alert":{
+    "type":"blocker",
+    "message":"Para que puedas enviar tu próxima Campaña, antes necesitamos verificar el origen de tus Suscriptores.",
+    "button":{
+    "action":"validateSubscribersPopup",
+    "text":"Verificar Ahora"
+    },
```
andresmoschini: Indentation?

andresmoschini: Spaces?

src/headerData.json
```diff
@@ -175,5 +175,17 @@
         "$$hashKey":"object:26"
       }
     ]
+  },
+  "alert":{
+    "type":"blocker",
+    "message":"Para que puedas enviar tu próxima Campaña, antes necesitamos verificar el origen de tus Suscriptores.",
+    "button":{
+    "action":"validateSubscribersPopup",
```
andresmoschini: It is out of the scope of your task, but this kind of alert will not work in our webapp, right?

andresmoschini:
image

src/headerData.json
cbernat: Yes you are right. This button activates a popup.

cbernat:
@andresmoschini yes it is broken. I will add a bug to fix this. There is an issue with the class that we use from webapp and the one that they use in UI library

src/headerData.json
andresmoschini: Spaces where fine, deleting my previous comment...

PR #62, opened 2019-03-15
Datahubclient abstraction

andresmoschini:
Hi team, continuing with my PR series, I am adding an abstraction for Datahub access.

BTW, it seems that Datahub API is different than we expect, but we will deal with it later.


andresmoschini:
I rebased these commits because there was an issue with data types and I have also removed "Verify domain again" entry.

cbernat:
👍

PR #63, opened 2019-03-15
Add Dockerfile to project

fgchaio:
Hi @FromDoppler/doppler-webapp 

In this pr Im adding a Dockerfile to the project, to allow building it in a docker image.
This will be of use to get the project running as a service in the server.

Please have a look at it, any review is welcome. 😄

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,17 @@
-#get node image, comes with yarn preinstalled
-
 FROM node:10 As build
 WORKDIR /app
 COPY package.json yarn.lock ./
 
-#make yarn process packages
+# Make yarn process packages
 RUN yarn install
 
-#get all files
+# Get all files
 COPY . .
 
-#Build project
+# Build project
 RUN yarn build
 
-#host project in nginx
+# Host project in nginx
 FROM nginx
 WORKDIR /app
 COPY --from=build /app/build /usr/share/nginx/html
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+#get node image, comes with yarn preinstalled
+
+FROM node:10 As build
+WORKDIR /app
+COPY package.json yarn.lock ./
+
+#make yarn process packages
+RUN yarn install
+
+#get all files
+COPY . .
+
+#Build project
+RUN yarn build
+
+#host project in nginx
+FROM nginx
+WORKDIR /app
+COPY --from=build /app/build /usr/share/nginx/html
+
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,22 @@
+#get node image, comes with yarn preinstalled
```
andresmoschini: I think that this comment is not necessary

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+FROM node:10 As build
+WORKDIR /app
+COPY package.json yarn.lock ./
+
+#make yarn process packages
```
andresmoschini: In general, we are using a space between the `#` and the text

```suggestion
# Make yarn process packages
```

andresmoschini:
I know that it is not necessary, because the way is always the same, but since we are a little neophyte in docker usage, what do you think about adding these two `.sh` files to illustrate how to use this file:

* `build-n-create-docker-image.sh`: to generate _doppler-webapp_ image locally, something [like this](https://github.com/MakingSense/doppler-reports-api/blob/master/build-w-docker.sh).
* `run-docker-container.sh`: to run the image locally, something [like this](https://github.com/MakingSense/doppler-reports-api/blob/master/run-docker-container.sh)

Dockerfile
```diff
@@ -0,0 +1,22 @@
+#get node image, comes with yarn preinstalled
+
+FROM node:10 As build
```
andresmoschini: I know that it is out of the scope of your work, but since node 10 is the last stable version and it works fine, could you update the other references to node images:
* https://github.com/FromDoppler/doppler-webapp/blob/master/build-w-docker.sh#L20
* https://github.com/FromDoppler/doppler-webapp/blob/master/watch-w-docker.sh#L19
* https://github.com/FromDoppler/doppler-webapp/blob/master/verify-w-docker.sh#L16
* Any other?

andresmoschini:
Nice work @fgchaio! let me know if you need something to configure the CI with Docker Hub.

fgchaio:
> I know that it is not necessary, because the way is always the same, but since we are a little neophyte in docker usage, what do you think about adding these two `.sh` files to illustrate how to use this file
> 
Sure, I like that idea!

Dockerfile
fgchaio: will do

andresmoschini:
Thanks @fgchaio! this step is very important. I really appreciate it.

PR #64, opened 2019-03-15
Improve buy form

andresmoschini:
Hi @cbernat, @fcoronelmakingsense and also @GusBaamonde and @damianmuti (because I am making some minor changes in HTML structure),

I was working on another task, and these changes come alone, I tried to split each different improvement in a different commit. 

Could you review?

### DIFF
--- a/src/components/UpgradePlanForm/UpgradePlanForm.js
+++ b/src/components/UpgradePlanForm/UpgradePlanForm.js
@@ -37,7 +37,11 @@ class UpgradePlanForm extends React.Component {
         ...prevState.userPlanModel,
         [name]: value,
       },
-      formIsValid: !!this.state.userPlanModel.Detail,
+    }));
+
+    // In a different setState to be sure of merging old state with the new changes
+    this.setState((prevState) => ({
+      formIsValid: !!prevState.userPlanModel.Detail,
     }));
   };
 
--- a/src/components/UpgradePlanForm/UpgradePlanForm.js
+++ b/src/components/UpgradePlanForm/UpgradePlanForm.js
@@ -78,13 +78,17 @@ class UpgradePlanForm extends React.Component {
                   <label htmlFor="message">
                     <FormattedMessage id="common.message" />
                   </label>
-                  <textarea
-                    onChange={this.changeHandler}
-                    value={this.state.userPlanModel.Detail || ''}
-                    name="Detail"
-                    id="Detail"
-                    placeholder="Tu mensage"
-                  />
+                  <FormattedMessage id="upgradePlanForm.message_placeholder">
+                    {(placeholderText) => (
+                      <textarea
+                        onChange={this.changeHandler}
+                        value={this.state.userPlanModel.Detail || ''}
+                        name="Detail"
+                        id="Detail"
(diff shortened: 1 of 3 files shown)


src/components/UpgradePlanForm/UpgradePlanForm.js
```diff
@@ -2,6 +2,11 @@ import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import { InjectAppServices } from '../../services/pure-di';
 
+const fieldNames = {
+  selectedPlanId: 'selectedPlanId',
+  message: 'message',
```
mvillaseco: extra ,

src/components/UpgradePlanForm/UpgradePlanForm.js
```diff
@@ -52,31 +66,33 @@ class UpgradePlanForm extends React.Component {
...
                   </label>
                   <textarea
                     onChange={this.changeHandler}
-                    value={this.state.userPlanModel.Detail || ''}
-                    name="Detail"
-                    placeholder="Tu mensage"
+                    value={formData[fieldNames.message] || ''}
+                    id={fieldNames.message}
+                    name={fieldNames.message}
+                    placeholder="Tu mensage" // TODO: translate it
```
andresmoschini: Oh, I will try to fix it before merging.

src/components/UpgradePlanForm/UpgradePlanForm.js
```diff
@@ -89,9 +105,10 @@ class UpgradePlanForm extends React.Component {
                 <FormattedMessage id="common.cancel" />
               </button>
               <button
+                type="submit"
```
andresmoschini: I will also apply this before merging:

image

src/components/UpgradePlanForm/UpgradePlanForm.js
andresmoschini: Ready.

src/components/UpgradePlanForm/UpgradePlanForm.js
```diff
@@ -2,6 +2,11 @@ import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import { InjectAppServices } from '../../services/pure-di';
 
+const fieldNames = {
```
andresmoschini: What do you think about it @fcoronelmakingsense? should I keep these constants for the names? or should I remove them?

andresmoschini: Any other consideration before merging?

cbernat: If we are going to use this type of consts for handling string constants in everywhere, we should agree on this, there are a lot of constants strings in doppler. In angularJS we used to have a constants file, here I'm not so sure of the best way to handle this. Perhaps we could inject constants? Keep in mind we will see this types again in the future surely @andresmoschini .

cbernat: I understand this constants are only for this model, ignore my previous comment , I will remove it. Will we handle string constants this way from now on? @fcoronelmakingsense @andresmoschini

PR #65, opened 2019-03-16
Fix issues with date representation

andresmoschini:
Hi team, 

I do not know why the tests are passing in CI, but they started to fail in my local environment and it was because I was using Dates in _Reports.test.js_'s assertions.

I fixed that, and I saw that besides that, we were rendering a wrong date because _FormattedDate_ uses current timezone by default. 

In order to avoid this problem by the moment **we should use _timeZone_ parameter** in this way:

```
<FormattedDate value={domainSelected.verified_date} timeZone="UTC" />
``` 

I have also removed some redundant assertions.

Could you kindly review it?

### DIFF
--- a/src/components/Reports/Reports.test.js
+++ b/src/components/Reports/Reports.test.js
@@ -4,16 +4,19 @@ import 'jest-dom/extend-expect';
 import DopplerIntlProvider from '../../DopplerIntlProvider.double-with-ids-as-values';
 import Reports from './Reports';
 
+const verifiedDateAsEngString = '12/17/2017';
+const verifiedDateAsDate = new Date('2017-12-17');
+
 const fakeData = [
   {
     id: 1,
     name: 'www.fromdoppler.com',
-    verified_date: new Date('2017-12-17'),
+    verified_date: verifiedDateAsDate,
   },
   {
     id: 2,
     name: 'www.makingsense.com',
-    verified_date: new Date('2010-12-17'),
+    verified_date: verifiedDateAsDate,
   },
 ];
 
@@ -46,9 +49,9 @@ describe('Reports page', () => {
       </DopplerIntlProvider>,
     );
 
-    await wait(() => getByText(fakeData[0].verified_date));
+    await wait(() => getByText(verifiedDateAsEngString));
 
-    const verifiedDate = getByText(fakeData[0].verified_date);
+    const verifiedDate = getByText(verifiedDateAsEngString);
     const domain = getByText(fakeData[1].name);
 
     expect(verifiedDate).toBeDefined();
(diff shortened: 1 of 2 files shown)


src/components/Reports/Reports.test.js
```diff
@@ -46,13 +49,8 @@ describe('Reports page', () => {
       </DopplerIntlProvider>,
     );
 
-    await wait(() => getByText(fakeData[0].verified_date));
-
-    const verifiedDate = getByText(fakeData[0].verified_date);
-    const domain = getByText(fakeData[1].name);
-
-    expect(verifiedDate).toBeDefined();
```
fcoronelmakingsense: i didn't know no need expect in a test. I always saw it in the examples

src/components/Reports/ReportsFilters/ReportsFilters.js
```diff
@@ -39,7 +39,7 @@ const ReportsFilters = ({
             {domainSelected ? (
               <span className="verified--domain">
                 <FormattedMessage id="reports_filters.verified_domain" />{' '}
-                <FormattedDate value={domainSelected.verified_date} />
+                <FormattedDate value={domainSelected.verified_date} timeZone="UTC" />
```
fcoronelmakingsense: 👍

src/components/Reports/Reports.test.js
cbernat: Why did you remove this expects? Shouldn't all tests have at least one expect?

andresmoschini: You said me that _getByText_ fails if the literal is not found, so, expected is redundant, right?

fcoronelmakingsense: yes but more clear with expect. for me at least 😄

andresmoschini: I removed because _getByText_ verify that the expected text exists, for example:

```html
<div>2018-01-01</div>
```

```js
const verifiedDate = getByText('2020-05-05'); // It breaks here.
expect(verifiedDate).toBeDefined(); // this line is never executed.
```

---

```html
<div>2018-01-01</div>
```

```js
const verifiedDate = getByText('2018-01-01'); // It pass, so verifiedDate is defined
expect(verifiedDate).toBeDefined(); // this line will pass
```

andresmoschini: Mmmm, in general, I think that it is not a good idea having redundant code, in my opinion, it tends to be dangerous. 

But, if you want I can add these lines again, by the moment it is not so important (for me)

andresmoschini: @fcoronelmakingsense @cbernat let me know your opinion, in order to update it (or not) before merging.

fcoronelmakingsense: i'm already put my opinion but it's works the 2 ways

cbernat: Yes it is redundant, and yes it is clearer with the expect. Both ways work for me. I think is an issue of how we want to work.

PR #66, opened 2019-03-16
Add console log to doubles

andresmoschini:
Hi team, this is only a thing that I consider useful when we run this project locally.

image

What do you think?

### DIFF
--- a/src/services/datahub-client.doubles.ts
+++ b/src/services/datahub-client.doubles.ts
@@ -31,13 +31,13 @@ const fakeData = [
 export class HardcodedDatahubClient implements DatahubClient {
   public async getAccountDomains() {
     console.log('getAccountDomains');
-    await timeout(3000);
+    await timeout(1500);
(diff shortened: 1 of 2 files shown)


src/services/datahub-client.doubles.ts
```diff
@@ -30,11 +30,13 @@ const fakeData = [
 
 export class HardcodedDatahubClient implements DatahubClient {
   public async getAccountDomains() {
+    console.log('getAccountDomains');
     await timeout(3000);
```
fcoronelmakingsense: maybe for local 3seconds it's a lot of time. what do you think?

fcoronelmakingsense:
Good! 👍

src/services/datahub-client.doubles.ts
andresmoschini: Agree, I am changing it. 

Thanks!

PR #67, opened 2019-03-17
YAPR - Add routing

andresmoschini:
Yes, another PR.

I am introducing [React Router](https://reacttraining.com/react-router/), by the moment I am not doing more than redirect to _reports_ page when the root path is open.

Could you review?

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import './App.css';
 import DopplerIntlProvider from './DopplerIntlProvider';
 import { FormattedMessage } from 'react-intl';
+import { BrowserRouter as Router, Route, Redirect } from 'react-router-dom';
 import Header from './components/Header/Header';
 import Footer from './components/Footer/Footer';
 import Reports from './components/Reports/Reports';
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -47,14 +48,19 @@ class App extends Component {
...
-          <>
-            <Header userData={userData} />
-            <Reports />
-            <Footer />
-          </>
+          <Router>
+            <div>
+              <Header userData={userData} />
+              <Route path="/" exact component={redirectToReports} />
+              <Route path="/reports/" exact component={Reports} />
```
fcoronelmakingsense: in doppler the routes are in Title Clase like `Reports` . no need maintain consistency?''''''''''''

fcoronelmakingsense:
Nice change 👍

src/App.js
andresmoschini: Mmm, in general apps has lowercase URLs, in my opinion, we should take advantage of this opportunity to change it.

andresmoschini: I did not get answer on this, I think that in general Doppler's URLs are really ugly, so I do not see bad take this opportunity.

Merging...

PR #69, opened 2019-03-18
fixed: to do header messages

fcoronelmakingsense:
- add modal when alert if type of upgrade plan
- add validation to not show alert when if to verify suscribers
- fix prettier

### DIFF
--- a/src/components/Header/HeaderMessages/HeaderMessages.js
+++ b/src/components/Header/HeaderMessages/HeaderMessages.js
@@ -1,28 +1,62 @@
 import React from 'react';
+import Modal from '../../../components/Modal/Modal';
+import UpgradePlanForm from '../../UpgradePlanForm/UpgradePlanForm';
 
-const HeaderMessages = ({ alert: { type, message, button } }) => {
-  return (
-    <div className={'messages-container ' + type}>
-      <div className="wrapper">
-        <p>{message}</p>
-        {button ? (
-          button.url ? (
-            <a
-              href={button.url}
-              className="button button--light button--tiny"
-              data-testid="linkButton"
-            >
-              {button.text}
-            </a>
-          ) : (
-            <button className="button button--light button--tiny" data-testid="actionButton">
-              {button.text}
-            </button>
-          )
-        ) : null}
-      </div>
-    </div>
-  );
-};
+class HeaderMessages extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      buyModalIsOpen: false,
+    };
+
+    this.handleOpenBuyModal = () => this.setState({ buyModalIsOpen: true });
+    this.handleCloseBuyModal = () => this.setState({ buyModalIsOpen: false });
+  }
+
+  render() {
+    const alert = this.props.alert;
+    const user = this.props.user;
+    //TODO implement max subscribers modal
+    if (alert.button && alert.button.action !== 'updatePlanPopup') {
+      return <></>;
+    } else {
+      return (
+        <div className={'messages-container ' + alert.type}>
+          <div className="wrapper">
+            <p>{alert.message}</p>
+            {alert.button && alert.button.url ? (
+              <a
+                href={alert.button.url}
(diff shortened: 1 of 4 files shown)


mvillaseco:
👍

src/headerData.json
```diff
@@ -180,8 +180,8 @@
     "type": "blocker",
     "message": "Para que puedas enviar tu próxima Campaña, antes necesitamos verificar el origen de tus Suscriptores.",
     "button": {
-      "action": "validateSubscribersPopup",
-      "text": "Verificar Ahora"
+      "action": "updatePlanPopup",
```
cbernat: Take in account that header type `blocker` is a red one.  If you want the one for update please check the type, `info` for green.

cbernat:
👍

src/components/Header/HeaderMessages/HeaderMessages.js
```diff
@@ -1,28 +1,64 @@
...
+    //TODO implement max subscribers modal
+    if (alert.button && alert.button.action === 'validateSubscribersPopup') {
+      return <></>;
+    } else {
+      return (
+        <div className={'messages-container ' + alert.type}>
+          <div className="wrapper">
+            <p>{alert.message}</p>
+            {alert.button ? (
+              alert.button.url ? (
```
andresmoschini: What do you think about writting it in the following way to avoid nesting?

```
alert.button && alert.button.url ? ( . . . )
: alert.button ? ( . . . )
: null
```

src/components/Header/HeaderMessages/HeaderMessages.js
```diff
@@ -1,28 +1,64 @@
...
+
+    this.handleOpenBuyModal = () => this.setState({ buyModalIsOpen: true });
+    this.handleCloseBuyModal = () => this.setState({ buyModalIsOpen: false });
+  }
+
+  render() {
+    const alert = this.props.alert;
+    const user = this.props.user;
+    //TODO implement max subscribers modal
+    if (alert.button && alert.button.action === 'validateSubscribersPopup') {
```
andresmoschini: What do you think, in place of catch the actions that we do not know, catch only the actions that we know.

I mean, maybe today we do not understand `validateSubscribersPopup`, but in the future maybe there will be more.

src/components/Header/HeaderMessages/HeaderMessages.js
andresmoschini: Based on my previous comment:

> What do you think, in place of catch the actions that we do not know, catch only the actions that we know.
> 
> I mean, maybe today we do not understand validateSubscribersPopup, but in the future maybe there will be more.

What do you think about this:

```
alert.button && alert.button.url ? ( . . . )
: alert.button && alert.button.action === 'expectedAction') ? ( . . . )
: null
```

src/components/Header/HeaderMessages/HeaderMessages.js
```diff
@@ -1,28 +1,64 @@
 import React from 'react';
+import Modal from '../../../components/Modal/Modal';
+import UpgradePlanForm from '../../UpgradePlanForm/UpgradePlanForm';
```
andresmoschini: Good catch! I did not know that one of the actions was "buy".

src/components/Header/HeaderMessages/HeaderMessages.js
andresmoschini: @fcoronelmakingsense si no me pude explicar, por favor hablemosló antes de mergearlo en esta forma.

fcoronelmakingsense: we have only 2 actions. upgrade plan and validate subscribers. i only catch the validate subscribers modal because only todo implement that

src/components/Header/HeaderMessages/HeaderMessages.js
fcoronelmakingsense: fixed

PR #70, opened 2019-03-18
fix: button styles

GusBaamonde:
**Add styles for button disabled inside modal form**
_The build 102 has updated_


andresmoschini:
👍

PR #73, opened 2019-03-19
fix: header messages tests

fcoronelmakingsense:
- Fix header message test

### DIFF
--- a/src/components/Header/HeaderMessages/HeaderMessages.js
+++ b/src/components/Header/HeaderMessages/HeaderMessages.js
@@ -18,7 +18,7 @@ class HeaderMessages extends React.Component {
     const alert = this.props.alert;
     const user = this.props.user;
     //TODO implement max subscribers modal
-    if (alert.button && alert.button.action !== 'updatePlanPopup') {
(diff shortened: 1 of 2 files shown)


src/components/Header/HeaderMessages/HeaderMessages.js
```diff
@@ -18,7 +18,7 @@ class HeaderMessages extends React.Component {
     const alert = this.props.alert;
     const user = this.props.user;
     //TODO implement max subscribers modal
-    if (alert.button && alert.button.action !== 'updatePlanPopup') {
+    if (alert.button && alert.button.action && alert.button.action !== 'updatePlanPopup') {
```
andresmoschini: ❤️

andresmoschini:
👍

src/components/Header/HeaderMessages/HeaderMessages.js
andresmoschini: Haha, may we do not need this code any more :P 

https://github.com/MakingSense/Doppler/pull/5778/files#diff-625a0ce670e4f88df70e080e87d6d270R2540

It does not care, do not change it, only to notice you.

PR #75, opened 2019-03-19
[DBR-112] Add Dockerfiles for int and qa

fgchaio:
hi @FromDoppler/doppler-webapp team, im adding dockerfiles for int and qa.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+# Any change to this file should be replecated in dockerfile.int and dockerfile.qa
 FROM node:10 As build
 WORKDIR /app
 COPY package.json yarn.lock ./
--- a/Dockerfile.QA
+++ b/Dockerfile.QA
@@ -1,3 +1,4 @@
+# Any change to this file should be replecated in dockerfile.int and dockerfile
 FROM node:10 As build
 WORKDIR /app
 COPY package.json yarn.lock ./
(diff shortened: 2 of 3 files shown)


Dockerfile.QA
```diff
@@ -0,0 +1,20 @@
...
+# Build project
+RUN yarn build:qa
+
+# Host project in nginx
+FROM nginx
+WORKDIR /app
+COPY --from=build /app/build /usr/share/nginx/html
+
+ENTRYPOINT ["nginx", "-g", "daemon off;"]
+EXPOSE 80 443
```
andresmoschini: Do you configure the real ports in swarm configuration?

Dockerfile.QA
```diff
@@ -0,0 +1,20 @@
+FROM node:10 As build
```
andresmoschini: What do you think about adding a comment in each dockerfile file about that all files have duplicated code and should be edited together?

andresmoschini:
❤️

Dockerfile.QA
fgchaio: Exactly, we where thinking of adding it to readme or some other file to give some visibility to it.

Dockerfile.QA
fgchaio: you are right. Forgot about it thx.

Dockerfile
```diff
@@ -1,3 +1,4 @@
+#Any change to this file should be replecated in dockerfile.int and dockerfile.qa
```
andresmoschini: Very, very minor ;)

```suggestion
# Any change to this file should be replecated in dockerfile.int and dockerfile.qa
```

fgchaio: 😆

PR #76, opened 2019-03-20
fix: favicon and title

fcoronelmakingsense:
- Replace react favicon for doppler favicon
- Change page tittle for Doppler | Reportes. I don't know if can translate this because put in html and don't react code here


public/index.html
```diff
@@ -21,7 +21,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>React App</title>
+    <title>Doppler | Reportes</title>
```
andresmoschini: Since this webapp will have more than reports, what do you think about hardcoding only `Doppler` in the HTML.

Besides that, optionally, you could change the title to `Doppler | Reportes` (with translation) in `componentDidMount` of `Reports`.

fcoronelmakingsense: Fixed

src/components/Reports/Reports.js
```diff
@@ -47,9 +49,19 @@ class Reports extends React.Component {
...
   render() {
     return (
       <>
+        <FormattedMessage id='reports_title'>
+          {(message) => {
+            if (document.title !== message) {
+              document.title = message
+            }
+
+            return null
```
andresmoschini:
```suggestion
            return null;
```

PR #77, opened 2019-03-21
Minor refactorings

andresmoschini:
Hi team, 

I was working in another task related to login form, and I have detected these refactorings. 

So, in order to keep the other PR clean, I am proposing these changes to review in advance.

Could you kindly review them?

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -52,7 +52,9 @@ class App extends Component {
 
     return (
       <DopplerIntlProvider locale={i18nLocale}>
-        {sessionStatus === 'authenticated' ? (
+        {sessionStatus === 'unknown' ? (
+          <Loading />
+        ) : sessionStatus === 'authenticated' ? (
           <Router>
             <div>
               <Header userData={userData} />
@@ -62,7 +64,7 @@ class App extends Component {
             </div>
           </Router>
         ) : (
-          <Loading />
+          <div>Not implemented</div>
         )}
       </DopplerIntlProvider>
     );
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -62,9 +62,7 @@ class App extends Component {
             </div>
           </Router>
         ) : (
-          <div>
-            <FormattedMessage id="loading" />
-          </div>
+          <Loading />
```
cbernat: Great! I think this will be used a lot. And it will made easier to refactor into a spinner or whatever later on.

src/services/session-manager.ts
```diff
@@ -71,17 +71,17 @@ export class OnlineSessionManager implements SessionManager {
...
+    // this.updateSession({ status: 'non-authenticated' });
   }
 
   // TODO: move into a dependency
   private redirect() {
     const currentUrlEncoded = encodeURI(window.location.href);
     // TODO: only use redirect on login, not in logout
     const loginUrl = `${process.env.REACT_APP_API_URL}/SignIn/index?redirect=${currentUrlEncoded}`;
-    window.setTimeout(() => {
-      // Redirecting in a timeout in order to allow React to update UI
```
cbernat: Why removing this timeout now? seems to be neccesary, did you add another thing to fix react ui update?

andresmoschini: Well, I thought the same, but @fcoronelmakingsense was right.

If we redirect directly to Doppler when the session status is _not-autenticated_, we do not need to deal with that state. Is simpler. 

Of course, we will need to add a login in our side, so, it will be necessary to deal with _not-autenticated_, but that is another topic.

andresmoschini: I just cleaned a little more and deleted deprecated comments.

PR #78, opened 2019-03-21
Public and private routes

andresmoschini:
Hi team, 

I am still working on it because I need to add some tests.

But, I think that, besides that, it is ready to be merged. I finally resolved how to deal with many problems:
* Public and private routes
* Two different login pages (internal one and in Legacy Doppler)
* The layout that changes based on the route, redirect

I think that it is easy to review it if you read commit by commit, maybe it is difficult to understand DI related commit because I am using our DI system to inject a React component, but I think that it is the most consistent way to do that. Let me know if you need any clarification on this.

Pending tests:

- [x] not authenticated user should be redirected to Legacy Doppler Login after open /reports (when using RedirectToLegacyLoginFactory)
- [x] not authenticated user should be redirected to Internal Login after open /reports (when using RedirectToInternalLogin)
- [x] not authenticated user should not be redirected after open /login
- [x] not authenticated user should be redirected to /login when the route does not exist
- [x] ~~not authenticated user should be redirected to original page after login~~ It is not possible testing it, because it is a _login page_ responsibility, and it is not created yet.
- [x] authenticated user should not be redirected after open /reports
- [x] authenticated user should be redirected to /reports when the route does not exist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -7,6 +7,12 @@ import Footer from './components/Footer/Footer';
 import Reports from './components/Reports/Reports';
 import { InjectAppServices } from './services/pure-di';
 import Loading from './components/Loading/Loading';
+// If we want to use Internal Login
+// import { RedirectToInternalLogin as RedirectToLogin } from './components/RedirectToLogin';
+
+// If we want to use redirect to Doppler Legacy Login
+import { RedirectToLegacyLoginFactory } from './components/RedirectToLogin';
+const RedirectToLogin = RedirectToLegacyLoginFactory(process.env.REACT_APP_API_URL, window);
 
 class App extends Component {
   constructor({ locale, dependencies: { sessionManager } }) {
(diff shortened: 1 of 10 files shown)


src/App.js
```diff
@@ -48,23 +50,28 @@ class App extends Component {
...
           <Router>
-            <div>
-              <Header userData={userData} />
-              <Route path="/" exact component={redirectToReports} />
-              <Route path="/reports/" exact component={Reports} />
-              <Footer />
-            </div>
+            <Switch>
+              <Route path="/" exact component={() => <Redirect to={{ pathname: '/reports' }} />} />
+              <PrivateRoute
```
fcoronelmakingsense: 👏

andresmoschini:
Fixed an issue in commit (_refactor: move private route to its own file_) and it became cleaner than before.

And added a test. 

Please, see the pending tests list and let me know if you have an idea about a better one.

andresmoschini:
uffff, it was hard, but I could test the router`s redirect!

Working in the other tests...

andresmoschini:
Well, I think that it is done, you have few minutes to review before I merge.

PR #80, opened 2019-03-22
add: reports box component

fcoronelmakingsense:
WIP
- Add report box component for visits reports
- Add Css styles to show box correctly
- Add method in datahub-client-double to get the visits for period

image


src/components/Reports/Reports.js
```diff
@@ -48,6 +51,10 @@ class Reports extends React.Component {
     this.setState({ pageSelected: pageFound });
   };
 
+  changePeriod = (id) => {
+    this.setState({ periodSelectedDays: id });
```
andresmoschini: I think that `id` is not the best name for a number of days.

src/components/Reports/ReportsBox/ReportsBox.js
```diff
@@ -0,0 +1,74 @@
...
+  constructor({ dependencies: { datahubClient } }) {
+    super();
+
+    /** @type { import('../../services/datahub-client').DatahubClient } */
+    this.datahubClient = datahubClient;
+
+    this.state = {
+      visits: 0,
+      dateFrom: null,
+      dateTo: new Date().toLocaleDateString(),
```
andresmoschini: What do you think about directly using dates in place of strings?

src/components/Reports/ReportsBox/ReportsBox.js
```diff
@@ -0,0 +1,74 @@
...
+  async getVisitsByPeriod(domainName, periodSelectedDays) {
+    let dateFrom = new Date();
+    dateFrom.setDate(dateFrom.getDate() - parseInt(periodSelectedDays));
+    const visits = await this.datahubClient.getVisitsByPeriod(domainName, dateFrom);
+    if (visits) {
+      this.setState({
+        visits: visits,
+        dateFrom: new Date(dateFrom).toLocaleDateString(),
+      });
+    }
```
andresmoschini: I think that you should always update it, even when there are no visits because in another way the results will not be related to current filters.

andresmoschini: Oh, `!visits` is not an empty list, is a null array, in that case, it should not happen, and if it happens it is an error. 

What do you think about showing the error in the box?

src/components/Reports/ReportsBox/ReportsBox.js
```diff
@@ -0,0 +1,74 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    this.getVisitsByPeriod(nextProps.domainName, nextProps.periodSelectedDays);
+  }
+
+  render() {
+    const {
+      state: { visits, dateFrom, dateTo },
+    } = this;
+    const isDataLoaded = !!visits;
```
andresmoschini: What do you think about adding a new value to the state about loading status?

Because, maybe `!visits` means something different than _loading status_, maybe an error.

src/components/Reports/ReportsBox/ReportsBox.js
```diff
@@ -0,0 +1,74 @@
...
+    return (
+      <>
+        {isDataLoaded ? (
+          <div className="reports-box--container">
+            <p className="title">{visits}</p>
+            <small className="text--uppercase">
+              <FormattedMessage id="reports_box.visits" />
+            </small>
+            <p className="small">
+              {dateFrom} a {dateTo}
```
andresmoschini: Here you should use _FormatDate_ helper to renderize the dates.

We should also translate the `a`

src/services/datahub-client.doubles.ts
```diff
@@ -44,4 +44,11 @@ export class HardcodedDatahubClient implements DatahubClient {
     }
     return domain.pages;
   }
+
+  public async getVisitsByPeriod(domainName: number, dateFrom: Date) {
+    console.log('getVisitsByPeriod', domainName, dateFrom);
```
andresmoschini: If you use an object, we will get better debug information:

```suggestion
    console.log('getVisitsByPeriod', { domainName, dateFrom });
```

is the same than:

```
    console.log('getVisitsByPeriod', { domainName: domainName, dateFrom: dateFrom });
```

src/components/Reports/ReportsBox/ReportsBox.css
```diff
@@ -0,0 +1,39 @@
...
+  height: 3px;
+  width: 127px;
+  background-color: #33AD73;
+}
+
+.reports-box--container .text--uppercase {
+  text-transform: uppercase;
+}
+
+
```
cbernat: extra space.

src/components/Reports/ReportsBox/ReportsBox.js
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+import { InjectAppServices } from '../../../services/pure-di';
+import { FormattedMessage } from 'react-intl';
+import './ReportsBox.css';
```
cbernat: Nice!

src/components/Reports/Reports.js
fcoronelmakingsense: yes i forgot to change. Thanks!

src/components/Reports/ReportsBox/ReportsBox.js
fcoronelmakingsense: Fixed

src/components/Reports/ReportsBox/ReportsBox.js
fcoronelmakingsense: changed

src/components/Reports/ReportsBox/ReportsBox.js
fcoronelmakingsense: fixed

src/components/Reports/ReportsBox/ReportsBox.js
fcoronelmakingsense: fixed

src/services/datahub-client.doubles.ts
fcoronelmakingsense: changed it

src/components/Reports/ReportsBox/ReportsBox.css
fcoronelmakingsense: fixed

fcoronelmakingsense:
Update PR

src/App.js
```diff
@@ -6,7 +6,7 @@ import Header from './components/Header/Header';
 import Footer from './components/Footer/Footer';
 import Reports from './components/Reports/Reports';
 import { InjectAppServices } from './services/pure-di';
-import Loading from './components/Loading';
+import Loading from './components/Loading/Loading';
```
andresmoschini: 👍

src/components/Reports/ReportsBox/ReportsBox.js
```diff
@@ -0,0 +1,72 @@
...
+        {isLoading ? (
+          <Loading />
+        ) : (
+          <div className="reports-box--container">
+            <p className="title">{visits}</p>
+            <small className="text--uppercase">
+              <FormattedMessage id="reports_box.visits" />
+            </small>
+            <p className="small">
+              {dateFrom.toLocaleDateString()} <FormattedMessage id="reports_box.to" />{' '}
```
andresmoschini:
```suggestion
              <FormattedDate value={dateFrom} timeZone="UTC" />  <FormattedMessage id="reports_box.to" />{' '}
```

src/components/Reports/ReportsBox/ReportsBox.js
```diff
@@ -0,0 +1,72 @@
...
+              {dateTo.toLocaleDateString()}
```
andresmoschini:
```suggestion
              <FormattedDate value={dateTo} timeZone="UTC" />
```

src/components/Reports/ReportsBox/ReportsBox.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  async getVisitsByPeriod(domainName, periodSelectedDays) {
+    let dateFrom = new Date();
+    dateFrom.setDate(dateFrom.getDate() - parseInt(periodSelectedDays));
+
+    const visits = await this.datahubClient.getVisitsByPeriod(domainName, dateFrom);
+    this.setState({
+      visits: visits,
+      dateFrom: new Date(dateFrom),
```
andresmoschini: Should you also update `dateTo`?

fcoronelmakingsense: the day of today never change. or i missing something?

andresmoschini: Imagine that you are at midnight: the day could change. 

Also, if I am not wrong, you are using complete dates (with time, I mean, you are not cleaning it), so, eventually, we would decide to show also the time.

Summarizing, visits has the count of visits between _dateFrom_ and the time of the request (in place of the first rendering _dateTo_).

Does it make sense?

src/components/Reports/ReportsBox/ReportsBox.js
fcoronelmakingsense: changed it

src/components/Reports/ReportsBox/ReportsBox.js
fcoronelmakingsense: changed it

src/components/Reports/ReportsBox/ReportsBox.js
fcoronelmakingsense: yes, changed it

andresmoschini:
:shipit:

PR #82, opened 2019-03-23
chore: avoid publishing prod build to cdn

andresmoschini:
Hi @fgchaio and @swidzinskiMS,

Since our Doppler backend is configured to only accept a CORS origin, and we are moving this app to his final domain, it will not be possible to continue using our CDN versions against Legacy Doppler in production. Am I missing something?

In the future, we could continue taking advantage of the CDN if our docker image only contains an HTML pointing to the right CDN build as I explain in a comment in the code. But it is only an idea.


andresmoschini:
Finally, I think that I found a solution to allow us to keep it working. So, I am closing this PR.

PR #84, opened 2019-03-25
Reports visits withmail

fcoronelmakingsense:
- Add property to distinguish visits with email to without
- Add styles to show in the same row
- Fix the texts

image

### DIFF
--- a/src/i18n/en.json
+++ b/src/i18n/en.json
@@ -33,9 +33,10 @@
   },
   "reports_title": "Doppler | Reports",
   "reports_box": {
-    "visits": "Total Visits",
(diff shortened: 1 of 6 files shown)


src/i18n/en.json
```diff
@@ -33,9 +33,10 @@
   },
   "reports_title": "Doppler | Reports",
   "reports_box": {
-    "visits": "Total Visits",
+    "visits": "Visits with emails",
```
andresmoschini: What do you think about renaming this key?

fcoronelmakingsense: i asumed this has with emails, but i can change 😃

PR #85, opened 2019-03-26
Add signup page

andresmoschini:
Hi team,

This signup page has a lot of holes yet, but I think that it could be a nice start to continue working on related things in parallel, for example, validations, tests, backend, usability, etc.

Could you review?

I would really appreciate if HTML/CSS guys (@GusBaamonde, @IgnacioRodrigues, @damianmuti) take a look and give me feedback about commit _"chore: add form helpers following our conventions"_ (probably _"chore: add formik with basic validation"_ could help to understand), because I want to know if the structure makes sense to be reused.

~~You can see it working at https://cdn.fromdoppler.com/doppler-webapp/build213/#/signup~~

**UPDATE:** It is updated with the last @IgnacioRodrigues fixes and with a PasswordFieldItem component that allows show and hide the password. 

See it working in: https://cdn.fromdoppler.com/doppler-webapp/build223/#/signup

### DIFF
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -2,7 +2,8 @@ import React from 'react';
 import { Link } from 'react-router-dom';
 import { FormattedHTMLMessage, injectIntl } from 'react-intl';
 import { timeout } from '../../utils';
-import { Formik, Form, Field } from 'formik';
+import { Formik, Form } from 'formik';
+import { FieldGroup, InputFieldItem, CheckboxFieldItem } from '../form-helpers/form-helpers';
 
 const fieldNames = {
   firstname: 'firstname',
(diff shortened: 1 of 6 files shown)


andresmoschini:
You can see it working at https://cdn.fromdoppler.com/doppler-webapp/build213/#/signup

src/components/Signup/Signup.js
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    if (!values[fieldNames.lastname]) {
+      errors[fieldNames.lastname] = _('validation_messages.error_required_field');
+    }
+
+    if (!values[fieldNames.phone]) {
+      errors[fieldNames.phone] = _('validation_messages.error_required_field');
+    } else {
+      // TODO: validate phone
```
fcoronelmakingsense: maybe can use REGEX from doppler app phone validation

src/components/Signup/Signup.js
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    if (!values[fieldNames.email]) {
+      errors[fieldNames.email] = _('validation_messages.error_required_field');
+    } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(values[fieldNames.email])) {
+      // TODO: review if this is our desired validation
```
fcoronelmakingsense: use the same as doppler not?

src/components/Signup/Signup.js
```diff
@@ -0,0 +1,175 @@
...
+ * It is required because in another way, the fields are not marked as touched.
+ */
+const getFormInitialValues = () =>
+  Object.keys(fieldNames).reduce(
+    (accumulator, currentValue) => ({ ...accumulator, [currentValue]: '' }),
+    {},
+  );
+
+export default injectIntl(function({ intl }) {
+  const _ = (id, values) => intl.formatMessage({ id: id }, values);
```
fcoronelmakingsense: this can add in a only place and can to reuse?

src/components/form-helpers/form-helpers.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
```
fcoronelmakingsense: The idea of this file is can reuse in all app?

andresmoschini: Yes, at least in signup, login and that kind of pages.

src/components/Signup/Signup.js
andresmoschini: Yes, but I do not know how to do it yet.

One possible approach could be injecting `_` dependency in our _PureCI_ container. But we would couple IntlProvider with our dependency system. Since it is a single line, I prefer to keep it as it is while we think about how to reuse it.

src/components/Signup/Signup.js
andresmoschini: Good catch! No, I do not think so. It should. 

I will try to fix in this PR or at least adding a TODO.

src/components/Signup/Signup.js
andresmoschini: No, Doppler uses three fields in place of one.

On the other hand, we want to use the same library that we are already using in sites to validate phone numbers, so, I think that it could be done as a different task.

src/components/Signup/Signup.js
andresmoschini: Oh, sorry, the TODO is already there.

src/components/Signup/Signup.js
fcoronelmakingsense: oh i didn't know in sites already using one

andresmoschini: Oh, yes, and it is great!. @IgnacioRodrigues do you have a link to try the phone validation?

IgnacioRodrigues: Yes, you can look at the following form in the field phone https://m.fromdoppler.com/contacto/

andresmoschini:
It is updated with the last @IgnacioRodrigues fixes and with a PasswordFieldItem component that allows show and hide the password. 

See it working in: https://cdn.fromdoppler.com/doppler-webapp/build223/#/signup

PR #88, opened 2019-03-26
test(reports): configure dependencies at container level for reports tests

andresmoschini:
Hi @fcoronelmakingsense, 

I cannot understand how the tests are passing in our CI environment, but there was an issue reproduced in my local environment because _ReportBox_ was trying to use a dependency that was only set for _Reports_.

This should fix the issue, what do you think?

### DIFF
--- a/src/components/Reports/Reports.test.js
+++ b/src/components/Reports/Reports.test.js
@@ -3,6 +3,7 @@ import { render, cleanup, wait } from 'react-testing-library';
 import 'jest-dom/extend-expect';
 import DopplerIntlProvider from '../../i18n/DopplerIntlProvider.double-with-ids-as-values';
 import Reports from './Reports';
+import { AppServicesProvider } from '../../services/pure-di';
 
 const verifiedDateAsEngString = '12/17/2017';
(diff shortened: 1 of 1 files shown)


andresmoschini:
> i didn't know this. When i create the pr test was passing. 

Yes, and they are still passing, but not in my environment  🤷‍♂️

> Always need `<AppServicesProvider>` when use datahubClient dependecy?

No, you still can inject the dependency directly in some cases. But in this case, you were injecting the dependency in the parent component (_Reports_), but the child component (_ReportBox_) was trying to use it. Does it make sense?

fcoronelmakingsense:
> > i didn't know this. When i create the pr test was passing.
> 
> Yes, and they are still passing, but not in my environment 🤷‍♂️
> 
> > Always need `<AppServicesProvider>` when use datahubClient dependecy?
> 
> No, you still can inject the dependency directly in some cases. But in this case, you were injecting the dependency in the parent component (_Reports_), but the child component (_ReportBox_) was trying to use it. Does it make sense?

Yes you need because the child use it

PR #89, opened 2019-03-26
chore(dependencies): upgrade all dependencies

andresmoschini:
Well, just that. I upgraded all the dependencies.

It is motivated in that [React Router released V5](https://reacttraining.com/blog/react-router-v5/), it has not broken changes, but it works better with our React version and fixes important issues.


fcoronelmakingsense:
Yes i think is better upgrade dependencies when has not broken changes and it's better now when starting with the project and still small

PR #92, opened 2019-03-27
add: test in reports page ranking

fcoronelmakingsense:
- Add 2 test in page ranking reports

### DIFF
--- a/src/components/Reports/ReportsPageRanking/ReportsPageRanking.test.js
+++ b/src/components/Reports/ReportsPageRanking/ReportsPageRanking.test.js
@@ -1,0 +1,68 @@
+import React from 'react';
+import { render, cleanup, wait } from 'react-testing-library';
+import 'jest-dom/extend-expect';
+import DopplerIntlProvider from '../../../i18n/DopplerIntlProvider.double-with-ids-as-values';
+import ReportsPageRanking from './ReportsPageRanking';
+import { AppServicesProvider } from '../../../services/pure-di';
+
+const fakeDate = new Date();
+
+const domain = 'www.fromdoppler.com';
+
+const fakePagesData = [
+  {
+    name: 'https://www.fromdoppler.com/email-marketing',
+    totalVisits: 10122,
+  },
+  {
+    name: 'https://www.fromdoppler.com/precios',
+    totalVisits: 9000,
+  },
+  {
+    name: 'https://www.fromdoppler.com/login',
+    totalVisits: 5001,
+  },
+];
+
+describe('Reports pages ranking', () => {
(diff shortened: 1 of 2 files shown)


src/components/Reports/ReportsPageRanking/ReportsPageRanking.test.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import { render, cleanup, wait } from 'react-testing-library';
+import 'jest-dom/extend-expect';
+import DopplerIntlProvider from '../../../i18n/DopplerIntlProvider.double-with-ids-as-values';
+import ReportsPageRanking from './ReportsPageRanking';
+import { AppServicesProvider } from '../../../services/pure-di';
+
+const dateTo = new Date();
+const dateFrom = new Date();
```
mvillaseco: I don't undestand if this dates are being modified or now, if not, only create one, if they are being modified, maybe should you use let?

fcoronelmakingsense: Yes i only create two to diferenciate but i can use one

fcoronelmakingsense:
Update Pr: fix duplicate date. Fix tests in general reports

src/components/Reports/ReportsPageRanking/ReportsPageRanking.test.js
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+    const { getByText } = render(
+      <AppServicesProvider forcedServices={{ datahubClient: datahubClientDouble }}>
+        <DopplerIntlProvider>
+          <ReportsPageRanking
+            domainName={domain}
+            dateTo={fakeDate}
+            dateFrom={fakeDate}
+            dependencies={{ datahubClient: datahubClientDouble }}
```
andresmoschini: This property (`dependencies`) should not be defined here because you are already defining it in `AppServicesProvider`.

src/components/Reports/ReportsPageRanking/ReportsPageRanking.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+  it('render component without pages', () => {
+    const datahubClientDouble = {
+      getPagesRankingByPeriod: async () => [],
+    };
+
+    render(
+      <AppServicesProvider forcedServices={{ datahubClient: datahubClientDouble }}>
+        <DopplerIntlProvider>
+          <ReportsPageRanking dependencies={{ datahubClient: datahubClientDouble }} />
```
andresmoschini: Again, `dependencies` property is not necessary.

PR #98, opened 2019-04-03
fix: disabled all pages dropdown

fcoronelmakingsense:
- Disabled dropdown all pages until implemented in datahub

image


andresmoschini:
👍

PR #99, opened 2019-04-05
add: change language by url parameter

fcoronelmakingsense:
- Add query-string library to manage and parse query string from location
https://github.com/sindresorhus/query-string

- Change languague by url `lang` parameter

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,13 +1,14 @@
 import React, { Component } from 'react';
 import './App.css';
 import DopplerIntlProvider from './i18n/DopplerIntlProvider';
-import { Route, Redirect, Switch } from 'react-router-dom';
+import { Route, Redirect, Switch, withRouter } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -26,21 +27,31 @@ class App extends Component {
 
   componentDidMount() {
     this.sessionManager.initialize(this.updateSession);
+    this.setUrlLanguage(this.state.dopplerSession);
```
andresmoschini: Parameter is not necessary, right?

```suggestion
    this.setUrlLanguage();
```

andresmoschini:
Nice! And, only to know, how does it work with authenticated pages?

fcoronelmakingsense:
> Nice! And, only to know, how does it work with authenticated pages?

for now, use user language in this cases

src/App.js
fcoronelmakingsense: Thanks! Fixed

PR #101, opened 2019-04-06
Fix build

andresmoschini:
It seems in an update of Jest or React-Scripts, something happens with the command `react-scripts test` and it was not returning an error code if `CI` environment variable was not true.

It fixes the issue. Now, some tests will fail, I am fixing them in this same PR.


andresmoschini:
Well, I think that it is done. Team, could you kindly review it?

src/components/Login/Login.js
```diff
@@ -117,9 +117,12 @@ class Login extends React.Component {
                 >
                   {this._('login.button_login')}
                 </button>
+                {/*
+                // TODO: implement forgot password
                 <a href={this._('login.forgot_password_url')} className="forgot-link">
                   {this._('login.forgot_password')}
                 </a>
+                */}
```
andresmoschini: Hi @fcoronelmakingsense, I have hidden this input because if I am not wrong forgot password is sadly not ready.

I think that we could do two things in order to have this Login page closer to be finished:

* If Doppler has a URL for forgot password, fix it using the right link.
* If Doppler has forgot password as part of login page and we cannot point to it with a URL, create an empty Forgot password on webapp side, and here, link to there.

PR #102, opened 2019-04-08
add: change language in login

fcoronelmakingsense:
- Refactor change language method
- Add html in login
- TODO add the final styles

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -80,15 +80,36 @@ class Login extends React.Component {
         <article className="main-panel">
           <header>
             <h1 className="logo-doppler-new">Doppler</h1>
-            <small className="content-signin">
-              {this._('login.you_want_create_account')}{' '}
-              <Link to="/signup" className="link-green uppercase">
-                {this._('login.signup')}
-              </Link>
-            </small>
+            <div className="language-selector">
+              {this.intl.locale === 'es' ? (
+                <div className="option" id="spanish-selector">
+                  <span>
+                    <strong>ES</strong>
+                  </span>
+                  <Link to="?lang=en" className="option--item">
+                    EN
+                  </Link>
+                </div>
+              ) : (
+                <div className="option" id="spanish-selector">
+                  <span>
+                    <strong>EN</strong>
+                  </span>
+                  <Link to="?lang=es" className="option--item">
+                    ES
+                  </Link>
+                </div>
+              )}
+            </div>
           </header>
           <h5>{this._('login.enter_doppler')}</h5>
           <p className="content-subtitle">{this._('login.enter_doppler_sub')}</p>
+          <p className="content-subtitle">
+            {this._('login.you_want_create_account')}{' '}
+            <Link to="/signup" className="link-green uppercase">
+              {this._('login.signup')}
+            </Link>
+          </p>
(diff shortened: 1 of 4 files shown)


src/components/Login/Login.js
```diff
@@ -80,15 +80,36 @@ class Login extends React.Component {
...
-                {this._('login.signup')}
-              </Link>
-            </small>
+            <div className="language-selector">
+              {this.intl.locale === 'es' ? (
+                <div className="option" id="spanish-selector">
+                  <span>
+                    <strong>ES</strong>
+                  </span>
+                  <Link to="/login?lang=en" className="link-green uppercase">
```
andresmoschini: Is it not possible making it agnostic to current page to allow us a more easily refactoring?

For example something like:

```suggestion
                  <Link to="?lang=en" className="link-green uppercase">
```

andresmoschini: `link-green` class is still necessary?

```suggestion
                  <Link to="/login?lang=en" className="uppercase">
```

src/i18n/DopplerIntlProvider.js
```diff
@@ -14,7 +14,11 @@ const messages = {
 addLocaleData([...en, ...es]);
 
 export default ({ locale, children }) => (
-  <IntlProvider locale={locale || 'en'} messages={flattenMessages(messages[locale || 'en'])}>
+  <IntlProvider
+    key={locale}
```
andresmoschini:
```suggestion
    key={locale || 'en'}
```

Like in `locale` property?

src/components/Login/Login.js
```diff
@@ -80,15 +80,36 @@ class Login extends React.Component {
         <article className="main-panel">
           <header>
             <h1 className="logo-doppler-new">Doppler</h1>
-            <small className="content-signin">
-              {this._('login.you_want_create_account')}{' '}
-              <Link to="/signup" className="link-green uppercase">
-                {this._('login.signup')}
-              </Link>
-            </small>
+            <div className="language-selector">
```
andresmoschini: What do you think about also apply it in Signup page and also refactoring it as a component?

fcoronelmakingsense: you talk always small tasks and small changes 🤣  but i can

andresmoschini: You can do it in another PR, no problem.

Just take into account that the task of "select language in non-authenticated pages" will be only completed if all non-authenticated pages have this option.

src/i18n/DopplerIntlProvider.js
fcoronelmakingsense: fixed

src/components/Login/Login.js
fcoronelmakingsense: fixed

fcoronelmakingsense: fixed

andresmoschini: Please, test it, I am not sure.

andresmoschini: Wow, and it works?

fcoronelmakingsense: yes!!

fcoronelmakingsense: yes it's works!

PR #105, opened 2019-04-08
Intl tel input

andresmoschini:
Well, it is ready, maybe it is pending only applying some better styles.

Please, consider review it commit by commit. I have added some `// TODO` comments for things that I like to fix in the future. I have also added some clarificatory comments in places where I had to do some weird things.

Could you review? 

You can play with this in: https://cdn.fromdoppler.com/doppler-webapp/int-build276/#/signup


src/components/form-helpers/form-helpers.css
```diff
@@ -0,0 +1,11 @@
+.intl-tel-input {
+  position: relative;
+  display: inline-block;
+  width: 100%;
+}
+
+.intl-tel-input .country-list {
+  font-family: proxima-nova,Helvetica;
```
cbernat: `proxima-nova, Helvetica` spacing

cbernat:
👍

PR #107, opened 2019-04-09
add: forgot password

fcoronelmakingsense:
- Add new page and new route
- Add texts
- TODO add final change language styles.

image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -8,6 +8,7 @@ import { InjectAppServices } from './services/pure-di';
 import Loading from './components/Loading/Loading';
 import Login from './components/Login/Login';
 import Signup from './components/Signup/Signup';
+import ForgotPassword from './components/ForgotPassword/ForgotPassword';
 import queryString from 'query-string';
 
 class App extends Component {
@@ -89,6 +90,7 @@ class App extends Component {
             />
             <Route path="/login/" exact component={Login} />
             <Route path="/signup/" exact component={Signup} />
+            <Route path="/forgot-password/" exact component={ForgotPassword} />
             {/* TODO: Implement NotFound page in place of redirect all to reports */}
             {/* <Route component={NotFound} /> */}
             <Route component={() => <Redirect to={{ pathname: '/reports' }} />} />
--- a/src/components/ForgotPassword/ForgotPassword.js
+++ b/src/components/ForgotPassword/ForgotPassword.js
@@ -1,0 +1,121 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { injectIntl } from 'react-intl';
+import { timeout } from '../../utils';
+import { Formik, Form } from 'formik';
+import { FieldGroup, InputFieldItem } from '../form-helpers/form-helpers';
+
+const fieldNames = {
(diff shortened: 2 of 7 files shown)


andresmoschini:
Remember adding the links to this page.

src/App.js
```diff
@@ -79,6 +80,7 @@ class App extends Component {
             />
             <Route path="/login/" exact component={Login} />
             <Route path="/signup/" exact component={Signup} />
+            <Route path="/forgotpassword/" exact component={ForgotPassword} />
```
andresmoschini: What do you think about

```suggestion
            <Route path="/forgot-password/" exact component={ForgotPassword} />
```

src/components/ForgotPassword/ForgotPassword.js
```diff
@@ -0,0 +1,122 @@
...
+    await timeout(1500);
+    setSubmitting(false);
+  };
+
+  return (
+    <main className="panel-wrapper">
+      <article className="main-panel">
+        <header>
+          <h1 className="logo-doppler-new">Doppler</h1>
+          <div className="language-selector">
```
andresmoschini: Maybe in another PR, but, what do you think about moving this code into a component? (We already meet [the rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)))

src/components/ForgotPassword/ForgotPassword.js
```diff
@@ -0,0 +1,122 @@
...
+const ForgotPassword = ({ intl }) => {
+
+  const _ = (id, values) => intl.formatMessage({ id: id }, values);
+
+  const validate = (values) => {
+    const errors = {};
+
+    if (!values[fieldNames.email]) {
+      errors[fieldNames.email] = _('validation_messages.error_required_field');
+    } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(values[fieldNames.email])) {
```
cbernat: Maybe add this regex in a more reusable place? you have password validation in signup and here.

andresmoschini: You are right, although this validation is still pending, there is a task to deal with these things: [DBR-135](http://jira.makingsense.com/browse/DBR-135)

fcoronelmakingsense: Yes, but there is another task for validations

src/components/ForgotPassword/ForgotPassword.js
fcoronelmakingsense: Yes i think it's better

andresmoschini:
> TODO add final change language styles.

If you are being blocked for this, do not worry, follow the same approach that you followed in the other pages, and then, @damianmuti could fix the HTML along with https://github.com/FromDoppler/Doppler-UI-System/pull/81

src/App.js
fcoronelmakingsense: changed

fcoronelmakingsense:
TODO @damianmuti apply language selector

PR #108, opened 2019-04-10
add: email validations

fcoronelmakingsense:
- Add new validations file with common validations
- Talk with @mvillaseco and update Email validation regex
- Apply validation in signup, login and forgot password

### DIFF
--- a/src/components/ForgotPassword/ForgotPassword.js
+++ b/src/components/ForgotPassword/ForgotPassword.js
@@ -4,6 +4,7 @@ import { injectIntl } from 'react-intl';
 import { timeout } from '../../utils';
 import { Formik, Form } from 'formik';
 import { FieldGroup, InputFieldItem } from '../form-helpers/form-helpers';
+import { validateEmail, validateRequiredField } from '../../validations';
 import LanguageSelector from '../shared/LanguageSelector/LanguageSelector';
 
 const fieldNames = {
@@ -25,11 +26,10 @@ const ForgotPassword = ({ intl }) => {
   const validate = (values) => {
     const errors = {};
 
-    if (!values[fieldNames.email]) {
-      errors[fieldNames.email] = _('validation_messages.error_required_field');
-    } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(values[fieldNames.email])) {
-      // TODO: review if this is our desired validation
-      errors[fieldNames.email] = _('validation_messages.error_invalid_email_address');
+    const emailMsgError =
+      validateRequiredField(values[fieldNames.email]) || validateEmail(values[fieldNames.email]);
+    if (emailMsgError) {
+      errors[fieldNames.email] = _(emailMsgError);
     }
 
     return errors;
@@ -64,7 +64,7 @@ const ForgotPassword = ({ intl }) => {
                 <InputFieldItem
                   fieldName={fieldNames.email}
                   label={_('signup.label_email')}
-                  type="email"
+                  type="text"
                   placeholder={_('signup.placeholder_email')}
                 />
               </FieldGroup>
@@ -74,7 +74,7 @@ const ForgotPassword = ({ intl }) => {
                 {_('login.button_login')}
               </button>
               <Link to="/login" className="forgot-link">
-                <span class="triangle-right" />
+                <span className="triangle-right" />
                 {_('forgot_password.back_login')}
               </Link>
             </fieldset>
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -4,6 +4,7 @@ import { injectIntl } from 'react-intl';
 import { timeout } from '../../utils';
 import { Formik, Form } from 'formik';
 import { FieldGroup, InputFieldItem, PasswordFieldItem } from '../form-helpers/form-helpers';
+import { validateEmail, validateRequiredField } from '../../validations';
 import LanguageSelector from '../shared/LanguageSelector/LanguageSelector';
 
 const fieldNames = {
@@ -33,11 +34,10 @@ const Login = ({ intl, location }) => {
   const validate = (values) => {
     const errors = {};
 
-    if (!values[fieldNames.user]) {
-      errors[fieldNames.user] = _('validation_messages.error_required_field');
-    } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(values[fieldNames.user])) {
-      // TODO: review if this is our desired validation
-      errors[fieldNames.user] = _('validation_messages.error_invalid_email_address');
+    const emailMsgError =
+      validateRequiredField(values[fieldNames.user]) || validateEmail(values[fieldNames.user]);
+    if (emailMsgError) {
+      errors[fieldNames.user] = _(emailMsgError);
     }
 
     if (!values[fieldNames.password]) {
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -10,6 +10,7 @@ import {
   PasswordFieldItem,
   PhoneFieldItem,
 } from '../form-helpers/form-helpers';
+import { validateEmail, validateRequiredField } from '../../validations';
 import LanguageSelector from '../shared/LanguageSelector/LanguageSelector';
 
 const fieldNames = {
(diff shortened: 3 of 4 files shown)


cbernat:
👍 Great! I will add my validations in that file too.

src/components/ForgotPassword/ForgotPassword.js
```diff
@@ -24,12 +25,8 @@ const ForgotPassword = ({ intl }) => {
...
     const errors = {};
 
-    if (!values[fieldNames.email]) {
-      errors[fieldNames.email] = _('validation_messages.error_required_field');
-    } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(values[fieldNames.email])) {
-      // TODO: review if this is our desired validation
-      errors[fieldNames.email] = _('validation_messages.error_invalid_email_address');
-    }
+    const emailMsgError = validateEmail(values[fieldNames.email]);
+    errors[fieldNames.email] = emailMsgError ? _(emailMsgError) : '';
```
andresmoschini: I think that it is better not adding the property when there is no error, what do you think about:

```suggestion
    if (emailMsgError) {
        errors[fieldNames.email] = _(emailMsgError);
    }
```

src/components/ForgotPassword/ForgotPassword.js
```diff
@@ -97,7 +94,7 @@ const ForgotPassword = ({ intl }) => {
                 {_('login.button_login')}
               </button>
               <Link to="/login" className="forgot-link">
-                <span class="triangle-right" />
+                <span className="triangle-right" />
```
andresmoschini: 👍

src/validations.js
```diff
@@ -0,0 +1,11 @@
+export function validateEmail(email) {
+  if (!email) {
+    return 'validation_messages.error_required_field';
```
andresmoschini: I would move this validation outside of _validateEmail_, in my opinion, it is not related. 

Imagine a scenario where the email field is optional.

src/validations.js
```diff
@@ -0,0 +1,11 @@
...
+    // eslint-disable-next-line
```
andresmoschini: Why do you need to disable eslint validation?

andresmoschini: Oh, I see:

image

But, ESLint is right, those escapes are not necessary because, for example, `.` does not require escape when it is inside `[. . .]`.

This regex seems to be equivalent: `^([a-z\d[\]])([\w-.+\]\[]*)@([ñ\w.-\]\[]+)\.[\w-\]\[.]{2,}(\?)?.*$`

src/validations.js
```diff
@@ -0,0 +1,11 @@
...
+  } else if (
+    !/^([a-z\d\[\]])([\w\-\.\+\]\[]*)@([ñ\w\.\-\]\[]+)\.[\w\-\]\[\.]{2,}(\?)?.*$/i.test(email)
```
andresmoschini:
```suggestion
    !/^([a-z\d[\]])([\w-.+\]\[]*)@([ñ\w.-\]\[]+)\.[\w-\]\[.]{2,}(\?)?.*$/i.test(email)
```

fcoronelmakingsense: Fixed

src/validations.js
fcoronelmakingsense: Fixed

src/components/ForgotPassword/ForgotPassword.js
fcoronelmakingsense: Yes i changed

src/components/Login/Login.js
```diff
@@ -32,12 +33,8 @@ const Login = ({ intl, location }) => {
...
     const errors = {};
 
-    if (!values[fieldNames.user]) {
-      errors[fieldNames.user] = _('validation_messages.error_required_field');
-    } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(values[fieldNames.user])) {
-      // TODO: review if this is our desired validation
-      errors[fieldNames.user] = _('validation_messages.error_invalid_email_address');
-    }
+    const emailMsgError = validateEmail(values[fieldNames.user]);
+    errors[fieldNames.user] = emailMsgError ? _(emailMsgError) : '';
```
andresmoschini: What dop you think about applying the same change here?

```js
    const emailMsgError = validateEmail(values[fieldNames.user]);
    if (emailMsgError) {
      errors[fieldNames.user] = _(emailMsgError);
    }
```

src/components/Signup/Signup.js
```diff
@@ -66,12 +67,8 @@ export default injectIntl(function({ intl }) {
...
     }
 
-    if (!values[fieldNames.email]) {
-      errors[fieldNames.email] = _('validation_messages.error_required_field');
-    } else if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(values[fieldNames.email])) {
-      // TODO: review if this is our desired validation
-      errors[fieldNames.email] = _('validation_messages.error_invalid_email_address');
-    }
+    const emailMsgError = validateEmail(values[fieldNames.email]);
+    errors[fieldNames.email] = emailMsgError ? _(emailMsgError) : '';
```
andresmoschini: And here.

src/validations.js
andresmoschini: ❤️
Thanks! my expression was wrong, sorry.

fcoronelmakingsense:
Update Pr

src/components/Login/Login.js
```diff
@@ -33,11 +34,14 @@ const Login = ({ intl, location }) => {
...
-      // TODO: review if this is our desired validation
-      errors[fieldNames.user] = _('validation_messages.error_invalid_email_address');
+    let emailMsgError = validateRequiredField(values[fieldNames.user]);
+    if (emailMsgError) {
+      errors[fieldNames.user] = _(emailMsgError);
+    } else {
+      emailMsgError = validateEmail(values[fieldNames.user]);
+      if (emailMsgError) {
+        errors[fieldNames.user] = _(emailMsgError);
+      }
```
andresmoschini: Mmm, in general if inside if are difficult to be read, what do you think about this:

```js
const emailMsgError = validateRequiredField(values[fieldNames.user]) || validateEmail(values[fieldNames.user]);
if (emailMsgError) {
  errors[fieldNames.user] = _(emailMsgError);
}
```

At the beginning, it could be difficult to be read, but at the second look (and the rest ones) it is not.

src/validations.js
```diff
@@ -0,0 +1,12 @@
+export function validateEmail(email) {
+  if (!email) {
+    return 'validation_messages.error_required_field';
```
andresmoschini: Remember remove _required validation_ from `validateEmail`

src/validations.js
```diff
@@ -0,0 +1,12 @@
+export function validateEmail(email) {
```
andresmoschini:
```suggestion
export function validateEmail(value) {
```

andresmoschini: What do you think about calling it `value`?

src/validations.js
```diff
@@ -0,0 +1,12 @@
...
+  if (!email) {
+    return 'validation_messages.error_required_field';
+  } else if (!/^([a-z\d[\]])([\w\-.+\][]*)@([ñ\w.\-\][]+)\.[\w\-\][.]{2,}(\?)?.*$/i.test(email)) {
+    return 'validation_messages.error_invalid_email_address';
+  }
+  return null;
+}
+
+export function validateRequiredField(field) {
```
andresmoschini: What do you think about calling it `value`?

```suggestion
export function validateRequiredField(value) {
```

src/validations.js
```diff
@@ -0,0 +1,12 @@
...
+  return !field ? 'validation_messages.error_required_field' : false;
```
andresmoschini: If I am not wrong, it is not compatible with boolean values, what do you think about: 

```suggestion
  return value !== undefined && value !== null && value !== '';
```


```suggestion
  return !value && value !== false;
```

src/validations.js
fcoronelmakingsense: yes, it's more descriptive

src/components/Login/Login.js
fcoronelmakingsense: yes, good change!

src/validations.js
fcoronelmakingsense: changed

PR #110, opened 2019-04-11
DBR-153 Language selector component

damianmuti:
Created a simple component to handle language selector.

### DIFF
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -100,7 +100,7 @@ export default injectIntl(function({ intl }) {
       <article className="main-panel">
         <header>
           <h1 className="logo-doppler-new">Doppler</h1>
-          <LanguageSelector lang={intl.locale} />
+          <LanguageSelector />
         </header>
         <h5>{_('signup.sign_up')}</h5>
         <p className="content-subtitle">{_('signup.sign_up_sub')}</p>
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -10,6 +10,7 @@ import {
   PasswordFieldItem,
   PhoneFieldItem,
 } from '../form-helpers/form-helpers';
+import LanguageSelector from '../shared/LanguageSelector/LanguageSelector';
 
 const fieldNames = {
   firstname: 'firstname',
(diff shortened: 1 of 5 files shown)


src/components/Signup/Signup.js
```diff
@@ -99,27 +100,7 @@ export default injectIntl(function({ intl }) {
...
-                <span>
-                  <strong>EN</strong>
-                </span>
-                <Link to="?lang=es" className="option--item">
-                  ES
-                </Link>
-              </div>
-            )}
-          </div>
+          <LanguageSelector lang={intl.locale} />
```
andresmoschini: I am not sure, is just an opinion, what do you think calling it `currentLang`?

src/components/shared/LanguageSelector/LanguageSelector.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+
+const LanguageSelector = ({ lang }) => {
```
andresmoschini: We could inject current lang in place of receiving it as a parameter. But I am not sure if it is a good idea, @fcoronelmakingsense, @damianmuti  what do you think?

```
/**
 * LanguageSelector
 * @param { Object } props
 * @param { import('react-intl').InjectedIntl } props.intl
 */
const LanguageSelector = ({ intl }) => {
  const lang = intl.locale;
  return (
    <div className="language-selector">
    // . . .
    </div>
  );
};
export default injectIntl(LanguageSelector);
```

src/components/Signup/Signup.js
damianmuti: I like that approach, @andresmoschini . Will do the proper changes 👍

src/components/shared/LanguageSelector/LanguageSelector.js
fcoronelmakingsense: always has the current language, and that is how we use in other pages. maybe is better this

damianmuti: This was my first thought when I've created this component, but my React knowledge isn't the best tho, so I couldn't get around it.
Thanks for the help, @andresmoschini ! I got it working this way, and I guess it's a good idea not to not have the locale as a parameter. 
@fcoronelmakingsense do you agree? If it's ok with you guys, I'll push the changes right away.

andresmoschini: 🚀

damianmuti: I'll update this PR right away 👍

andresmoschini:
:shipit:

PR #112, opened 2019-04-11
fix: add missing behavior when domain list is empty

andresmoschini:
Hi team, 

I fixed this scenario:

image

I am creating a new task to improve it in the future.

### DIFF
--- a/src/components/Reports/Reports.js
+++ b/src/components/Reports/Reports.js
@@ -3,22 +3,23 @@ import ReportsFilters from './ReportsFilters/ReportsFilters';
 import ReportsBox from './ReportsBox/ReportsBox';
 import ReportsPageRanking from './ReportsPageRanking/ReportsPageRanking';
 import { InjectAppServices } from '../../services/pure-di';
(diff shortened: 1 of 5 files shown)


src/components/Reports/Reports.js
```diff
@@ -32,20 +33,18 @@ class Reports extends React.Component {
...
-      this.setState({
-        domains: domains,
-        domainSelected: domainSelected,
-        pages: pages,
-        pageSelected: pageSelected,
-        dateFrom: dateFrom,
-      });
-    }
+    const domainSelected = domains.length ? domains[0] : null;
+    const pages = [];
```
fcoronelmakingsense: no need and if here? if doesn't has domains not pages too..

andresmoschini: I am thinking about removing all pages related code by the moment...

andresmoschini: I think that this solution is ugly, but safe regarding to that, so, by the moment I will keep it as it is.

Thanks @fcoronelmakingsense !

PR #113, opened 2019-04-11
Dplr-163: password form validation

cbernat:
Added password validation for signup only. 
I thought to add validations in validations file when ready. 
Is not ready for merge because it's pending to add this validation in forgot password.

Not touched
image

Completed
image

Autocompleted
image

Failed
image
image

Could you please review? Maybe some things are not as you are used to work. @andresmoschini @fcoronelmakingsense 

--- 

https://cdn.fromdoppler.com/doppler-webapp/int-build316/#/signup

### DIFF
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -7,7 +7,7 @@ import {
   FieldGroup,
   InputFieldItem,
   CheckboxFieldItem,
-  PasswordValidatedFieldItem,
+  ValidatedPasswordFieldItem,
   PhoneFieldItem,
 } from '../form-helpers/form-helpers';
 import { validateEmail, validateRequiredField } from '../../validations';
@@ -149,7 +149,7 @@ export default injectIntl(function({ intl }) {
                   type="email"
                   placeholder={_('signup.placeholder_email')}
                 />
-                <PasswordValidatedFieldItem
+                <ValidatedPasswordFieldItem
                   fieldName={fieldNames.password}
                   label={_('signup.label_password')}
                   placeholder={_('signup.placeholder_password')}
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -7,7 +7,7 @@ import {
   FieldGroup,
   InputFieldItem,
   CheckboxFieldItem,
-  PasswordFieldItem,
+  PasswordValidatedFieldItem,
   PhoneFieldItem,
 } from '../form-helpers/form-helpers';
 import { validateEmail, validateRequiredField } from '../../validations';
@@ -149,7 +149,7 @@ export default injectIntl(function({ intl }) {
                   type="email"
                   placeholder={_('signup.placeholder_email')}
                 />
-                <PasswordFieldItem
+                <PasswordValidatedFieldItem
                   fieldName={fieldNames.password}
                   label={_('signup.label_password')}
                   placeholder={_('signup.placeholder_password')}
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -10,7 +10,7 @@ import {
   ValidatedPasswordFieldItem,
   PhoneFieldItem,
 } from '../form-helpers/form-helpers';
-import { validateEmail, validateRequiredField } from '../../validations';
+import { validateEmail, validateRequiredField, validatePassword } from '../../validations';
 import LanguageSelector from '../shared/LanguageSelector/LanguageSelector';
 
 const fieldNames = {
@@ -74,17 +74,7 @@ export default injectIntl(function({ intl }) {
       errors[fieldNames.email] = _(emailMsgError);
     }
 
-    const digitRegex = /[0-9]/;
-    if (!values[fieldNames.password]) {
-      errors[fieldNames.password] = { empty: true };
-    } else if (values[fieldNames.password].length < 8) {
-      errors[fieldNames.password] = { charLength: true };
-      if (!digitRegex.test(values[fieldNames.password])) {
-        errors[fieldNames.password] = { charLength: true, digit: true };
-      }
-    } else if (!digitRegex.test(values[fieldNames.password])) {
-      errors[fieldNames.password] = { digit: true };
-    }
+    errors[fieldNames.password] = validatePassword(values[fieldNames.password]);
 
     if (!values[fieldNames.accept_privacy_policies]) {
       // TODO: show the right message
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -74,11 +74,16 @@ export default injectIntl(function({ intl }) {
       errors[fieldNames.email] = _(emailMsgError);
     }
 
+    const digitRegex = /[0-9]/;
     if (!values[fieldNames.password]) {
-      // TODO: I think that password validation has a different format
-      errors[fieldNames.password] = _('validation_messages.error_required_field');
-    } else {
-      // TODO: validate password
+      errors[fieldNames.password] = { empty: true };
+    } else if (values[fieldNames.password].length < 8) {
+      errors[fieldNames.password] = { charLength: true };
+      if (!digitRegex.test(values[fieldNames.password])) {
+        errors[fieldNames.password] = { charLength: true, digit: true };
+      }
+    } else if (!digitRegex.test(values[fieldNames.password])) {
+      errors[fieldNames.password] = { digit: true };
     }
 
     if (!values[fieldNames.accept_privacy_policies]) {
--- a/src/components/form-helpers/form-helpers.js
+++ b/src/components/form-helpers/form-helpers.js
@@ -52,44 +52,44 @@ export const FieldItem = connect(
   ),
 );
 
-export const PasswordWrapper = connect(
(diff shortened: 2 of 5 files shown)


src/components/form-helpers/form-helpers.js
```diff
@@ -52,6 +52,46 @@ export const FieldItem = connect(
...
+  ({ className, fieldName, children, formik: { errors, touched } }) => (
+    <li
+      className={concatClasses(className, touched[fieldName] && errors[fieldName] ? 'error' : '')}
+    >
+      {children}
+      <div className="wrapper-password">
+        <p className="password-message">
+          <span
+            className={
+              !touched[fieldName]
```
fcoronelmakingsense: it's not better create a function for class?

src/components/form-helpers/form-helpers.js
```diff
@@ -52,6 +52,46 @@ export const FieldItem = connect(
...
+                ? 'waiting-message'
```
andresmoschini: Is waiting status really necessary? I mean, I thought that the validation status was always visible.

andresmoschini:
@cbernat please fix the tests in order to have it deployed.

andresmoschini:
> Not touched
> image

What happens when you touch it and then clean the input?

andresmoschini:
> Completed
> image
> 
> Autocompleted
> image

Should these two scenarios not show green circles? or at least be the same?

src/components/Signup/Signup.js
```diff
@@ -74,11 +74,16 @@ export default injectIntl(function({ intl }) {
...
-    } else {
-      // TODO: validate password
+      errors[fieldNames.password] = { empty: true };
+    } else if (values[fieldNames.password].length < 8) {
+      errors[fieldNames.password] = { charLength: true };
+      if (!digitRegex.test(values[fieldNames.password])) {
+        errors[fieldNames.password] = { charLength: true, digit: true };
+      }
+    } else if (!digitRegex.test(values[fieldNames.password])) {
+      errors[fieldNames.password] = { digit: true };
```
andresmoschini: I understand the reason for having it in _validate_ function (or the same, in the _validation_ helper that @fcoronelmakingsense is creating), but I am still not comfortable in having this so far from _PasswordWrapper_ because both things are really coupled.

But, I am not sure about how to improve it, using _ref_ is an option, but I do not like it either. Maybe we should learn more about _Formik_ for further releases.

src/components/form-helpers/form-helpers.js
```diff
@@ -52,6 +52,46 @@ export const FieldItem = connect(
...
+                : touched[fieldName] &&
+                  errors[fieldName] &&
+                  (errors[fieldName].charLength || errors[fieldName].empty)
+                ? 'lack-message'
+                : 'complete-message'
+            }
+          >
+            <FormattedMessage id="validation_messages.error_password_character_length" />
```
andresmoschini: _FormattedMessage_ already creates a _span_ element, maybe you could apply the class to this element.

src/components/form-helpers/form-helpers.js
```diff
@@ -52,6 +52,46 @@ export const FieldItem = connect(
...
+              !touched[fieldName]
+                ? 'waiting-message'
+                : touched[fieldName] &&
+                  errors[fieldName] &&
+                  (errors[fieldName].digit || errors[fieldName].empty)
+                ? 'lack-message'
+                : 'complete-message'
+            }
+          >
+            <FormattedMessage id="validation_messages.error_password_digit" />
```
andresmoschini: The same comment about spam and FormattedMessage here.

src/components/form-helpers/form-helpers.js
```diff
@@ -139,14 +179,14 @@ export const InputFieldItem = ({ className, fieldName, label, type, placeholder
   </FieldItem>
 );
 
-export const PasswordFieldItem = ({ className, fieldName, label, placeholder, helpText }) => {
+export const PasswordFieldItem = ({ className, fieldName, label, placeholder }) => {
```
andresmoschini: 👍

src/components/form-helpers/form-helpers.js
```diff
@@ -139,14 +179,14 @@ export const InputFieldItem = ({ className, fieldName, label, type, placeholder
...
   const [passVisible, setPassVisible] = useState(false);
   const type = passVisible ? 'text' : 'password';
   const autocomplete = passVisible ? 'off' : 'current-password';
   const buttonClasses = passVisible ? 'show-hide icon-hide ms-icon' : 'show-hide ms-icon icon-view';
 
   return (
-    <FieldItem className={concatClasses('field-item', className)} fieldName={fieldName}>
+    <PasswordWrapper className={concatClasses('field-item', className)} fieldName={fieldName}>
```
andresmoschini: Oh, take care of this. This PasswordWrapper does not apply for all password fields, only for ones who define password: _signup_ and _change password_, but not for _login_.

src/components/form-helpers/form-helpers.js
andresmoschini: I verified:

Pristine:

image

Touched and cleaned:

image

--- 

In my opinion, for the circles, it should not care if the field has been touched or not, the red border is enough to show that there is an error. But it is just an opinion, you could talk it with Santiago or @IgnacioRodrigues

andresmoschini:
@IgnacioRodrigues and @GusBaamonde worked in the very first HTML, maybe they want to take a look. You can see it in action in https://cdn.fromdoppler.com/doppler-webapp/int-build304/#/signup

src/components/form-helpers/form-helpers.js
andresmoschini: image

It does not look bad, but, in my opinion, as I talked with Santiago, it is not a good idea having this kind of validation on password validation fields because: imagine that the rules changed, the old password follow the old rules and the message could be confusing.

src/components/Signup/Signup.js
cbernat: It would be great to have this validation in form helper, since it's small, and as you've said highly coupled with display in PasswordWrapper. I will try adding it there.

cbernat:
> > Completed
> > image
> > Autocompleted
> > image
> 
> Should these two scenarios not show green circles? or at least be the same?

I'm missing the green state my bad. It's not reusing this display with circles but with a new phrase in green.

src/components/form-helpers/form-helpers.js
cbernat: I didn't notice this, thanks!

src/components/form-helpers/form-helpers.js
cbernat: No can do. https://github.com/yahoo/react-intl/issues/704 I've tried but seems to be not allowed.

src/components/form-helpers/form-helpers.js
cbernat: Fixed this.

src/components/form-helpers/form-helpers.js
cbernat: By default is waiting state, in pristine state. I don't think they should be the same, the issue here is avoid showing errors prematurely to the user, imagine if you go to a form and it's all red and you didn't even typed a thing. O_O

src/components/form-helpers/form-helpers.js
andresmoschini: Oh, sorry, you are RIGHT!

I was really convinced that it just works, see this code:

https://github.com/FromDoppler/doppler-webapp/blob/634eda7080d234d8a9041feba52efe4ba7822eed/src/components/Loading/Loading.js#L9

Well, I think that without flash it is better haha.

andresmoschini: The other option is using _FormattedMessage_ as a wrapper, or simply having two spans as we have now.

src/components/Signup/Signup.js
andresmoschini: We have talked about it. Resolving the conversation...

src/components/form-helpers/form-helpers.js
andresmoschini: No, it does not work.

src/components/form-helpers/form-helpers.js
```diff
@@ -173,6 +213,40 @@ export const PasswordFieldItem = ({ className, fieldName, label, placeholder, he
   );
 };
 
+export const PasswordValidatedFieldItem = ({ className, fieldName, label, placeholder }) => {
```
andresmoschini: What do you think about `ValidatedPasswordFieldItem` or `PasswordDefinitionFieldItem`?

andresmoschini: If I am not wrong, a big part of _PasswordValidatedFieldItem_ content and _PasswordFieldItem_ is the same, you could move it into a private component and reuse it in both components.

cbernat: This one `ValidatedPasswordFieldItem` it's better. I like the word validated because the difference between this field and the one in the login is the validation. `Definition` I think would not fit here. My order of words is a bit weird so I will change it. Thanks!

andresmoschini:
:shipit:

PR #118, opened 2019-04-13
Unify submit buttons behavior

andresmoschini:
Hi team, 

I forgot to apply the defined submit button behavior in Signup page and then the same happened with the rest of public pages.

These changes apply the same style with a spinner in Signup, Login and Forgot Password pages, but more important: disable the button in order to avoid multiple submits.

Could you review?


cbernat:
👍

PR #119, opened 2019-04-13
fix: use type text in place of type email for email fields

andresmoschini:
Hi team, 

Based on what @fcoronelmakingsense and @mvillaseco have researched, at least by the moment, we are not using `type="email"` on our inputs. See #108.

Here, I am applying in Login and Signup pages the same change that @fcoronelmakingsense already applied in Forgot Password page.


andresmoschini:
It is also included in #121

PR #121, opened 2019-04-14
Refactor validations

andresmoschini:
Hi team, I think that it fixes the weird problem that we had related to reusing parameters and coupled validations and fields.

In the beginning, we did not pay attention to [Formik's field-level validations](https://jaredpalmer.com/formik/docs/guides/validation#field-level-validation), with them I think that we could improve a lot our _form helpers_ API. 

Now: 
* the form's "validate" function is not necessary in normal cases (as you can see, I have removed it from Login, Signup and Forgot Password pages)
* the validation is configurable at _FieldItem_ level (for example `required` and `checkRequired` props)
* some _FieldItems_ has predefined validations (for example _EmailFieldItem_ and _PhoneFieldItem_)

To get here, I had to refactor some code, so this PR is a little big, but I think that it is easy to review commit by commit.

There are also some opinable changes, like:

* using _locale literal keys_ in place of already translated strings in validations
* using _true_ in place of a _locale literal key_ when there is a validation error without a message.
* allowing to override _required_ validation message (because we are not taking advantage of it, but I think that we will need it soon)

Could you review it?


src/components/form-helpers/form-helpers.js
```diff
@@ -86,6 +107,26 @@ const _PhoneFieldItem = ({
     }
   };
 
+  const validatePhone = (value) => {
```
cbernat: Maybe you could move this logic into `validations.js`? as the rest of the validations.

andresmoschini: Yes, but in this case, it is coupled to `iti` object, I could pass it to `validations.js`, but I am not sure, this validation is different to the rest validations. what do you think?

cbernat: What about sending this `iti` as a parameter?. I think it would be more clear to have all validations in one place.

andresmoschini: Yes, it is possible, let me continue thinking about it. Thanks!

PR #123, opened 2019-04-15
Add dynamic meta description in page Signup and Login

IgnacioRodrigues:
In this commit, we add that a package called "React-Helmet" is a reusable React component that manages changes in the header of the document and with this, the content of the meta description is added in the login and registration page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "query-string": "^6.4.2",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
+	"react-helmet": "^5.2.0",
     "react-intl": "^2.8.0",
     "react-router-dom": "^5.0.0",
     "react-scripts": "^2.1.8",
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -3,6 +3,7 @@ import { Link, Redirect } from 'react-router-dom';
 import { FormattedHTMLMessage, injectIntl } from 'react-intl';
 import { timeout } from '../../utils';
 import { Formik, Form } from 'formik';
+import { Helmet } from 'react-helmet';
 import {
   EmailFieldItem,
   FieldGroup,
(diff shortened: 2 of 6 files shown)


andresmoschini:
Por favor @IgnacioRodrigues tratá de mantener las ramas de features limpias y sin merges internos

image

Si podés acomodá eso rabaseando en lugar de mergear master en tu rama.

IgnacioRodrigues:
> acomodá

IgnacioRodrigues:
Sorry, I did not want to close this pull request.

.editorconfig
```diff
@@ -10,8 +10,4 @@ max_line_length = 100
 indent_size = 2
 
 [*.md]
-trim_trailing_whitespace = false
-
-[*.{js,ts}]
-block_comment_start = /*
-block_comment_end = */
```
andresmoschini: Why did you remove these rules?

andresmoschini:
Oh, there is a lot of unrelated things here.

Please, fix the PR and let us know.

Today I cannot help you on this, but maybe @mvillaseco or @VicThorX could.

.editorconfig
IgnacioRodrigues: This change is the commit for revert merge

andresmoschini:
Helmet looks very interesting. 

As part of this task (or another one), we should apply this same to other places where we are updating the title, and we should also apply it to change the language, see: https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L2

Would it be fine?

IgnacioRodrigues:
> Helmet looks very interesting.
> 
> As part of this task (or another one), we should apply this same to other places where we are updating the title, and we should also apply it to change the language, see: https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L2
> 
> Would it be fine?

Okey, added the new change in this pull request.

andresmoschini:
> Okey, added the new change in this pull request.

I understand that you have other responsibilities than this webapp, so, maybe it is a good idea to merge current work and reserve my proposals for another PR.

IgnacioRodrigues:
Ok, I'm going to merge this extraction request and the other changes will create a new pull request. Thanks @andresmoschini

andresmoschini:
:shipit:

PR #125, opened 2019-04-15
Fix change domain

andresmoschini:
Hi team, 

There was a problem using real Datahub domains data because it does not contain id field.

I fixed it, and also removed all page related logic because it is not being used.


cbernat:
👍

PR #127, opened 2019-04-16
Accept register user results

andresmoschini:
Hi team, 

I applied these changes to deal with register user errors, by the moment we accept two kinds of errors:

* Email already exists
* Unexpected errors

Could you review it?

Here you have some examples of how it works:

Duplicated email address (cached):

Unexpected error:

Fixes: DBR-174, DBR-186

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -11,7 +11,7 @@
     -->
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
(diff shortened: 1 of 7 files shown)


andresmoschini:
Here you have some examples of how it works:

Duplicated email address (cached):



Unexpected error:

andresmoschini:
If we want to reproduce these scenarios inlocal environment, you can edit `doppler-legacy-client.doubles`:

* Success 

image

* Already existent email

image

* Unexpected error

image

PR #130, opened 2019-04-18
feat: deal with unexpected errors in Forgot Password

andresmoschini:
It simply shows an unexpected error message in Forgot Password when the result (of nothing by the moment) is not successful.


cbernat:
Maybe is out of scope, but the placeholder for email is a bit weird don't you think? Perhaps as in signup, something like "Escribe tu email de usuario aqui"?

andresmoschini:
> Maybe is out of scope, but the placeholder for email is a bit weird don't you think? Perhaps as in signup, something like "Escribe tu email de usuario aqui"?

It is not the placeholder, it is my email 😂

PR #131, opened 2019-04-18
feat: deal with unexpected errors in Login

andresmoschini:
It simply shows an unexpected error message in Login when the result (of nothing by the moment) is not successful.

(Like #130 but for Login)


cbernat:
👍

PR #133, opened 2019-04-19
Deal with origin

andresmoschini:
Hi team, 

These changes are the complement to https://github.com/MakingSense/Doppler/pull/5886

We are capturing the `origin` parameter in any webapp route.

If the user navigates inside the app, this value is not lost and if the user signs-up, the value is sent to Doppler Legacy as `origin` parameter.

Besides that, the first time that the `origin` parameter is read, we stored it in the local storage, then if the user signs-up, the value is sent to Doppler Legacy as `firstOrigin` parameter.

I am also storing the date of this first origin in the local storage in case of we want to implement any case of expiration in the future.


src/App.test.js
```diff
@@ -444,4 +463,103 @@ describe('App component', () => {
...
+        </AppServicesProvider>,
+      );
+
+      // Assert
+      const localStorageItems = dependencies.localStorage.getAllItems();
+      expect(localStorageItems['dopplerOrigin.value']).toBeDefined();
+      expect(localStorageItems['dopplerOrigin.value']).toEqual(oldValue);
+    });
+
+    it('should not be cleanes in local storage when there is not origin URL parameter', () => {
```
cbernat: Maybe `cleanes` --> `cleaned`?

cbernat:
👍

PR #138, opened 2019-04-21
DBR-173 - Captcha

andresmoschini:
Hi Team,

This is the second try of the captcha field in signup form.

I have added the most popular related library and the real keys for each environment.

Finally, I decided to use Invisible captcha and I prepared the most elegant solution that I could think (that is not so much) to integrate it with Formik.

Could you review?

(I am working in some images to illustrate how it works)

### Legal notes

image

### Captcha pass transparentely

captcha-pass

### Captcha pass after challenge

captcha_pass_with_challenge

### User closes the challenge (**Dead end**)

captcha_user_closes_challenge

If the challenge window is closed, we do not have feedback, so, by the moment, we will keep the submit button disabled.

See more details in https://stackoverflow.com/questions/43488605/detect-when-challenge-window-is-closed-for-google-recaptcha


andresmoschini:
Hi Team,

This is the second try of the captcha field in signup form.

I have added the most popular related library and the real keys for each environment.

Finally, I decided to use Invisible captcha and I prepared the most elegant solution that I could think (that is not so much) to integrate it with Formik.

Could you review?

(I am working in some images to illustrate how it works)

mvillaseco:
The comments makes it more easy to understand, thanks!

👍

cbernat:
Did you take in account that besides the captcha we count user attempts and block the user right? https://github.com/MakingSense/Doppler/blob/develop/Doppler.Presentation.MVC/Controllers/SignInController.cs#L334

cbernat:
Right now I don't see anything weird. 👍

PR #143, opened 2019-04-24
Add Google Analytic

IgnacioRodrigues:
Add [react-ga](https://www.npmjs.com/package/react-ga) module that can be used to include Google Analytics tracking code in a website or app that uses React for its front-end codebase.


src/index.js
```diff
@@ -22,6 +24,17 @@ const forcedServices =
...
     : {};
 
+// Initialize Google Analytics ID
+ReactGA.initialize('UA-532159-1');
+
+const history = createBrowserHistory();
+
+// Get the current location
+history.listen((location, action) => {
+	ReactGA.set({ page: location.pathname });
```
andresmoschini: Did you test it, for example with `console.log()` to verify that it is working?

yarn.lock
```diff
@@ -8836,6 +8836,11 @@ react-fast-compare@^2.0.1:
   resolved "https://registry.yarnpkg.com/react-fast-compare/-/react-fast-compare-2.0.4.tgz#e84b4d455b0fec113e0402c329352715196f81f9"
   integrity sha512-suNP+J1VU1MWFKcyt7RtjiSWUjvidmQSlqu+eHslq+342xCbGTYmC0mEhPCOHxlW0CywylOC1u2DFAT+bv4dBw==
 
+react-ga@^2.5.7:
+  version "2.5.7"
+  resolved "https://registry.yarnpkg.com/react-ga/-/react-ga-2.5.7.tgz#1c80a289004bf84f84c26d46f3a6a6513081bf2e"
+  integrity sha512-UmATFaZpEQDO96KFjB5FRLcT6hFcwaxOmAJZnjrSiFN/msTqylq9G+z5Z8TYzN/dbamDTiWf92m6MnXXJkAivQ==
+
```
andresmoschini: _history_ is not included in the changes, are you sure that you are including all changes of _yarn.lock_?

andresmoschini:
Please confirm that you have tried navigation changes (or call me if you do not understand me), run prettier, and then, feel free to merge it.

Good work!

src/index.js
IgnacioRodrigues: Yes, add `console.log()´ for test this function and this working.

yarn.lock
IgnacioRodrigues: mm that's weird, I'm including every change of thread. What change do you think is missing?

andresmoschini: Nah, maybe we already had that dependency (as a dependency of another dependency) and for that reason, it does not impact in the lock file.

IgnacioRodrigues: Yes, the dependency history already had.

andresmoschini:
Please, do not forget to run prettier (you can also install it as a [plugin in your favorite editor](https://prettier.io/docs/en/editors.html))

PR #150, opened 2019-04-30
chore: change report time filter values

cbernat:
Changed time filter options in reports. 

image

Issue: DBR-184

### DIFF
--- a/src/i18n/es.json
+++ b/src/i18n/es.json
@@ -67,10 +67,9 @@
     "domain_not_verified": "Dominio no verificado",
     "pages": "Páginas",
     "rank_time": "Rango de tiempo",
-    "rank_time_item1": "última semana",
-    "rank_time_item2": "último mes",
-    "rank_time_item3": "últimos 3 meses",
-    "rank_time_item4": "últimos 6 meses",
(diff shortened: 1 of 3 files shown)


src/i18n/es.json
```diff
@@ -65,10 +65,9 @@
     "domain_not_verified": "Dominio no verificado",
     "pages": "Páginas",
     "rank_time": "Rango de tiempo",
-    "rank_time_item1": "última semana",
-    "rank_time_item2": "último mes",
-    "rank_time_item3": "últimos 3 meses",
-    "rank_time_item4": "últimos 6 meses",
+    "rank_time_item1": "1 semana",
+    "rank_time_item2": "2 semanas",
+    "rank_time_item3": "3 semanas",
```
andresmoschini: What do you think about renaming it as something related to actual values?

cbernat: If you think it's best... I don't know how sure are we of these values...

andresmoschini: It is not really necessary, but what do you think about using pluralization features?

andresmoschini:
👍

PR #156, opened 2019-05-01
chore: update library version to support buttons within error messages

andresmoschini:


andresmoschini:
Very minor, merging...

PR #158, opened 2019-05-01
DBR-213 - Login error "Email not validated"

andresmoschini:
Hi team, and specially @carodipietro and @santiagopaolucci,

This is the error when the user did not finish the registration process because he did not validate the email.

We are using captcha in all actions (login and resend email), the message disappears when the user does something in the login form fields and when the user clicks in the action, the email is resent.

Again, in this case, you can propose quick improvements in the content, or better enhancements could be done in future PR.

### DIFF
--- a/src/components/Login/LoginErrorAccountNotValidated.js
+++ b/src/components/Login/LoginErrorAccountNotValidated.js
@@ -1,0 +1,47 @@
+import React, { useState } from 'react';
+import { useCaptcha } from '../form-helpers/captcha-utils';
+import { injectIntl } from 'react-intl';
+import { InjectAppServices } from '../../services/pure-di';
+
+export const LoginErrorAccountNotValidated = injectIntl(
+  InjectAppServices(
+    /**
+     * @param { Object } props
+     * @param { string } props.email
+     * @param { import('react-intl').InjectedIntl } props.intl
+     * @param { import('../../services/pure-di').AppServices } props.dependencies
+     */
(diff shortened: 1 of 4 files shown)


src/components/Login/LoginErrorAccountNotValidated.js
```diff
@@ -0,0 +1,47 @@
...
+            {intl.formatMessage({ id: 'signup.email_not_received' })}{' '}
+            <button type="button" onClick={incrementAndResend}>
+              {intl.formatMessage({ id: 'signup.resend_email' })}
+            </button>
+            .
+          </p>
+        </>
+      ) : (
+        // TODO: review content
+        <p>{intl.formatMessage({ id: 'signup.no_more_resend' })}</p>
```
andresmoschini: @carodipietro, in this case, we have no new content because it is the same content that we shown in the confirmation page after signup, you can see it in:

* https://github.com/FromDoppler/doppler-webapp/blob/master/src/i18n/es.json#L98
* https://github.com/FromDoppler/doppler-webapp/blob/master/src/i18n/es.json#L88
* https://github.com/FromDoppler/doppler-webapp/blob/master/src/i18n/es.json#L106
* https://github.com/FromDoppler/doppler-webapp/blob/master/src/i18n/en.json#L98
* https://github.com/FromDoppler/doppler-webapp/blob/master/src/i18n/en.json#L88
* https://github.com/FromDoppler/doppler-webapp/blob/master/src/i18n/en.json#L106

PR #160, opened 2019-05-02
fix: not allow dot before @

lucianaDilizio:
- Do not allow dot before @


andresmoschini:
Great, but please talk with @mvillaseco to review what validation are we applying in other places

lucianaDilizio:
The fix was applied in Automation too.

andresmoschini:
> The fix was applied in Automation too.

 👏

PR #163, opened 2019-05-02
chore: add zoho chat in spanish only

cbernat:
Add zoho chat, as it is in doppler and sites. 
Changed font handling, and renamed some variables. 
image

https://cdn.fromdoppler.com/doppler-webapp/int-build521/#/signup


andresmoschini:
👍  Nice!

Please add a task to allow changing the language. And consider a future task for moving it to Doppler UI Library.

PR #164, opened 2019-05-03
fix: fix email validation

andresmoschini:
Hi team, 

I found that the last changes in email validation do not allow emails with only a letter before @, I fixed that and added some tests.

Could you review?

CC: @CarlosSampedro, maybe you could think in other scenarios.
CC: @lucianaDilizio, if I am not wrong, you have applied the previous change in another place, you should consider also apply this one.

### DIFF
--- a/src/validations.test.js
+++ b/src/validations.test.js
@@ -63,6 +63,17 @@ describe('validations', () => {
       expect(result).toEqual('validation_messages.error_invalid_email_address');
     });
 
+    it('should accept an email address with multiple levels of subdomains', () => {
+      // Arrange
+      const emailAddress = 'test@test.test.com.ar';
+
+      // Act
+      const result = validateEmail(emailAddress);
+
+      // Assert
+      expect(result).toBeNull();
+    });
+
     it('should not accept an email address starting with dot (.)', () => {
       // Arrange
       const emailAddress = '.test@test.com';
--- a/src/validations.test.js
+++ b/src/validations.test.js
@@ -41,9 +41,42 @@ describe('validations', () => {
       expect(result).toBeNull();
     });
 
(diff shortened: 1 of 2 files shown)


src/validations.test.js
```diff
@@ -52,6 +52,61 @@ describe('validations', () => {
       expect(result).toBeNull();
     });
 
+    it('should not accept an email address with dot (.) before @', () => {
```
cbernat: What do you think about adding a test with an email like `email@mail.com.ar`?

lucianaDilizio:
This is the PR with the same changes in Automation: https://github.com/MakingSense/Doppler/pull/5944

CarlosSampedro:
Hi @andresmoschini ,

I have a question regarding the email validation:
Should this email validation be applied just un logn in/ Sign up or should it be applied with subscribers too?

On the other hand, I have verified the regular expression you made with the email validation (sign up) in Production, and I have  detected differencies with some chars. The following emails are refused by the regular expression but they are accepted in production:
T&P@gmail.com.ar
T/P@gmail.com.ar
T~P@gmail.com.ar

Later I will take a look the Unit Test escenarios looking for a new ones.

Regards

andresmoschini:
Buen trabajo @CarlosSampedro, según estuvimos chateando, deberíamos haber utilizado la misma expresión regular que se está usando en Doppler.

Por ahora, vamos a investigar si hay usuarios con caracteres no soportados, y en ese caso corregiremos la expresión regular actual. 

Luego, con más tiempo intentaremos unificar ambas validaciones.

andresmoschini:
@CarlosSampedro @lucianaDilizio I have updated it to support some weird real scenarios and simplified the regular expression.

BTW, thanks @fgchaio for the research!

PR #168, opened 2019-05-03
feat: deal with account canceled error

lucianaDilizio:
Catch error in login when the account is canceled:

image

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -15,6 +15,7 @@ import LanguageSelector from '../shared/LanguageSelector/LanguageSelector';
 import RedirectToLegacyUrl from '../RedirectToLegacyUrl';
 import { InjectAppServices } from '../../services/pure-di';
 import { LoginErrorAccountNotValidated } from './LoginErrorAccountNotValidated';
+import LoginErrorCancelatedAccount from './LoginErrorCancelatedAccount';
 
 const fieldNames = {
   user: 'user',
@@ -71,6 +72,8 @@ const Login = ({ intl, location, dependencies: { dopplerLegacyClient, sessionMan
         setErrors({ _general: 'validation_messages.error_unexpected' });
       } else if (result.expectedError && result.expectedError.accountNotValidated) {
         setErrors({ _general: <LoginErrorAccountNotValidated email={values[fieldNames.user]} /> });
+      } else if (result.expectedError && result.expectedError.cancelatedAccount) {
+        setErrors({ _general: <LoginErrorCancelatedAccount /> });
       } else {
         console.log('Unexpected error', result);
         setErrors({ _general: 'validation_messages.error_unexpected' });
--- a/src/i18n/es.json
+++ b/src/i18n/es.json
@@ -115,6 +115,7 @@
     "title": "Solicita una actualización de tu plan"
   },
   "validation_messages": {
+    "error_account_is_canceled_HTML" : "Tu cuenta se encuentra cancelada. Para más información <a target=\"_blank\" href=\"http://www.fromdoppler.com/en/contact.html\">contáctanos</a>.",
     "error_checkbox_policy": "¡Ouch! No has aceptado la Política de Privacidad de Doppler.",
     "error_email_already_exists": "¡Ouch! Ya posees una cuenta en Doppler.",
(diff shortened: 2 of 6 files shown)


src/components/Login/Login.js
```diff
@@ -71,6 +71,9 @@ const Login = ({ intl, location, dependencies: { dopplerLegacyClient, sessionMan
         setErrors({ _general: 'validation_messages.error_unexpected' });
       } else if (result.expectedError && result.expectedError.accountNotValidated) {
         setErrors({ _general: <LoginErrorAccountNotValidated email={values[fieldNames.user]} /> });
+      } else if (result.expectedError && result.expectedError.accountIsCanceled) {
+        // TODO: define content when the account is canceled
```
andresmoschini: Could you validate it with @carodipietro before merging?

src/i18n/es.json
```diff
@@ -115,6 +115,7 @@
     "title": "Solicita una actualización de tu plan"
   },
   "validation_messages": {
+    "error_account_is_canceled" : "Esta cuenta se encuentra cancelada. Por favor, contacta a soporte.",
```
andresmoschini: @carodipietro could you review this content?

src/components/Login/Login.js
andresmoschini: Oh, this content already exists in Doppler, could you use Doppler content by default? (maybe you are already doing that, in that case, please remove the TODO comment)

src/services/doppler-legacy-client.ts
```diff
@@ -26,7 +26,8 @@ interface PayloadWithCaptchaToken {
 export type LoginErrorResult =
   | { blockedAccountNotPayed?: false; accountNotValidated?: false }
   | { blockedAccountNotPayed: true; accountNotValidated?: false }
-  | { accountNotValidated: true; blockedAccountNotPayed?: false };
+  | { accountNotValidated: true; blockedAccountNotPayed?: false }
+  | { accountIsCanceled: true; blockedAccountNotPayed?: false };
```
andresmoschini: I know that it is not the best, but could you add the new error flag in all error result types?

Something like:

```
{ blockedAccountNotPayed?: false; accountNotValidated?: false; accountIsCanceled?: false }
```

It will help in intellisense.

src/services/doppler-legacy-client.ts
```diff
@@ -251,6 +252,10 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
         return { expectedError: { accountNotValidated: true } };
       }
 
+      if (!response.data.success && response.data.error == 'AccountIsCanceled') {
```
andresmoschini: This code seems to be wrong, did you tested it? please check [the ticket|http://jira.makingsense.com/browse/DBR-192] and fix it if the information is wrong.

image

andresmoschini:
Great work @lucianaDilizio!

Please ensure to use the same content than Doppler does, also, please check it against INT or your local environment to be sure that it is working as expected.

src/services/doppler-legacy-client.ts
lucianaDilizio: I took this from `ApplicationErrorCode.cs`

image

andresmoschini: Ok, probalo y en todo caso actualizá el ticket por favor.

lucianaDilizio: You were right, I have to use CancelatedAccount 😅

src/i18n/es.json
lucianaDilizio: By the moment I will apply the same that we are using in now.

src/components/Login/Login.js
lucianaDilizio: No, I invented this comment but for now I will use it the same that in Doppler. I'd leave this TODO comment to remember change it when Carolina has the real content.

carodipietro: Este contenido no está generado para esta versión. Podés usar lo que está en Doppler actualmente hasta que se generen los nuevos contenidos.

src/i18n/es.json
carodipietro: Este contenido no está generado para esta versión. Podés usar lo que está en Doppler actualmente hasta que se generen los nuevos contenidos.

PR #173, opened 2019-05-06
Better than json

andresmoschini:
Hi Team, 

This is a polemic PR, but I think that it will be very useful for us to review and understand our content files.

It is also useful to avoid pass redundant parameters like year or Doppler base URL.

I think that reviewing first the [README.md file](https://github.com/FromDoppler/doppler-webapp/pull/173/files?short_path=09f437c#diff-09f437c778eec10fbc446e4550b5d308) is useful. There I try to explain the rationale and guidelines of this approach.


src/i18n/en.js
```diff
@@ -0,0 +1,232 @@
...
+    forgot_password: `Do not you remember your password?`,
+    head_description: `Attract, engage and convert clients using the Email Marketing Automation power. Try out Doppler!`,
+    head_title: `Free Email Marketing Automation with no sending limits | Doppler`,
+    label_user: `Username: `,
+    signup: `Sign up free`,
+    you_want_create_account: `Do not you have an account yet?`,
+  },
+  reports: {
+    allow_enable_trial_HTML: `
+<p>
```
lucianaDilizio: The tabulation here sounds me a little weird. I imagine this tabulated with respect to `allow_enable_trial_HTML`.

andresmoschini: You are right, as you propose is nicer, and it could work on HTML, but not in Markdown.

Because, since we are using string templates, the spaces are part of the value.

When the value has more than a line, the current approach is: 

1. add a line break before the value
2. indent based on column 0
3. finish with a new line break

image

If you like, we could use your proposed indentation for HTML, but not in Markdown, it will be inconsistent. Thoughts?

lucianaDilizio: I'm agree to use the identacion for HTML 😄. At least till we replace them with MD.

src/i18n/en.js
```diff
@@ -0,0 +1,232 @@
...
+// Markdown Links
+const linkPrivacyAndLegals_MD = `[Privacy and Legal Policy](${urlPrivacy})`;
+
+// HTML Links
+const linkSiteTracking_HTML = `<a href="${urlSiteTracking}">On-Site Tracking</a>`;
+const linkAddYourDomain_HTML = `<a href="${urlSiteTracking}">Add your domain</a>`;
+// TODO: confirm if utm parameter goes here
+const linkPrivacy_HTML = `<a target="_blank" href="${urlPrivacy}?utm_source=app&utm_medium=landing&utm_campaign=signup">Privacy Policy</a>`;
+const linkBuyMonthly_HTML = `<a href="${urlBuyMonthly}">UPGRADE NOW</a>`;
+const linkContactUs_HTML = `<a target="_blank" href="${urlContact}">contact us</a>`;
```
lucianaDilizio: I'm not sure if this is a good idea. Maybe the contents could be:

- To know more please **contact as**.

or

- For security reasons we've temporarily disabled your account. **Contact Us**.

We should have a variable for each different way of the same text. It could be result el little tedious.

andresmoschini: Ok, so, to clarify, I should remove `linkContactUs_HTML`, right?

lucianaDilizio: Yes, I don't know if we are going to have the same problem with the other texts. I was thinking that maybe it's not a good idea to have the HTML links in variables. But it is just only my opinion. I would like to hear other voices.

andresmoschini: Based on current examples, you are right, so, by the moment I will omit link constants, but we could consider that if from the content side they want to unify communication.

cbernat:
👍

andresmoschini:
Hey, @lucianaDilizio! I have followed your recommendations, would you like to take a look while I work on rebasing it?

lucianaDilizio:
Thanks @andresmoschini!! I like it better now.

PR #176, opened 2019-05-06
chore: add BE redirect url handling

cbernat:
For some integrations in Doppler (Tienda Nube and Mercadoshops ), we have to allow login and redirect to continue. 

https://cdn.fromdoppler.com/doppler-webapp/int-build596/#/login

### DIFF
--- a/src/services/doppler-legacy-client.doubles.ts
+++ b/src/services/doppler-legacy-client.doubles.ts
@@ -19,7 +19,8 @@ export class HardcodedDopplerLegacyClient implements DopplerLegacyClient {
   public async login(model: LoginModel): Promise<LoginResult> {
     console.log(this.login, model);
     await timeout(1500);
-    return { success: true };
+    return { success: true, value: {} };
+    // return { success: true, value: { redirectUrl: 'Integration/Integration/TiendaNubeIntegration' } };
     // return { expectedError: { blockedAccountNotPayed: true } };
     // return { expectedError: { accountNotValidated: true } };
     // return { expectedError: { cancelatedAccount: true } };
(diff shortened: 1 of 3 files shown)


src/services/doppler-legacy-client.doubles.ts
```diff
@@ -20,6 +20,7 @@ export class HardcodedDopplerLegacyClient implements DopplerLegacyClient {
     console.log(this.login, model);
     await timeout(1500);
     return { success: true };
+    //return { success: true, redirectUrl: "Integration/Integration/TiendaNubeIntegration"};
```
andresmoschini:
```suggestion
    // return { success: true, redirectUrl: 'Integration/Integration/TiendaNubeIntegration' };
```

src/services/doppler-legacy-client.ts
```diff
@@ -16,7 +16,10 @@ export interface DopplerLegacyClient {
 type UnexpectedError = { success?: false; message?: string | null; error?: any };
 type ErrorResult<TError> = { success?: false; expectedError: TError } | UnexpectedError;
 type Result<TResult, TError> = { success: true; value: TResult } | ErrorResult<TError>;
-type EmptyResult<TError> = { success: true } | ErrorResult<TError>;
+type EmptyResult<TError> =
+  | { success: true }
+  | { success: true; redirectUrl?: string }
```
andresmoschini: _EmptyResult_ tries to refer to a generic result without output payload. I think that it is not a good idea change it in this way.

andresmoschini:
Great work! but please, review the changes in the type, I think that it is not the good path.

cbernat:
> Great work! but please, review the changes in the type, I think that it is not the good path.

I've added a new type `RedirectResult` what do you think?

PR #184, opened 2019-05-07
feat: add content forgot password

GusBaamonde:
content_forgot

### DIFF
--- a/src/i18n/en.js
+++ b/src/i18n/en.js
@@ -43,7 +43,7 @@ export default {
     iso: `ISO Quality Certification 9001:2008`,
   },
   forgot_password: {
-    back_login: `Did you remember your password and want to return?`,
+    back_login: `Did you remember your password? Go back to Log in.`,
     back_login_after_forgot: `Return to login`,
     confirmation_message_HTML: `
       <p>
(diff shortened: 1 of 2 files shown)


src/i18n/en.js
```diff
@@ -43,7 +43,7 @@ export default {
     iso: `ISO Quality Certification 9001:2008`,
   },
   forgot_password: {
-    back_login: `Did you remember your password and want to return?`,
+    back_login: `Remembered your password? Go back to Log in.`,
```
andresmoschini: Está bien este texto? en general las preguntas comienzan con un auxiliar, no? @carodipietro ?

src/i18n/en.js
```diff
@@ -53,19 +53,19 @@ export default {
         You’ll find an Email with steps to follow.
       </p>
       `,
-    description: `Do not worry! It happens to everyone.`,
-    description2: `Enter your email and we will help you`,
+    description: `Don’t worry! It happens.`,
+    description2: `Enter your Email and we’ll be glad to help you.`,
```
andresmoschini: @carodipietro could you confirm what is the desired apostrophe? We need to unify it in all the texts.

src/i18n/en.js
```diff
@@ -53,19 +53,19 @@ export default {
...
   },
   loading: `Loading...`,
   login: {
-    button_login: `Enter`,
+    button_login: `Request`,
     enter_doppler: `Enter Doppler`,
     enter_doppler_sub: `Enjoy the benefits of Email Marketing.`,
     error_payment: `Account blocked, please contact support.`,
-    forgot_password: `Do not you remember your password?`,
+    forgot_password: `Forgot your password?`,
```
andresmoschini: Is this question fine @carodipietro ?

PR #190, opened 2019-05-09
feat: deal with register denied error

lucianaDilizio:
Catch denied register error:
To test it: https://cdn.fromdoppler.com/doppler-webapp/int-build648/#/signup

image

### DIFF
--- a/src/services/doppler-legacy-client.doubles.ts
+++ b/src/services/doppler-legacy-client.doubles.ts
@@ -32,6 +32,7 @@ export class HardcodedDopplerLegacyClient implements DopplerLegacyClient {
     console.log(this.registerUser, model);
     await timeout(1500);
     return { success: true };
+    //return { expectedError: { registerDenied: true } };
   }
 
   public async resendRegistrationEmail(model: ResendRegistrationModel) {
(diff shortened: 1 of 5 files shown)


andresmoschini:
Nice!

src/services/doppler-legacy-client.doubles.ts
```diff
@@ -32,6 +32,7 @@ export class HardcodedDopplerLegacyClient implements DopplerLegacyClient {
     console.log(this.registerUser, model);
     await timeout(1500);
     return { success: true };
+    //return { expectedError: { registerDenied: true } };
```
cbernat: @lucianaDilizio comment?

lucianaDilizio: This is to test different responses in localhost.

andresmoschini: These comments are here to test the scenario easily in the local environment.

cbernat: Well I see both of you have the same idea about this. It should be nice in the future perhaps adding a comment stating the use of this line.

andresmoschini: Nah, if we all already know, the comment is not necessary :P

PR #193, opened 2019-05-10
DBR-257 - Remove "loading" word from page loading element

andresmoschini:
Hi Team,

@CarlosSampedro detected that the loading component's message is in the wrong language for some time.

I knew about that issue, but it seems not important to me, I think that the user has no issue with that word, but it seems that it is not comfortable for some users.

So, since it is not possible knowing the user language early in an elegant and always effective way, I have chosen directly remove the text.

What do you think?

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -31,7 +31,7 @@
     <div id="root">
       <div class="wrapper-loading">
         <div class="loading-page">
-          <p class="flash">Loading...</p>
+          <p class="flash">&nbsp;</p>
         </div>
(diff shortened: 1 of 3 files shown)


GusBaamonde:
We will have to work so that the spinner is rendered in the desired components and perhaps unify the color

andresmoschini:
I have to fix test that are depending on loading text to determine the language :(

cbernat:
I like it better without words :)

PR #198, opened 2019-05-14
feat: add autoFocus property to first field on each form

andresmoschini:
In some places, for example https://davidwalsh.name/react-autofocus, they say that _autoFocus_ is not enough.

However, in some places, for example https://stackoverflow.com/questions/42324423/autofocus-an-input-element-in-react-js#42324485
other said that _autoFocus_ should work.

By the moment, it is working, we can improve it in the future, for example scanning autoFocus fields in FormWithCaptcha or other places.

Could you review?

You could test the build in:

* https://cdn.fromdoppler.com/doppler-webapp/int-build675/#/login
* https://cdn.fromdoppler.com/doppler-webapp/qa-build675/#/login
* https://cdn.fromdoppler.com/doppler-webapp/build675/#/login

### DIFF
--- a/src/components/ForgotPassword/ForgotPassword.js
+++ b/src/components/ForgotPassword/ForgotPassword.js
@@ -92,6 +92,7 @@ const ForgotPassword = ({ intl, dependencies: { dopplerLegacyClient } }) => {
             <fieldset>
               <FieldGroup>
                 <EmailFieldItem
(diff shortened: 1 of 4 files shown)


damianmuti:
Seems to be working properly. Great usability improvement, @andresmoschini ! 👏

PR #200, opened 2019-05-14
feat: do not stop in toggle password visibility icon

andresmoschini:
Hi team, 

By the moment, I simply disabled tab stop for password visibility toggle button. 

In the future we could move the tab stop after the input, as Google does:

Thoughts?


damianmuti:
A good idea would be to switch the DOM order of elements.

image

andresmoschini:
Nice @damianmuti, as we talked, it is a little risky that I made that change, and we have higher priorities.

So, maybe you could apply that change in the future, maybe we could even move the toggle button inside the input.

PR #202, opened 2019-05-14
fix: issues with line-height and font declaration specificity

damianmuti:
Bumped library version based on a fix there. 
Currently `<p>` tags have a default line-height value of 22px. Based on my fix, the idea would be to have a dynamic unitless line height value that is proportional to the default line height for 15px of font-size.

NOTE: I've tested this change, and despite of visually changing line heights, I guess it's much better to have a proportional ratio for font-sizes to improve legibility. 

image

image

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -11,7 +11,7 @@
     -->
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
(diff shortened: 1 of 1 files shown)


andresmoschini:
Merging in order of continuing with other PRs.

PR #209, opened 2019-05-16
DBR-260 - Clean language files

andresmoschini:
Hi team, 

I have applied some minor changes in language files, removing unused entries and having a constant for Advanced Reports URL.

I have also found that the content of "activate trial" page in Spanish laked of the help link, so, I have added.

Could you review?

### DIFF
--- a/src/i18n/es.js
+++ b/src/i18n/es.js
@@ -26,7 +26,6 @@ export default {
   common: {
     cancel: `Cancelar`,
     copyright_MD: `© ${year} Doppler LLC. Todos los derechos reservados. [Política de Privacidad y Legales](${urlPrivacy}).`,
-    help: `Ayuda`,
     hide: `Ocultar`,
     message: `Mensaje`,
     recaptcha_legal_HTML: `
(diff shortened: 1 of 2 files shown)


src/i18n/es.js
```diff
@@ -92,31 +89,19 @@ export default {
       <p>
         Activa el periodo de prueba y accede a Reportes detallados sobre el comportamiento de los
         usuarios en tu Sitio Web o E-commerce. Descubre cuáles son las páginas más visitadas, cuántos
-        visitantes poseen un Email que Doppler ha identificado y cuántos no.
+        visitantes poseen un Email que Doppler ha identificado y cuántos no. ¿Necesitas ayuda?
+        <a target="_blank" href="${urlHelpAdvancedReports}" ${patchForBlank}>HELP</a>.
```
andresmoschini: Hi @carodipietro, I have added this text in Spanish on the page of trial activation.

I think that we have missed it before. 

image

What do you think?

src/i18n/es.js
```diff
@@ -149,7 +134,7 @@ export default {
         Descubre cuáles son las páginas más visitadas de tu Sitio Web o E-commerce, cuántos
         visitantes poseen un Email que Doppler ha identificado y cuántos no. ¡Sigue el recorrido de
         los usuarios, detecta puntos de fuga y oportunidades de mejora! Si necesitas ayuda, presiona
```
GusBaamonde: Si necesitas ayuda, presiona Ayuda! Suena a Moria, si queres llorar llora :)

andresmoschini: jeje, así está en los demás.

GusBaamonde: Suena raro

src/i18n/es.js
andresmoschini: Mergeo, si está mal, me avisan y revierto.

PR #213, opened 2019-05-16
It is only another ugly patch to force a build :P

andresmoschini:


andresmoschini:
But it did not work :(

PR #217, opened 2019-05-21
DBR-171 - Do not show validation errors before first submit

andresmoschini:
Hi team, 

As we have talked, I changed the behavior of the form in order to do not show validation errors before the first form submission.

Could you review it?

--- 

image

No, no, come on @matiasbeckerle, this time I added tests!

### DIFF
--- a/src/components/UpgradePlanForm/UpgradePlanForm.js
+++ b/src/components/UpgradePlanForm/UpgradePlanForm.js
@@ -73,7 +73,7 @@ class UpgradePlanForm extends React.Component {
             validate={validate}
             onSubmit={onSubmit}
           >
-            {({ isSubmitting, errors, touched }) => (
+            {({ errors, touched, submitCount }) => (
               <Form className="form-request">
                 <fieldset>
                   <ul className="field-group">
@@ -98,7 +98,9 @@ class UpgradePlanForm extends React.Component {
                     <li
                       className={
                         'field-item' +
-                        (touched[fieldNames.message] && errors[fieldNames.message] ? ' error' : '')
+                        (submitCount && touched[fieldNames.message] && errors[fieldNames.message]
+                          ? ' error'
+                          : '')
                       }
                     >
                       <label htmlFor={fieldNames.message}>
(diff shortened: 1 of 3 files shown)


src/components/UpgradePlanForm/UpgradePlanForm.js
```diff
@@ -98,7 +98,9 @@ class UpgradePlanForm extends React.Component {
                     <li
                       className={
                         'field-item' +
-                        (touched[fieldNames.message] && errors[fieldNames.message] ? ' error' : '')
+                        (submitCount && touched[fieldNames.message] && errors[fieldNames.message]
```
cbernat: I like this magic submitCount that comes with formik. 😄

cbernat:
Very nice! 👍

src/components/UpgradePlanForm/UpgradePlanForm.js
andresmoschini: Haha, it is not magic: all changes are explicit and no convention based.

src/components/form-helpers/form-helpers.js
```diff
@@ -155,13 +155,16 @@ const ErrorMessage = injectIntl(({ intl, error }) =>
 );
 
 export const FieldItem = connect(
-  ({ className, fieldName, children, formik: { errors, touched } }) => (
+  ({ className, fieldName, children, formik: { errors, touched, submitCount } }) => (
     <li
-      className={concatClasses(className, touched[fieldName] && errors[fieldName] ? 'error' : '')}
+      className={concatClasses(
+        className,
+        submitCount && touched[fieldName] && errors[fieldName] ? 'error' : '',
```
fcoronelmakingsense: if you put in fieldItem its apply for all forms isnt??

andresmoschini: Almost. Because we are not using _FieldItem_ in all forms.

I searched for _touched_ and I also found _UpgradePlanForm_.

fcoronelmakingsense: but if in future want one form validate before first submit need change right? its for understand only

andresmoschini: Right, in that scenario we could avoid using _FieldItem_, or we could use a flag in _FieldItem_.

PR #218, opened 2019-05-21
DBR-313 - Semantic Release

andresmoschini:
Hi team,

This is an end-user application, so, maybe semantic versioning is not the best option, but maybe it is useful because we need number versions, and it is automatic and gives us some kind of information.

Following (in part) [this article](https://medium.com/@u_glow/making-sense-of-semantic-versioning-for-end-user-software-applications-a3049d97478b) we can apply these rules:

* `MAJOR` version  (`BREAKING CHANGES`): when we make changes incompatible with current Doppler Backend, DataHub ApiSec, or other backend APIs. Also when we remove or change drastically some end-user features. And also when our changes affect the installation-requirements or infrastructure.
* `MINOR` version (`feat`): when we add functionality in a backwards-compatible manner (e.g. passing additional data to an already-provisioned API or adding any end-user functionality that does not affect the installation-requirements)
* `PATCH` version (`fix`): when we make backwards-compatible bug fixes (e.g. fixing any end-user bug that does not affect the installation requirements).

Regarding performance or security improvements, in my opinion, we need to consider them as _fixes_ or even _features_, in order to increment the version number.

What do you think?

Please review it well, because I cannot test it before merging.

### DIFF
--- a/build-n-release-w-docker.sh
+++ b/build-n-release-w-docker.sh
@@ -1,0 +1,86 @@
+#!/bin/sh
+
+pkgName="doppler-webapp"
+cdnBaseUrl=${2:-"//cdn.fromdoppler.com/$pkgName"}
+pkgVersionQaPrefix="qa-"
+pkgVersionIntPrefix="int-"
+pkgVersionDevelopmentPrefix="dev-"
+
+# Stop script on NZEC
+set -e
+# Stop script if unbound variable found (use ${var:-} if intentional)
+set -u
+
+# Lines added to get the script running in the script path shell context
+# reference: http://www.ostricher.com/2014/10/the-right-way-to-get-the-directory-of-a-bash-script/
+cd $(dirname $0)
+
+# To avoid issues with MINGW y Git Bash, see:
+# https://github.com/docker/toolbox/issues/673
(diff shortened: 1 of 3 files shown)


build-n-release-w-docker.sh
```diff
@@ -0,0 +1,88 @@
...
+
+echo Publishing to Akamai...
+echo pkgName: $pkgName
+echo pkgVersion: $pkgVersion
+echo cdnBaseUrl: $cdnBaseUrl
+
+# Force pull the latest image version due to the cache not always is pruned immediately after an update is uploaded to docker hub
+docker pull dopplerrelay/doppler-relay-akamai-publish
+
+docker run --rm \
```
cbernat: This is production build right?

andresmoschini: Mmm, it depends on what is "production" for you.

It copies the result files of production build (the result of `...&& yarn semantic-release`) to our CDN.

Does it answer your question?

andresmoschini:
test this please

build-n-release-w-docker.sh
cbernat: So it is production then 👍

PR #221, opened 2019-05-23
DBR-321 - Fix issue with browser language

andresmoschini:
Hi team, 

I received a message from support area, about a user that cannot use our signup page. 

I am pretty sure that his browser has a different language configured than English or Spanish.

So, in my opinion, it should fix the issue.

Could you review the changes?

I promise to add a test for this issue today, but I want to merge it soon in order to deploy the fix to production ASAP.

### DIFF
--- a/src/i18n/DopplerIntlProvider.js
+++ b/src/i18n/DopplerIntlProvider.js
@@ -14,7 +14,10 @@ const messages = {
 addLocaleData([...en, ...es]);
 
 export default ({ locale, children }) => (
-  <IntlProvider locale={locale || 'en'} messages={flattenMessages(messages[locale || 'en'])}>
+  <IntlProvider
+    locale={locale || 'en'}
+    messages={flattenMessages(messages[locale || 'en'] || messages['en'])}
+  >
     {children}
   </IntlProvider>
 );
--- a/src/index.js
+++ b/src/index.js
@@ -21,7 +21,9 @@ if (document.querySelector('body').setActive) {
 }
 
 // TODO: this hardcoded data will depend by the app language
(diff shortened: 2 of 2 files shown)


src/i18n/DopplerIntlProvider.js
```diff
@@ -14,7 +14,10 @@ const messages = {
 addLocaleData([...en, ...es]);
 
 export default ({ locale, children }) => (
-  <IntlProvider locale={locale || 'en'} messages={flattenMessages(messages[locale || 'en'])}>
+  <IntlProvider
+    locale={locale || 'en'}
+    messages={flattenMessages(messages[locale || 'en'] || messages['en'])}
```
fcoronelmakingsense: only format change?

src/index.js
```diff
@@ -21,7 +21,9 @@ if (document.querySelector('body').setActive) {
 }
 
 // TODO: this hardcoded data will depend by the app language
-const locale = navigator.language.toLowerCase().split(/[_-]+/)[0] || 'en';
+const localeFromNavigator = navigator.language.toLowerCase().split(/[_-]+/)[0] || 'en';
+// TODO: remove language validation code
+const locale = ['es', 'en'].includes(localeFromNavigator) ? localeFromNavigator : 'en';
```
fcoronelmakingsense: good catch. Maybe you right they have another language

src/i18n/DopplerIntlProvider.js
andresmoschini: No, now when `messages[locale || 'en']` is _falsy_ we will use `messages['en']`

andresmoschini:
Thanks for your reviews! I really appreciate it.

PR #222, opened 2019-05-23
DBR-321 - Fix issue with browser language (tests and refactoring)

andresmoschini:
Hi team,

As I promised in #221, I am adding tests and some minor refactoring related to language behavior.

Based on a @fgchaio comment, I am also changing the default language for Spanish.

Could you review it?

### DIFF
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -90,7 +90,7 @@ describe('App component', () => {
       getByText('Política de Privacidad y Legales');
     });
 
-    it('should use English when language is not supported', () => {
+    it('should use Spanish when language is not supported', () => {
       // Arrange
       const dependencies = {
         sessionManager: createDoubleSessionManager(),
@@ -106,7 +106,8 @@ describe('App component', () => {
       );
 
       // Assert
-      getByText('Privacy Policy & Legals');
+      getByText('Política de Privacidad y Legales');
+    });
     });
 
     it("should be updated based on user's data", () => {
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -90,6 +90,25 @@ describe('App component', () => {
       getByText('Política de Privacidad y Legales');
     });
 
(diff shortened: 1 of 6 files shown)


src/App.test.js
```diff
@@ -90,6 +90,25 @@ describe('App component', () => {
...
+      // Arrange
+      const dependencies = {
+        sessionManager: createDoubleSessionManager(),
+      };
+
+      // Act
+      const { getByText } = render(
+        <AppServicesProvider forcedServices={dependencies}>
+          <Router initialEntries={['/login']}>
+            <App locale="fr" />
```
cbernat: 👍

src/i18n/utils.js
```diff
@@ -12,3 +12,11 @@ export function flattenMessages(nestedMessages, prefix = '') {
...
   }, {});
 }
+
+export const defaultLanguage = 'en';
+
+export const availableLanguageOrNull = (lang) =>
+  (lang && ['es', 'en'].includes(lang) && lang) || null;
+
+export const availableLanguageOrDefault = (lang) =>
+  availableLanguageOrNull(lang) || defaultLanguage;
```
cbernat: Why not have all in one function? `availableLanguageOrNull` as part of `availableLanguageOrDefault`

andresmoschini: Because in a place I need _null_ result. I did not analyze if it is really necessary (I think so) or not.

cbernat: You think you might need it, ok.

cbernat:
👍

src/i18n/utils.js
andresmoschini: Here: https://github.com/FromDoppler/doppler-webapp/pull/222/files#diff-14b1e33d5bf5649597cdc0e4f684daddR48

PR #223, opened 2019-05-24
fix: reports ranking spinner

fcoronelmakingsense:
- Fix spinner in reports page when filter for domains, pages or period
- Use <Loading> component


src/components/Reports/Reports.js
```diff
@@ -77,7 +78,7 @@ class Reports extends React.Component {
             changePeriod={this.changePeriod}
           />
           {!this.state.domains ? (
-            <div className="loading-box" />
+            <Loading />
```
andresmoschini: ❤️

PR #227, opened 2019-05-27
feat: modification of messages and validations

GusBaamonde:
**Implemented changes in icon sizes**

mensajes-paginas-publicas2

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -12,7 +12,7 @@
     -->
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
-    <link rel="stylesheet" type="text/css" href="https://cdn.fromdoppler.com/doppler-ui-library/v1.0.2/css/styles.css">
+    <link rel="stylesheet" type="text/css" href="https://cdn.fromdoppler.com/doppler-ui-library/build289/css/styles.css">
     <link rel="stylesheet" type="text/css" href="zoho-chat.css">
     <!--
       Notice the use of %PUBLIC_URL% in the tags above.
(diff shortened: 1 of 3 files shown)


andresmoschini:
@santiagopaolucci would you like to play a little with this? https://cdn.fromdoppler.com/doppler-webapp/int-build742/#/login/reset-password

cbernat:
👍 Very nice!

GusBaamonde:
@santiagopaolucci please confirm me the changes we talked about, thanks!

public/index.html
```diff
@@ -12,7 +12,7 @@
     -->
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
-    <link rel="stylesheet" type="text/css" href="https://cdn.fromdoppler.com/doppler-ui-library/v1.0.2/css/styles.css">
+    <link rel="stylesheet" type="text/css" href="https://cdn.fromdoppler.com/doppler-ui-library/build292/css/styles.css">
```
andresmoschini: Remember update it with the final version before merge.

src/components/ForgotPassword/ForgotPassword.js
```diff
@@ -106,9 +106,11 @@ const ForgotPassword = ({ intl, location, dependencies: { dopplerLegacyClient }
         <h5>{_('login.forgot_password')}</h5>
         <p className="content-subtitle">{_('forgot_password.description')}</p>
         {sentEmail ? (
-          <div className="forgot-message bounceIn">
-            <FormattedHTMLMessage tagName="div" id="forgot_password.confirmation_message_HTML" />
-            <LinkToLoginSuccess />
+          <div className="forgot-message">
```
andresmoschini: Why do we have a _forgot-message_ class? it does not look as something reusable.

Anyway, this was already here, so fixing it is out of the scope of this PR.

andresmoschini:
Nice work, the changes are simple and looks nice, thanks!

src/components/ForgotPassword/ForgotPassword.js
GusBaamonde: We use this class within the web application so that when you clean the form, keep the height and do not lift your foot sharply. It is something that we could eliminate and look for a better way that does not happen.

andresmoschini:
❤️ 
🚀 :shipit:

PR #230, opened 2019-06-04
feat: Refactored language selector

damianmuti:
Since we needed to improve keyboard navigation on login/reset-password pages, I've refactored the language selector to support keyboard events.

image
image

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -12,7 +12,7 @@
     -->
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
-    <link rel="stylesheet" type="text/css" href="https://cdn.fromdoppler.com/doppler-ui-library/build308/css/styles.css">
+    <link rel="stylesheet" type="text/css" href="https://cdn.fromdoppler.com/doppler-ui-library/v2.0.0/css/styles.css">
     <link rel="stylesheet" type="text/css" href="zoho-chat.css">
     <!--
       Notice the use of %PUBLIC_URL% in the tags above.
@@ -47,7 +47,7 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
-    <script type="text/javascript" src="https://cdn.fromdoppler.com/doppler-ui-library/build308/js/app.js"></script>
+    <script type="text/javascript" src="https://cdn.fromdoppler.com/doppler-ui-library/v2.0.0/js/app.js"></script>
     <script type="text/javascript" src="zoho-chat.js"></script>
     <link rel="stylesheet" type="text/css" href="zoho-chat.css">
   </body>
--- a/public/index.html
+++ b/public/index.html
@@ -12,7 +12,8 @@
     -->
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
-    <link rel="stylesheet" type="text/css" href="https://cdn.fromdoppler.com/doppler-ui-library/v1.1.1/css/styles.css">
+    <link rel="stylesheet" type="text/css" href="https://cdn.fromdoppler.com/doppler-ui-library/build308/css/styles.css">
+    <link rel="stylesheet" type="text/css" href="zoho-chat.css">
     <!--
       Notice the use of %PUBLIC_URL% in the tags above.
       It will be replaced with the URL of the `public` folder during the build.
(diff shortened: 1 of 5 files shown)


public/index.html
```diff
@@ -47,7 +47,7 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
-    <script type="text/javascript" src="https://cdn.fromdoppler.com/doppler-ui-library/v1.1.1/js/app.js"></script>
+    <script type="text/javascript" src="https://cdn.fromdoppler.com/doppler-ui-library/build308/js/app.js"></script>
```
fcoronelmakingsense: dont forget before merge put the correct version

src/components/shared/LanguageSelector/LanguageSelector.js
```diff
@@ -9,29 +9,52 @@ import { injectIntl } from 'react-intl';
  */
 const LanguageSelector = ({ intl }) => {
   const lang = intl.locale;
+  const languageDropdown = useRef(null);
+
+  useEffect(() => {
+    const menubutton = new window.Menubutton(languageDropdown.current);
```
fcoronelmakingsense: maybe can add a comment with mention "MenuButton" is define in UI Library

public/index.html
andresmoschini: @damianmuti since the merge has been already done in the _UI Library_ you can update it right now.

src/components/shared/LanguageSelector/LanguageSelector.js
```diff
@@ -9,29 +9,52 @@ import { injectIntl } from 'react-intl';
  */
 const LanguageSelector = ({ intl }) => {
   const lang = intl.locale;
+  const languageDropdown = useRef(null);
```
andresmoschini:
```suggestion
  const languageDropdownEl = useRef(null);
```

andresmoschini:
```suggestion
  const languageDropdownRef = useRef(null);
```

src/components/shared/LanguageSelector/LanguageSelector.js
damianmuti: That would be a good idea 👍

src/components/shared/LanguageSelector/LanguageSelector.js
damianmuti: That's a good idea, @andresmoschini . After @cbernat is finish fixing some tests that are currently failing, I'll do that improvement 👍

src/components/Signup/Signup.test.js
```diff
@@ -6,13 +6,24 @@ import { MemoryRouter as Router } from 'react-router-dom';
 import { AppServicesProvider } from '../../services/pure-di';
 import 'jest-dom/extend-expect';
 
+class MenubuttonFake {
+  // eslint-disable-next-line no-useless-constructor
+  constructor() {}
```
andresmoschini: Maybe declaring the constructor is not necessary, could you try removing it?

src/App.test.js
```diff
@@ -48,6 +48,11 @@ const RouterInspector = withRouter(({ match, location, history, target }) => {
   return null;
 });
 
+class MenubuttonFake {
+  constructor() {}
```
andresmoschini: Maybe declaring the constructor is not necessary, could you try removing it?

andresmoschini:
Nice work @cbernat!

I would like to be sure about how it works, I would like to test it with some console log to be sure if we are understanding t well and not having memory leaks. Maybe you could do that if you want to merge it sooner.

---

Refactor it could be another useful thing, because there is a lot of duplicated code:

image

Maybe @damianmuti could do it for himself, I can help him today.

src/components/Signup/Signup.test.js
cbernat: Sure, I thought I needed this since the tests gave an error about a constructor missing. I've tested and it is not neccesary.

andresmoschini:
Nice refactoring!

andresmoschini:
I am pretty sure that there are not memory leaks, and if there are, they will not be important. I will be very happy when this code be merged.

Thanks for your work @damianmuti and @cbernat !

src/components/shared/LanguageSelector/LanguageSelector.js
```diff
@@ -1,41 +1,51 @@
...
-            </Link>
-          </div>
-        </>
-      )}
+    <div className="dp-c-dropdown language-selector">
+      <button
+        className={'lang--' + lang}
+        id="menubutton"
+        aria-haspopup="true"
+        aria-controls="menu2"
```
fcoronelmakingsense: the aria attr it's use in UI library?

damianmuti: It is, @fcoronelmakingsense , or at least we are using it in components that need some user interaction, such as this one.

PR #232, opened 2019-06-07
fix: remove zoho bubble in private pages

fcoronelmakingsense:
- Remove zoho help bubble in private pages. only show in public pages
- Cant use ref to access to DOM because the element is create by zoho widget. I dont know if exist a better solution for this case because dont have control for js widget.

image

image


fcoronelmakingsense:
Update solution using helmet

PR #235, opened 2019-06-10
Shopify: manage connected/disconnected states

cbernat:
Shopify section:
- Add Shopify client and shopify hardcoded client
- Add tests
- Add view with title only, access via /#/integrations/shopify
- Add header without section

image

https://cdn.fromdoppler.com/doppler-webapp/build883/#/integrations/shopify

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -4,8 +4,8 @@ import HeaderMessages from './HeaderMessages/HeaderMessages';
 import HeaderUserMenu from './HeaderUserMenu/HeaderUserMenu';
 import { FormattedMessage } from 'react-intl';
 
-const Header = ({ userData: { user, nav, alert }, location: {pathname} }) => {
-  const inactiveSection = pathname.match(/^\/integrations\/*/) !== null; 
+const Header = ({ userData: { user, nav, alert }, location: { pathname } }) => {
+  const inactiveSection = pathname.match(/^\/integrations\/*/) !== null;
   return (
     <div>
       {alert ? <HeaderMessages alert={alert} user={user} /> : null}
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -4,14 +4,20 @@ import HeaderMessages from './HeaderMessages/HeaderMessages';
 import HeaderUserMenu from './HeaderUserMenu/HeaderUserMenu';
 import { FormattedMessage } from 'react-intl';
 
-const Header = ({ userData: { user, nav, alert } }) => {
+const Header = ({ userData: { user, nav, alert }, location: {pathname} }) => {
+  const inactiveSection = pathname.match(/^\/integrations\/*/) !== null; 
   return (
     <div>
       {alert ? <HeaderMessages alert={alert} user={user} /> : null}
       {/* //TODO: Refactor backend to send proper active values. Class 'header-is-active' must be removed */}
       <header
         className={
-          user.clientManager ? 'header-main header-open dp-header--cm' : 'header-main header-open'
+          'header-main' +
+          (!inactiveSection
+            ? user.clientManager
+              ? ' header-open dp-header--cm'
+              : ' header-open'
+            : '')
         }
       >
         {user.clientManager ? (
@@ -28,7 +34,7 @@ const Header = ({ userData: { user, nav, alert } }) => {
           <div className="logo">
             <span className="ms-icon icon-doppler-logo" />
           </div>
-          <HeaderNav nav={nav} />
+          <HeaderNav nav={nav} inactiveSection={inactiveSection} />
           <nav className="nav-right-main">
             <ul className="nav-right-main--list">
               <li>
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -6,44 +6,45 @@ import Loading from '../../Loading/Loading';
 const shopifyClient = new HardcodedShopifyClient();
 
 const Shopify = () => {
-  const [shopName, setShopName] = useState(null);
+  const [shopName, setShopName] = useState('');
+  const [isConnected, setIsConnected] = useState(false);
   const [error, setError] = useState(null);
+  const [isLoading, setIsLoading] = useState(true);
+
   useEffect(() => {
     const getData = async () => {
       const result = await shopifyClient.getShopifyData();
       if (result.success && result.value.length) {
+        setIsConnected(true);
         setShopName(result.value[0].shopName);
-      } else if (result.success) {
-        setShopName('');
       } else if (
         !result.success &&
         result.expectedError &&
         result.expectedError.cannotConnectToAPI
       ) {
         setError('Error: No hemos podido conectar con la Api de Shopify, vuelve a intentar luego.');
-      } else {
+      } else if (!result.success) {
         setError('Error: Error inesperado.');
       }
+      setIsLoading(false);
     };
     getData();
   }, []);
 
   return (
     <>
-      <Helmet>
-        <title>Doppler | Shopify</title>
-      </Helmet>
+      <Helmet title={'Doppler | Shopify'} />
       <section style={{ width: '100%', padding: '60px 30px' }}>
         Panel de control | Integraciones y preferencias avanzadas
         <h1>Panel de Control - Shopify</h1>
-        {shopName === null && error === null ? (
+        {isLoading ? (
           <Loading />
-        ) : shopName !== null && shopName.length ? (
-          'Shopify conectado a tienda: ' + shopName
-        ) : error === null ? (
-          'Shopify desconectado'
-        ) : (
+        ) : error ? (
           error
+        ) : isConnected ? (
+          'Shopify conectado a tienda: ' + shopName
+        ) : (
+          'Shopify desconectado'
         )}
       </section>
     </>
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -12,17 +12,13 @@ const Shopify = ({ dependencies: { shopifyClient } }) => {
   useEffect(() => {
     const getData = async () => {
       const result = await shopifyClient.getShopifyData();
-      if (result.success && result.value.length) {
-        setIsConnected(true);
-        setShops(result.value);
-      } else if (
-        !result.success &&
-        result.expectedError &&
-        result.expectedError.cannotConnectToAPI
-      ) {
+      if (result.expectedError && result.expectedError.cannotConnectToAPI) {
         setError('Error: No hemos podido conectar con la Api de Shopify, vuelve a intentar luego.');
       } else if (!result.success) {
         setError('Error: Error inesperado.');
+      } else if (result.value.length) {
+        setIsConnected(true);
+        setShops(result.value);
       }
       setIsLoading(false);
     };
@@ -32,6 +28,7 @@ const Shopify = ({ dependencies: { shopifyClient } }) => {
   return (
     <>
       <Helmet title={'Doppler | Shopify'} />
+      {/* inline style will be removed in next PR */}
       <section style={{ width: '100%', padding: '60px 30px' }}>
         Panel de control | Integraciones y preferencias avanzadas
         <h1>Panel de Control - Shopify</h1>
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -1,9 +1,9 @@
 import React, { useState, useEffect } from 'react';
 import { Helmet } from 'react-helmet';
 import Loading from '../../Loading/Loading';
-import {InjectAppServices} from '../../../services/pure-di'
+import { InjectAppServices } from '../../../services/pure-di';
 
-const Shopify = ({dependencies: {shopifyClient} }) => {
+const Shopify = ({ dependencies: { shopifyClient } }) => {
   const [shops, setShops] = useState([]);
   const [isConnected, setIsConnected] = useState(false);
   const [error, setError] = useState(null);
@@ -38,7 +38,7 @@ const Shopify = ({dependencies: {shopifyClient} }) => {
         {isLoading ? (
           <Loading />
         ) : error ? (
-          error
+          <p>{error}</p>
         ) : isConnected ? (
           <>
             <p>Shopify conectado a: </p>
@@ -49,7 +49,7 @@ const Shopify = ({dependencies: {shopifyClient} }) => {
             </ul>
           </>
         ) : (
-          'Shopify desconectado'
+          <p>Shopify desconectado</p>
         )}
       </section>
     </>
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -1,11 +1,9 @@
 import React, { useState, useEffect } from 'react';
 import { Helmet } from 'react-helmet';
-import { HardcodedShopifyClient } from '../../../services/shopify-client.doubles';
 import Loading from '../../Loading/Loading';
+import {InjectAppServices} from '../../../services/pure-di'
 
-const shopifyClient = new HardcodedShopifyClient();
-
-const Shopify = () => {
+const Shopify = ({dependencies: {shopifyClient} }) => {
   const [shops, setShops] = useState([]);
   const [isConnected, setIsConnected] = useState(false);
   const [error, setError] = useState(null);
@@ -58,4 +56,4 @@ const Shopify = () => {
   );
 };
 
-export default Shopify;
+export default InjectAppServices(Shopify);
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -1,0 +1,42 @@
+import React, { useState, useEffect } from 'react';
+import { Helmet } from 'react-helmet';
+import { HardcodedShopifyClient } from '../../../services/shopify-client.doubles';
+import Loading from '../../Loading/Loading';
+
+const shopifyClient = new HardcodedShopifyClient();
+
+const Shopify = () => {
+  const [shopName, setShopName] = useState(null);
+  useEffect(() => {
+    const getData = async () => {
+      const result = await shopifyClient.getShopifyData();
+      if (result.length) {
+        setShopName(result[0].shopName);
+      } else {
+        setShopName('');
+      }
+    };
+    getData();
+  }, []);
+
+  return (
+    <>
+      <Helmet>
+        <title>Doppler | Shopify</title>
+      </Helmet>
+      <section style={{ width: '100%', padding: '60px 30px' }}>
+        Panel de control | Integraciones y preferencias avanzadas
+        <h1>Panel de Control - Shopify</h1>
+        {shopName === null ? (
+          <Loading />
+        ) : shopName.length ? (
+          'Shopify conectado a tienda: ' + shopName
+        ) : (
+          'Shopify desconectado'
+        )}
+      </section>
+    </>
+  );
+};
+
+export default Shopify;
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -27,7 +27,7 @@ const Shopify = ({ dependencies: { shopifyClient } }) => {
       setIsLoading(false);
     };
     getData();
-  }, []);
+  }, [shopifyClient]);
 
   return (
     <>
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -7,13 +7,22 @@ const shopifyClient = new HardcodedShopifyClient();
 
 const Shopify = () => {
   const [shopName, setShopName] = useState(null);
+  const [error, setError] = useState(null);
   useEffect(() => {
     const getData = async () => {
       const result = await shopifyClient.getShopifyData();
-      if (result.length) {
-        setShopName(result[0].shopName);
-      } else {
+      if (result.success && result.value.length) {
+        setShopName(result.value[0].shopName);
+      } else if (result.success) {
         setShopName('');
+      } else if (
+        !result.success &&
+        result.expectedError &&
+        result.expectedError.cannotConnectToAPI
+      ) {
+        setError('Error: No hemos podido conectar con la Api de Shopify, vuelve a intentar luego.');
+      } else {
+        setError('Error: Error inesperado.');
       }
     };
     getData();
@@ -27,12 +36,14 @@ const Shopify = () => {
       <section style={{ width: '100%', padding: '60px 30px' }}>
         Panel de control | Integraciones y preferencias avanzadas
         <h1>Panel de Control - Shopify</h1>
-        {shopName === null ? (
+        {shopName === null && error === null ? (
           <Loading />
-        ) : shopName.length ? (
+        ) : shopName !== null && shopName.length ? (
           'Shopify conectado a tienda: ' + shopName
-        ) : (
+        ) : error === null ? (
           'Shopify desconectado'
+        ) : (
+          error
         )}
       </section>
     </>
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -46,7 +46,7 @@ const Shopify = () => {
             <p>Shopify conectado a: </p>
             <ul>
               {shops.map((shop) => (
-                <li> {shop.shopName} </li>
+                <li key={shop.shopName}> {shop.shopName} </li>
               ))}{' '}
             </ul>
           </>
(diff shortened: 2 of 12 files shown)


src/services/shopify-client.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface ShopifyConnection {
+  shop_name: string;
+  synchronization_date: Date | null;
+  list: SubscriberList;
+  connected: boolean;
+}
+
+export interface ShopifyClient {
+  getShopifyData(email: string, apiKey: string): Promise<ShopifyConnection>;
+  disconnect(email: string, apikey: string): Promise<boolean>;
```
andresmoschini: Why does it return a boolean? If the idea is return _false_ when there is an error, I would suggest ussing something like [EmptyResult<TError>](https://github.com/FromDoppler/doppler-webapp/blob/master/src/services/doppler-legacy-client.ts#L19).

Maybe those classes could be moved to a common place.

src/services/shopify-client.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export interface ShopifyConnection {
+  shop_name: string;
+  synchronization_date: Date | null;
+  list: SubscriberList;
+  connected: boolean;
+}
+
+export interface ShopifyClient {
+  getShopifyData(email: string, apiKey: string): Promise<ShopifyConnection>;
```
andresmoschini: If I am not wrong, there could be more than a shop connected, what do you think about:

```suggestion
  getShopifyData(email: string, apiKey: string): Promise<ShopifyConnection[]>;
```

src/services/shopify-client.doubles.ts
```diff
@@ -0,0 +1,47 @@
+import {
```
andresmoschini: Si no me equivoco, ni _ShopifyClient_, ni _HardcodedShopifyClient_ se están usando, tal vez convenga mergear solo la ruta y la página.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Helmet } from 'react-helmet';
+
+const Shopify = () => (
+  <>
+    <Helmet>
+      <title>Doppler | Shopify </title>
+    </Helmet>
+    <div> Control panel page for shopify</div>
```
```suggestion
    <div>Control panel page for Shopify</div>
```

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Helmet } from 'react-helmet';
+
+const Shopify = () => (
+  <>
+    <Helmet>
+      <title>Doppler | Shopify </title>
```
```suggestion
      <title>Doppler | Shopify</title>
```

src/services/shopify-client.ts
```diff
@@ -0,0 +1,25 @@
+export enum SubscriberListState {
+  ready,
+  synchronizing_contacts,
```
andresmoschini: What do you think about:

```suggestion
  synchronizingContacts,
```

src/services/shopify-client.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export interface SubscriberList {
+  name: string;
+  id: number;
+  amount_subscribers: number;
```
andresmoschini: What do you think about:

```suggestion
  subscribersCount: number;
```

src/services/shopify-client.ts
```diff
@@ -0,0 +1,25 @@
...
+  state: SubscriberListState;
+}
+
+export interface ShopifyConnection {
+  shop_name: string;
```
andresmoschini: What do you think about:

```suggestion
  shopName: string;
```

src/services/shopify-client.ts
cbernat: About the EmptyResult, yes maybe it's better. Which classes? I don't get you with this.

src/services/shopify-client.doubles.ts
cbernat: Puede ser, o lo que me comentaste por chat, podría subir solamente que muestre si esta integrado o no. Usando el client.

src/services/shopify-client.ts
cbernat: We don't use underscores anywhere right?

andresmoschini: Mmm, no, in general, we are using camel case.

cbernat:
Test this please.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -1,16 +1,33 @@
...
-  </>
-);
+const shopifyClient = new HardcodedShopifyClient();
+
+const Shopify = () => {
+  const [shopName, setShopName] = useState('');
+  useEffect(() => {
+    const getData = async () => {
+      const result = await shopifyClient.getShopifyData();
+      if (result.length) {
```
andresmoschini: What happens in the other case?

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -1,16 +1,33 @@
...
+        setShopName(result[0].shopName);
+      }
+    };
+    getData();
+  }, []);
```
andresmoschini: Why do you need `getData` function? I think that it will work in the same way:

```js
  useEffect(() => {
      const result = await shopifyClient.getShopifyData();
      if (result.length) {
        setShopName(result[0].shopName);
      }
  }, []);
```

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -1,16 +1,33 @@
...
+
+  return (
+    <>
+      <Helmet>
+        <title>Doppler | Shopify</title>
+      </Helmet>
+      <section style={{ width: '100%', padding: '60px 30px' }}>
+        Panel de control | Integraciones y preferencias avanzadas
+        <h1>Panel de Control - Shopify</h1>
+        {shopName.length ? 'Shopify conectado a tienda: ' + shopName : 'Shopify desconectado'}
```
andresmoschini: I think that you are missing some possible states:
- [ ] Undetermined (when ajax call did not finished yet)
- [x] When there is not a shop
- [x] When there is a shop
- [ ] When there was an error in ajax call

src/services/shopify-client.doubles.ts
```diff
@@ -15,12 +10,12 @@ const fakeData = [
...
       state: SubscriberListState.ready,
-    }
-  }
+    },
+  },
 ];
 
 export class HardcodedShopifyClient implements ShopifyClient {
-  public async getShopifyData(email: string, apiKey: string): Promise<ConnectedShops[]> {
+  public async getShopifyData(): Promise<ConnectedShops[]> {
```
andresmoschini: In my opinion, the result type should be a wrapper that takes into account errors.

cbernat: Yes you are right, I will make a small refactor to reuse dopplerLegacyClient results. In a separate commit.

src/components/Integrations/Shopify/Shopify.js
cbernat: You are clearly not connected, without shops that is.

src/components/Integrations/Shopify/Shopify.js
cbernat: I need to tell the function is async somewhere 
image

cbernat: I've tried adding async into useEffect hook but it also gives error, in this post https://www.robinwieruch.de/react-hooks-fetch-data/#react-hooks-programmatically says that the ways is with a function.

andresmoschini: Right, I would like to understand why the `useEffect(async () => {` does not work, do you have a screenshot of the error? and if it does not work, could you explain to me in a short way? (I have still pending reading the article)

src/components/Integrations/Shopify/Shopify.js
andresmoschini: In that case, I think that you should make it explicit.

Maybe something like this:

```js
  const [shopName, setShopName] = useState(null);
  // shopName === null would mean "loading"
  useEffect(() => {
    const getData = async () => {
      const result = await shopifyClient.getShopifyData();
      if (result.length) {
        setShopName(result[0].shopName);
        // shopName === something would mean "there is a shop"
      } else {
        setShopName('');
        // shopName === '' would mean "there is not a shop"
    };
    getData();
  }, []);
```

I think that it could work, but maybe it would be better having more than a _state_ in order to make it more explicit. Or having something different, for example an object:


```js
  cont _thereIsNoShop = {};
  // . . .
  const [shop, setShop] = useState(null);
  // shop === null would mean "loading"
  useEffect(() => {
    const getData = async () => {
      const result = await shopifyClient.getShopifyData();
      if (result.length) {
        setShopName(result[0]);
        // shop would contains "the shop".
      } else {
        setShop(_thereIsNoShop);
        // shop === _thereIsNoShop would mean that there is no job
    };
    getData();
  }, []);
```

src/components/Integrations/Shopify/Shopify.js
cbernat: According to the documentation every function annotated with async returns an implicit promise: “The async function declaration defines an asynchronous function, which returns an AsyncFunction object. An asynchronous function is a function which operates asynchronously via the event loop, using an implicit Promise to return its result. “. However, an effect hook should return *nothing or a clean up function*. That’s why you may see the following warning in your developer console log: 07:41:22.910 index.js:1452 Warning: useEffect function must return a cleanup function or nothing. Promises and useEffect(async () => …) are not supported, but you can call an async function inside an effect.. That’s why using async directly in the useEffect function isn’t allowed.

andresmoschini: Oh, I understand now, thank you very much for the clear explanation!

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -7,13 +7,25 @@ const shopifyClient = new HardcodedShopifyClient();
...
+          setShopName('');
+        } else {
+          if (result.expectedError && result.expectedError.cannotConnectToAPI) {
+            setError(
+              'Error: No hemos podido conectar con la Api de Shopify, vuelve a intentar luego.',
+            );
+          } else {
+            setError('Error: Error inesperado.');
+          }
+        }
```
andresmoschini: que opinás de reescribirlo de esta manera para reducir la anidación?

```js
        } else  if (result.expectedError && result.expectedError.cannotConnectToAPI) {
          setError(
            'Error: No hemos podido conectar con la Api de Shopify, vuelve a intentar luego.',
          );
        } else {
          setError('Error: Error inesperado.');
        }
      }
```

src/services/shopify-client.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface ConnectedShops {
+  shopName: string;
+  synchronization_date: Date | null;
+  list: SubscriberList;
+}
+
+export type ShopifyErrorResult = { cannotConnectToAPI: true };
+export interface ShopifyClient {
+  getShopifyData(): Promise<ConnectedShops[] | EmptyResult<ShopifyErrorResult>>;
```
andresmoschini: Creo que el tipo de resultado debería ser `Result<ConnectedShops[], ShopifyErrorResult>`

```suggestion
  getShopifyData(): Promise<Result<ConnectedShops[], ShopifyErrorResult>>;
```

cbernat: This should behave in the same way right?

andresmoschini: So so, it will enforce you to wrap the result and include the `success: true` field, like here:

https://github.com/FromDoppler/doppler-webapp/blob/7b5399b8ec18d8fbbf94032ab75e2f3bc7428af9/src/services/doppler-legacy-client.ts#L321-L328

src/components/Integrations/Shopify/Shopify.js
andresmoschini: Ah, hay otra anidación más. En general, este tipo de lógica se pueden leer mucho mejor cuando hay un solo nivel de anidación.

En este caso yo sugeriría la siguiente estructura:

```js
      const result = await shopifyClient.getShopifyData();
      if (result.success) { 
        // ahora mismo no funciona porque el tipo no es Result
        // . . .
      } else if (!result.value.length) { 
        // el campo value tambiés es parte del tipo Result
        // Tal vez incluso se pueda definir un tipo diferente de resultado para
        // cuando no hay items
        // . . .
      } else if (result.expectedError && result.expectedError.cannotConnectToAPI) {
        // . . .
      } else {
        // . . .
      }
```

cbernat: I get your point.

src/services/shopify-client.ts
cbernat: I've tried but for some reason it gives error. 
image

andresmoschini: It is not empty any more hehe

andresmoschini: please, let me know if it is not clear, we can talk about it.

cbernat: Yeah I've noticed it. I'm using `Result` now locally.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -0,0 +1,54 @@
...
+      <Helmet title={'Doppler | Shopify'} />
+      <section style={{ width: '100%', padding: '60px 30px' }}>
+        Panel de control | Integraciones y preferencias avanzadas
+        <h1>Panel de Control - Shopify</h1>
+        {isLoading ? (
+          <Loading />
+        ) : error ? (
+          error
+        ) : isConnected ? (
+          shops.map((shop) => 'Shopify conectado a tienda: ' + shop.shopName)
```
andresmoschini: If I am not wrong, it returns an array, what do you think about one of these options?

```suggestion
          `Shopify conectado a tienda: ${shops[0].shopName}`
```

Or:

```suggestion
          <p>Conectado a:</p>
          <ul>
            {shops.map((shop) => (<li>{shop.shopName}</li>)}
          </ul>
```

cbernat: I like the second option best.

src/components/Header/Header.js
```diff
@@ -4,14 +4,15 @@ import HeaderMessages from './HeaderMessages/HeaderMessages';
 import HeaderUserMenu from './HeaderUserMenu/HeaderUserMenu';
 import { FormattedMessage } from 'react-intl';
 
-const Header = ({ userData: { user, nav, alert } }) => {
+const Header = ({ userData: { user, nav, alert }, location }) => {
+  const inactiveSection = location.pathname.match(/^\/integrations\/*/) !== null;
```
andresmoschini: Maybe using deconstruction, the dependency will be more explicit:

```js
const Header = ({ userData: { user, nav, alert }, location: { pathname }}) => {
  const inactiveSection = pathname.match(/^\/integrations\/*/) !== null; 
```

src/components/form-helpers/captcha-utils.tsx
```diff
@@ -6,7 +6,7 @@ type CaptchaUtils = Readonly<
   [
     (props: any) => JSX.Element,
     () => Promise<CaptchaVerificationResult>,
-    React.RefObject<ReCAPTCHA>,
+    React.RefObject<ReCAPTCHA>
```
andresmoschini: Please revert this change, it seems that it is being generated by a wrong _Prettier_ version.

Try this:

```bash
$ rm -rf  node_modules
$ yarn
$ yarn prettier
```

src/services/shopify-client.doubles.ts
```diff
@@ -20,37 +20,10 @@ const oneShop = [
...
-
 export class HardcodedShopifyClient implements ShopifyClient {
   public async getShopifyData(): Promise<Result<ConnectedShop[], ShopifyErrorResult>> {
     console.log('getShopifyData');
     await timeout(1500);
-    //return {success: false, message: 'Some random error'};
-    //return {success: false, expectedError: {cannotConnectToAPI: true}};
-    //return { success: true, value: twoShops };
     return { success: true, value: oneShop };
-    //return { success: true, value:[]};
```
andresmoschini: What do you think about keeping these commented lines? I think that they could help us to test manually.

andresmoschini:
Muy bueno! creo que ya se le podría ir quitando el WIP y mergeando, ¿que opinás?

cbernat:
> Muy bueno! creo que ya se le podría ir quitando el WIP y mergeando, ¿que opinás?

Tiene como veinte millones de commits, lo emprolijo en otro PR mejor :)

andresmoschini:
Oh, remember to also take into account Travis errors (We are still not paying attention to them in our Jenkins task):

image

image

cbernat:
Test this please

src/doppler-types.ts
```diff
@@ -0,0 +1,10 @@
+// TODO: move it a common place if it will be reused
```
andresmoschini: Esta línea creo que ya no es necesaria.

andresmoschini:
:shipit:

GusBaamonde:
This is weird, this later I check it

shopy

src/components/Header/Header.js
```diff
@@ -4,14 +4,19 @@ import HeaderMessages from './HeaderMessages/HeaderMessages';
...
     <div>
       {alert ? <HeaderMessages alert={alert} user={user} /> : null}
       {/* //TODO: Refactor backend to send proper active values. Class 'header-is-active' must be removed */}
       <header
         className={
-          user.clientManager ? 'header-main header-open dp-header--cm' : 'header-main header-open'
+          inactiveSection
+            ? 'header-main'
+            : user.clientManager
+            ? 'header-main header-open dp-header--cm'
```
fcoronelmakingsense: cant do this in a better way? because repeat 'header main' always

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -0,0 +1,59 @@
...
+      if (result.success && result.value.length) {
+        setIsConnected(true);
+        setShops(result.value);
+      } else if (
+        !result.success &&
+        result.expectedError &&
+        result.expectedError.cannotConnectToAPI
+      ) {
+        setError('Error: No hemos podido conectar con la Api de Shopify, vuelve a intentar luego.');
+      } else if (!result.success) {
```
fcoronelmakingsense: maybe its better ask first this and then dont need to repeat ask for `!result.success`

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -0,0 +1,59 @@
...
+      setIsLoading(false);
+    };
+    getData();
+  }, [shopifyClient]);
+
+  return (
+    <>
+      <Helmet title={'Doppler | Shopify'} />
+      <section style={{ width: '100%', padding: '60px 30px' }}>
+        Panel de control | Integraciones y preferencias avanzadas
```
fcoronelmakingsense: dont use i18n texts?

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -0,0 +1,59 @@
...
+      }
+      setIsLoading(false);
+    };
+    getData();
+  }, [shopifyClient]);
+
+  return (
+    <>
+      <Helmet title={'Doppler | Shopify'} />
+      <section style={{ width: '100%', padding: '60px 30px' }}>
```
fcoronelmakingsense: its better use css class, what do you think?

GusBaamonde: I think it's great, but it's something to fix in another PR. Then we see it.  ;)

src/services/shopify-client.doubles.ts
```diff
@@ -0,0 +1,56 @@
...
+    list: {
+      name: 'MyList',
+      id: 1251,
+      amountSubscribers: 2,
+      state: SubscriberListState.ready,
+    },
+  },
+];
+
+/*const twoShops = [
```
fcoronelmakingsense: need this code?

src/services/shopify-client.doubles.ts
```diff
@@ -0,0 +1,56 @@
...
+      state: SubscriberListState.ready,
+    },
+  },
+];*/
+
+export class HardcodedShopifyClient implements ShopifyClient {
+  public async getShopifyData(): Promise<Result<ConnectedShop[], ShopifyErrorResult>> {
+    console.log('getShopifyData');
+    await timeout(1500);
+    //return {success: false, message: 'Some random error'};
```
fcoronelmakingsense: need all the comments? what do you think @andresmoschini

src/components/Integrations/Shopify/Shopify.js
cbernat: Not for now. This was meant to be very simple.

src/components/Integrations/Shopify/Shopify.js
cbernat: This is temporal, next I will be adding styles already merged into ui library, check https://cdn.fromdoppler.com/doppler-ui-library/latest/integrations.html

src/services/shopify-client.doubles.ts
cbernat: Yes, for quick testing purposes.

src/doppler-types.ts
cbernat: > This is weird, this later I check it
> 
> shopy



> This is weird, this later I check it
> 
> shopy

I don't see how to reproduce this, can you tell me please?

cbernat:
> This is weird, this later I check it
> 
> shopy

I don't get how you reproduce this, would you mind telling me how?

src/services/shopify-client.doubles.ts
cbernat: I've added this again, at @andresmoschini request. For quick testing purposes.

GusBaamonde:
> No entiendo cómo reproduces esto, ¿te importaría decirme cómo?
Yes! You enter https://cdn.fromdoppler.com/doppler-webapp/build883/#/integrations/shopify
and unfold the user menu

src/components/Integrations/Shopify/Shopify.js
andresmoschini: The alternative are _nested ifs_, and they tend to be less clear. Intendation is our enemy! :P

src/services/shopify-client.doubles.ts
andresmoschini: These comments are really useful for testing weird scenarios in the local environment.

src/components/Integrations/Shopify/Shopify.js
fcoronelmakingsense: can add html comment? to dont forget change

src/services/shopify-client.doubles.ts
fcoronelmakingsense: thanks for answer!

src/components/Integrations/Shopify/Shopify.js
fcoronelmakingsense: maybe can use something like that too.
```js
if(result.expectedError &&
        result.expectedError.cannotConnectToAPI){}
    else if(!result.sucess){}
    else if(result.value.length){}
```js

Comment (cbernat) [reply src/components/Integrations/Shopify/Shopify.js]: I will try this out. Thanks!

Comment (cbernat) [reply src/components/Integrations/Shopify/Shopify.js]: I went out perfect! Thanks!

File: src/components/Integrations/Shopify/Shopify.js (hunk 23 above)
Comment (cbernat) [reply src/components/Integrations/Shopify/Shopify.js]: @fcoronelmakingsense I can't seem to add html commented code in between. But I will remember, my next pr will be about applying styles in the library.

File: src/components/Integrations/Shopify/Shopify.js [hunk 26]
```diff
@@ -0,0 +1,55 @@
...
+    const getData = async () => {
+      const result = await shopifyClient.getShopifyData();
+      if (result.expectedError && result.expectedError.cannotConnectToAPI) {
+        setError('Error: No hemos podido conectar con la Api de Shopify, vuelve a intentar luego.');
+      } else if (!result.success) {
+        setError('Error: Error inesperado.');
+      } else if (result.value.length) {
+        setIsConnected(true);
+        setShops(result.value);
+      }
```
andresmoschini: Creo que falta esto:

```suggestion
      } else {
        setIsConnected(false);
        setShops(result.value);
      }
```

cbernat: That's not quite right, if I remove the condition from this last if in the case that is disconnected, it fails one test. 
```
return { success: true, value:[]};
```
If there is no correct data change through out the if, by default is disconnected.

cbernat: I don't need an extra else, because is disconnected by default. Perhaps it's confusing...

andresmoschini: In general, having deterministic functions that do not depend on external context (original value) is a good thing.

Take into account that this function will be executed whenever React wants. So, maybe this function could be executed twice. 

Suppose this scenario: the current user is changed, I do not know, a login in another page, or something happens with current shopify integration, before there was a shop, and now there is not a shop, before was connected, and now it is not.

You are missing this scenario here.

PR #257, opened 2019-06-19
fix: add .dockerignore file

andresmoschini:
It seems that we were copying all the files (including `node_modules`) to the container that generates the build, so, we were including unnecessary and probably wrong files.

Could you verify and let me know if do you think that other files could be ignored?

(I will merge it soon, but we can improve it later anyway)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+node_modules
+build
+.vscode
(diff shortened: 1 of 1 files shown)


cbernat:
Great! ❤️

PR #271, opened 2019-06-25
fix(shopify): add al possible path in getData function

cbernat:
In order to have useEffect function calling itself several times, and have the correct state of view. 
Related to https://github.com/FromDoppler/doppler-webapp/pull/235#discussion_r296875387


andresmoschini:
👍

PR #273, opened 2019-06-25
Clear class that no use in the UI library

IgnacioRodrigues:
In this pull request, I worked to eliminate classes that are no longer useful, this is because I am working on cleaning the UI library so that the code is as clean and neat as possible. In this case, only the call to the class "link - title" was deleted

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -158,10 +158,7 @@ const Login = ({ intl, location, dependencies: { dopplerLegacyClient, sessionMan
         <h5>{_('login.enter_doppler')}</h5>
         <p className="content-subtitle">{_('login.enter_doppler_sub')}</p>
         <p className="content-subtitle">
-          {_('login.you_want_create_account')}{' '}
-          <Link to="/signup" className="link--title">
-            {_('login.signup')}
-          </Link>
+          {_('login.you_want_create_account')} <Link to="/signup">{_('login.signup')}</Link>
         </p>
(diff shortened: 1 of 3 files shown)


src/components/Login/Login.js
```diff
@@ -159,7 +159,7 @@ const Login = ({ intl, location, dependencies: { dopplerLegacyClient, sessionMan
         <p className="content-subtitle">{_('login.enter_doppler_sub')}</p>
         <p className="content-subtitle">
           {_('login.you_want_create_account')}{' '}
-          <Link to="/signup" className="link--title">
+          <Link to="/signup">
```
andresmoschini: Me parece que esto no es un breaking change.

andresmoschini: Menos para la webapp, donde el cliente es un usuario humano (en general más tolerante a los cambios que una máquina)

IgnacioRodrigues: Que le pondrías vos? Que es un refactor?

cbernat: If we are eliminanting redundant classes we are not changing nothing substantial, I woudl say this is a chore.

andresmoschini:
👍

andresmoschini:
:shipit: 

But, you will need to prettify it before:

image

IgnacioRodrigues:
> But, you will need to prettify it before:

Yes! Now run prettier 🐱‍💻

PR #281, opened 2019-06-27
Change the default name in manifest for Doppler

IgnacioRodrigues:
Remplace name in manifest.json for Doppler.
ezgif-5-3625b76b7e71


andresmoschini:
Good catch (again)!

Thanks @IgnacioRodrigues, :shipit:.

PR #282, opened 2019-07-01
Dbr 341 header fixes

cbernat:
- Add fixes for client manager when inactive section, so logo can be seen in shopify

image

- Also add client manager user json, for quick testing

Related to https://github.com/FromDoppler/Doppler-UI-System/pull/129

https://cdn.fromdoppler.com/doppler-webapp/build942/#/integrations/shopify

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -17,11 +17,8 @@ const Header = ({
       <header
         className={
           'header-main' +
-          (!inactiveSection
-            ? user.clientManager
-              ? ' header-open dp-header--cm'
-              : ' header-open'
-            : '')
+          (inactiveSection ? ' ' : ' header-open') +
+          (user.clientManager ? ' dp-header--cm' : ' ')
         }
       >
         {user.clientManager ? (
--- a/public/index.html
+++ b/public/index.html
@@ -12,7 +12,7 @@
     -->
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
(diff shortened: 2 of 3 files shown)


src/components/Header/Header.js
```diff
@@ -17,11 +17,8 @@ const Header = ({
       <header
         className={
           'header-main' +
-          (!inactiveSection
-            ? user.clientManager
-              ? ' header-open dp-header--cm'
-              : ' header-open'
-            : '')
+          (inactiveSection ? ' ' : ' header-open') +
+          (user.clientManager ? ' dp-header--cm' : ' ')
```
andresmoschini: 👍

fcoronelmakingsense:
its need the other .json?

cbernat:
> its need the other .json?

Yes for testing, whenever you need to test a cm user, just copy and paste into headerData.json, we could add more files for specific and rare cases of users.

src/dopplerUserJson/cmUser.json
```diff
@@ -0,0 +1,194 @@
+{
```
andresmoschini: What do you think about renaming it in a more descriptive way?

For example: `/src/headerData-for-client-manager-user.json `

cbernat: Yes it sounds more understandable.

cbernat: I've also remembered YAGNI principle.

andresmoschini: Just curiosity, how does YAGNI apply here?

cbernat: The folder, may not be needed.

PR #283, opened 2019-07-01
add: unique visits in reports

fcoronelmakingsense:
- Change datahub request to obtain unique visits. related with [CDH-245](http://jira.makingsense.com/browse/CDH-245)
- Remove total visits ranking because now dont use now

image
image

### DIFF
--- a/src/components/Reports/ReportsPageRanking/ReportsPageRanking.js
+++ b/src/components/Reports/ReportsPageRanking/ReportsPageRanking.js
@@ -95,7 +95,7 @@ class ReportsPageRanking extends React.Component {
                     {item.name}
                   </a>
                   <p className="text-ranking">
-                    <strong>{item.totalVisits}</strong>{' '}
+                    <strong>{item.totalVisitors}</strong>{' '}
                     <FormattedMessage id="reports_pageranking.total_visits" />
                   </p>
                 </div>
--- a/src/services/datahub-client.ts
+++ b/src/services/datahub-client.ts
@@ -19,7 +19,7 @@ export interface DatahubClient {
   getPagesRankingByPeriod(query: {
     domainName: number;
     dateFrom: Date;
-  }): Promise<{ name: string; totalVisits: number; url: string }[]>;
+  }): Promise<{ name: string; totalVisitors: number; url: string }[]>;
 }
 
 export class HttpDatahubClient implements DatahubClient {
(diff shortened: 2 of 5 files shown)


andresmoschini:
@fcoronelmakingsense please include in this PR all content changes, because the information now is different, so the related content should be also different.

andresmoschini:
Also, the names of the methods, for example: If I am not wrong, `getVisitsByPeriod` should be `getVisitorssByPeriod` now, right?

src/services/datahub-client.ts
```diff
@@ -106,6 +106,7 @@ export class HttpDatahubClient implements DatahubClient {
       `domains/${domainName}/events/summarized-by-page`,
       {
         startDate: dateFrom.toISOString(),
+        sortBy: 'visitors',
```
andresmoschini: @leoslopez, @fcoronelmakingsense,  is it enough to ask for the _visitors_? is the result still in the count property?

image

image

andresmoschini: image

I think that _totalVisits_ name is wrong now, because this value is not the _total of visits_ anymore, now it is the _count of visitors_ or _unique visits_, Am I missing something?

fcoronelmakingsense: sorry my bad. Its visitorsQuantity

fcoronelmakingsense: i fixed now

fcoronelmakingsense: fixed

fcoronelmakingsense:
> @fcoronelmakingsense please include in this PR all content changes, because the information now is different, so the related content should be also different.

i ask to @mfantini for the content

andresmoschini:
> > @fcoronelmakingsense please include in this PR all content changes, because the information now is different, so the related content should be also different.
> 
> i ask to @mfantini for the content

Si, está bien, pero mientras tanto, en los lugares donde dice "visitas" podés poner "visitantes" o "visitas únicas", etc. 

Osea, el contenido antes indicaba visitas totales y ahora ya no, más allá de que contenido nos pasen, deberíamos tratar de usar el contenido correcto, lo mismo para nuestros nombres de variables, métodos, archivos, etc.

src/components/Reports/ReportsPageRanking/ReportsPageRanking.js
```diff
@@ -95,7 +95,7 @@ class ReportsPageRanking extends React.Component {
                     {item.name}
                   </a>
                   <p className="text-ranking">
-                    <strong>{item.totalVisits}</strong>{' '}
+                    <strong>{item.totalVisitors}</strong>{' '}
                     <FormattedMessage id="reports_pageranking.total_visits" />
```
andresmoschini: What do you think about:

```suggestion
                    <FormattedMessage id="reports_pageranking.visitors_count" />
```

fcoronelmakingsense: yes! when i think change when have the content, but can change now

PR #284, opened 2019-07-02
feat: add datahub track

cbernat:
Add data hub track for current page.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -46,7 +46,7 @@ history.listen((location) => {
   const locationPage = location.hash && location.hash[0] === '#' && location.hash.slice(1);
   ReactGA.set({ page: locationPage });
   ReactGA.pageview(locationPage);
-  // TODO: Also notify navigation to DataHub (DataHub does not have this feature yet)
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -46,7 +46,7 @@ history.listen((location) => {
   const locationPage = location.hash && location.hash[0] === '#' && location.hash.slice(1);
   ReactGA.set({ page: locationPage });
   ReactGA.pageview(locationPage);
-  // TODO: Also notify navigation to DataHub (DataHub does not have this feature yet)
+  _dha.track({ navigatedPage: locationPage }); // eslint-disable-line
```
andresmoschini: What is the disabled rule?

andresmoschini: Maybe you could do something like this:

```js
   window._dha && window._dha.track({ navigatedPage: locationPage });
```

cbernat: To prevent eslint from throwing a warning, because `_dha` is a global variable not defined in `index.js`

cbernat:
I will merge this, I've tested manually and rechecked with Leo from DataHub.

PR #293, opened 2019-07-10
feat: add shopify logo

cbernat:
Added shopify logo, and updated library to fix header. 

image

http://cdn.fromdoppler.com/doppler-webapp/build984/#/integrations/shopify


src/components/Integrations/Shopify/Shopify.js
```diff
@@ -2,6 +2,8 @@ import React, { useState, useEffect } from 'react';
 import { Helmet } from 'react-helmet';
 import Loading from '../../Loading/Loading';
 import { InjectAppServices } from '../../../services/pure-di';
+import logo from './logo.svg';
+import './Shopify.css';
```
andresmoschini: What do you think about using lowercase by default for the file names?

andresmoschini: (if you change it, remember to also change the file name)

src/components/Integrations/Shopify/Shopify.css
```diff
@@ -0,0 +1,8 @@
+img {
+    width: 10%;
```
andresmoschini: Why is it a percentaje? is it fine apply this style for all `img`s?

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -73,7 +75,7 @@ const Shopify = ({ dependencies: { shopifyClient } }) => {
                 <div className="block dp-integration__status">
                   <div className="status__info">
                     <div>
-                      {/*<img src="img/integration__logo.png" alt="Shopify logo"/> this should be added as a class, and will be added later*/}
+                      <img src={logo} alt="Shopify logo" />
```
IgnacioRodrigues: What do you think about using class for show logo the Shopify or integrations? As Doppler logo? 
`<h1 class="logo-doppler-new"><a target="_blank" href="https://www.fromdoppler.com/?utm_source=app&amp;utm_medium=landing&amp;utm_campaign=login" rel="noopener noreferrer">Doppler</a></h1>`

src/components/Integrations/Shopify/Shopify.js
cbernat: For all? We were using uppercase for components...

src/components/Integrations/Shopify/Shopify.css
cbernat: All images elements inside shopify component only. But I better add a class..

src/components/Integrations/Shopify/Shopify.js
andresmoschini: Oh, I did not understand the logic, it has the same name than the component. I understand it now. Thanks.

src/components/Integrations/Shopify/Shopify.css
andresmoschini: 10% regarding what?

cbernat: The idea of having css inside a component is that it applies only to this component.

cbernat: The outer parent container.
image mmm doesn't seem to be a good idea. I will change it.

PR #299, opened 2019-07-11
fix: add navigation tracking on first load

cbernat:
For Google Analytics and Doppler Datahub, add track for first load. To avoid tracking of root --> /.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -41,12 +41,18 @@ ReactGA.initialize('UA-532159-1');
 
 const history = createBrowserHistory();
 
-// Get the current location
-history.listen((location) => {
+const trackNavigation = (location) => {
   const locationPage = location.hash && location.hash[0] === '#' && location.hash.slice(1);
   ReactGA.set({ page: locationPage });
   ReactGA.pageview(locationPage);
   window._dha && window._dha.track({ navigatedPage: locationPage });
+};
+
+trackNavigation(window.location);
+
+// Get the current location
+history.listen((location) => {
+  trackNavigation(location);
 });
 
 ReactDOM.render(


src/index.js
```diff
@@ -41,12 +41,18 @@ ReactGA.initialize('UA-532159-1');
 
 const history = createBrowserHistory();
 
-// Get the current location
-history.listen((location) => {
+const trackNavigation = (location) => {
   const locationPage = location.hash && location.hash[0] === '#' && location.hash.slice(1);
   ReactGA.set({ page: locationPage });
   ReactGA.pageview(locationPage);
```
andresmoschini: I understand that removing _locationPage_ parameter from here is out of the scope of this PR.

src/index.js
```diff
@@ -41,12 +41,18 @@ ReactGA.initialize('UA-532159-1');
...
   window._dha && window._dha.track({ navigatedPage: locationPage });
```
andresmoschini: I understand that removing send right parameters to DataHub (we should split page, querystring, hash, etc) is out of the scope of this PR.

andresmoschini:
👍

src/index.js
cbernat: I don't get your point here. Why remove this `locationPage`?

cbernat: I'm obligued to add location page in pageview, by this library we are using now. I've tried with no luck. `ReactGA.pageview()`;

cbernat:
I'm merging this to proper testing of this issue in production.

src/index.js
andresmoschini: 👍

PR #307, opened 2019-07-16
Add styled components and apply in Shopify Logo

cbernat:
- Add styled components
- Use in shopify image logo

### DIFF
--- a/src/components/Integrations/Shopify/Shopify.css
+++ b/src/components/Integrations/Shopify/Shopify.css
@@ -1,8 +1,5 @@
-img.shopify-logo {
-    width: 80px;
-}
 @media only screen and (max-width: 600px) {
-    img.shopify-logo {
-        display: none;
-    }
+  img.shopify-logo {
+    display: none;
+  }
 }
--- a/src/components/Integrations/Shopify/Shopify.css
+++ b/src/components/Integrations/Shopify/Shopify.css
@@ -1,5 +1,0 @@
-@media only screen and (max-width: 600px) {
-  img.shopify-logo {
-    display: none;
-  }
-}
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -3,7 +3,6 @@ import { Helmet } from 'react-helmet';
 import Loading from '../../Loading/Loading';
 import { InjectAppServices } from '../../../services/pure-di';
 import logo from './logo.svg';
-import './Shopify.css';
 import { FormattedHTMLMessage, FormattedMessage, injectIntl } from 'react-intl';
 import styled from 'styled-components';
 
@@ -18,6 +17,8 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
   );
   const StyledShopifyLogo = styled(ShopifyLogo)`
     width: 80px;
+    @media only screen and (max-width: 600px) {
+      display: none;
     }
   `;
 
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -5,6 +5,7 @@ import { InjectAppServices } from '../../../services/pure-di';
 import logo from './logo.svg';
 import './Shopify.css';
 import { FormattedHTMLMessage, FormattedMessage, injectIntl } from 'react-intl';
+import styled from 'styled-components';
 
 const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
   const [shops, setShops] = useState([]);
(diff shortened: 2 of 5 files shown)


src/components/Integrations/Shopify/Shopify.js
```diff
@@ -5,13 +5,21 @@ import { InjectAppServices } from '../../../services/pure-di';
...
 import { FormattedHTMLMessage, FormattedMessage, injectIntl } from 'react-intl';
+import styled from 'styled-components';
 
 const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
   const [shops, setShops] = useState([]);
   const [isConnected, setIsConnected] = useState(null);
   const [error, setError] = useState(null);
   const [isLoading, setIsLoading] = useState(true);
   const _ = (id, values) => intl.formatMessage({ id: id }, values);
+  const ShopifyLogo = ({ className, svglogo }) => (
```
andresmoschini: Is `className` parameter required? how does it work?

andresmoschini:
Looks nice!

src/components/Integrations/Shopify/Shopify.css
```diff
@@ -1,8 +1,5 @@
...
 @media only screen and (max-width: 600px) {
-    img.shopify-logo {
-        display: none;
-    }
-}
\ No newline at end of file
+  img.shopify-logo {
+    display: none;
+  }
+}
```
andresmoschini: You will completely remove this file, right?

src/components/Integrations/Shopify/Shopify.js
cbernat: I've read it is neccesary here: https://www.styled-components.com/docs/basics#styling-any-component

src/components/Integrations/Shopify/Shopify.css
cbernat: No, because I need to add the media query first in the styled component. Then I will remove the file.

fcoronelmakingsense: i don't understand why use mediaquery 600px. we support mobile?

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -82,7 +90,7 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
                 <div className="block dp-integration__status">
                   <div className="status__info">
                     <div>
-                      <img className="shopify-logo" src={logo} alt="Shopify logo" />
+                      <StyledShopifyLogo svglogo={logo} />
```
fcoronelmakingsense: The idea is use the word "Styled" first when use styled components?

cbernat: Yes this way is suggested in Styled Components page. To keep a standard, because we have two components here: `ShopifyLogo` and `StyledShopifyLogo`

cbernat: It's the way I found, if you can think of a better one, please tell me.

src/components/Integrations/Shopify/Shopify.css
cbernat: If I reduce the size of the window, the icon is like this:
image
That's why I think would be better to hide it.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -3,15 +3,24 @@ import { Helmet } from 'react-helmet';
...
 import { FormattedHTMLMessage, FormattedMessage, injectIntl } from 'react-intl';
+import styled from 'styled-components';
 
 const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
   const [shops, setShops] = useState([]);
   const [isConnected, setIsConnected] = useState(null);
   const [error, setError] = useState(null);
   const [isLoading, setIsLoading] = useState(true);
   const _ = (id, values) => intl.formatMessage({ id: id }, values);
+  const ShopifyLogo = ({ className, svglogo }) => (
```
andresmoschini: Since this component is named _ShopifyLogo_, what do you think about hardcoding the src in place of accepting it by parameter?

```suggestion
  const ShopifyLogo = ({ className }) => (
```

BTW, the logo is already hardcoded 

```
<img className={className} src={logo} alt="Shopify logo" />
                                ^^^^ It saids logo (hardcoded) in place of svglogo (parameter) , I personally prefeer hardcoded one.
```

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -82,7 +91,7 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
                 <div className="block dp-integration__status">
                   <div className="status__info">
                     <div>
-                      <img className="shopify-logo" src={logo} alt="Shopify logo" />
+                      <StyledShopifyLogo svglogo={logo} />
```
```suggestion
                      <StyledShopifyLogo />
```

cbernat:
Test this please.

andresmoschini:
🚀

lucianaDilizio:
Nice! 😄 ❤️

PR #311, opened 2019-07-18
feat: allow custom tracking for datahub

cbernat:
- Add datahub script as suggested in https://github.com/MakingSense/Customer-Data-Hub/pull/368
- Added `url-parse` library
- Refactor parse into utils
- Add tests

Pending to do in another PR by @andresmoschini and @fcoronelmakingsense : 
- [ ] Add `url-parse` library in the rest of the app
- [ ] Add promotions testing for url change and parse

To replace queryString library in the future take in account this example:
`queryString.parse(props.location.search)` --> `urlParse(props.location.search, true)`

http://cdn.fromdoppler.com/doppler-webapp/build1053/#/integrations/shopify

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -41,11 +41,33 @@ ReactGA.initialize('UA-532159-1');
 
 const history = createBrowserHistory();
 
+const parseUrl = (partialUrl) => {
+  const hash =
+    partialUrl.indexOf('#') !== -1
+      ? partialUrl.substring(partialUrl.indexOf('#') + 1, partialUrl.length)
+      : '';
+  const search =
+    partialUrl.indexOf('?') !== -1
+      ? hash.length
+        ? partialUrl.substring(partialUrl.indexOf('?') + 1, partialUrl.indexOf('#'))
+        : partialUrl.substring(partialUrl.indexOf('?') + 1, partialUrl.length)
+      : '';
+  const page =
+    partialUrl.split('?').length >= 2
+      ? partialUrl.split('?')[0]
+      : partialUrl.split('#').length >= 2
+      ? partialUrl.split('#')[0]
+      : partialUrl;
+  return { hash, search, page };
+};
+
 const trackNavigation = (location) => {
   const locationPage = location.hash && location.hash[0] === '#' && location.hash.slice(1);
   ReactGA.set({ page: locationPage });
   ReactGA.pageview(locationPage);
-  window._dha && window._dha.track({ navigatedPage: locationPage });
+  const result = parseUrl(locationPage);
+  window._dha &&
+    window._dha.track({ navigatedPage: result.page, hash: result.hash, search: result.search });
 };
 
 trackNavigation(window.location);
--- a/src/index.js
+++ b/src/index.js
@@ -16,6 +16,7 @@ import 'polyfill-array-includes';
 import 'promise-polyfill/src/polyfill';
 import { availableLanguageOrDefault } from './i18n/utils';
 import { HardcodedShopifyClient } from './services/shopify-client.doubles';
+import urlParse from 'url-parse';
 
 polyfill();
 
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -41,11 +41,33 @@ ReactGA.initialize('UA-532159-1');
 
 const history = createBrowserHistory();
 
+const parseUrl = (partialUrl) => {
```
leoslopez: @cbernat, maybe you can use something like: [url-parse](https://www.npmjs.com/package/url-parse), what do you think?

andresmoschini: Good catch @leoslopez.

@cbernat we are already using [query-string library](https://www.npmjs.com/package/query-string) to parse URLs. For example https://github.com/FromDoppler/doppler-webapp/blob/a74e292ec583f733f650b7563677334d140929f5/src/components/Login/Login.js#L34

cbernat: @andresmoschini this library you've mention parses urls relaying in location base parsing, does not detect hash as part of a query string for example. 
 If you do ```queryString.parseUrl('https://foo.bar?foo=bar#hash'); ```
//=> {url: 'https://foo.bar', query: {foo: 'bar'}} --> ignoring hash
image

I think I will go with Leo's suggestion.

andresmoschini: We will chat about it: https://makingsense.slack.com/archives/CGJ72QFQX/p1563477192179800

src/index.js
```diff
@@ -45,7 +46,8 @@ const trackNavigation = (location) => {
   const locationPage = location.hash && location.hash[0] === '#' && location.hash.slice(1);
   ReactGA.set({ page: locationPage });
   ReactGA.pageview(locationPage);
-  window._dha && window._dha.track({ navigatedPage: locationPage });
+  const dataHubParams = getDataHubParams(locationPage);
+  window._dha && window._dha.track({ ...dataHubParams });
```
leoslopez: @cbernat, I am ingnorant in the use of  `...` (among other things :D ). But would not be enough call to `track` in this way: `_dha.track(dataHubParams)`. 
What extra value does the 3 points add?

cbernat: I agree, as we talked earlier, I will change this. Thanks for the insight Leo!

andresmoschini: good catch!

if I am not wrong, both these calls are aquivalent:

```js
f(obj);
f({...obj});
```

Well, in fact, there is a minor difference. But I think that it does not affects us:

Imagine this:

```js
function f(obj) {
  obj.anotherProp = 'added';
}

const t1 = { test: 't1' };
f(t1);
console.log(JSON.stringify(t1)); 
// { "test": "t1", "anotherProp": "added" }

const t2 = { test: 't2' };
f({ ...t2 });
console.log(JSON.stringify(t2)); 
// { "test": "t2" }
```

cbernat:
image

Leo verified this build. I will merge this shortly if there are no objections.

src/utils.test.js
```diff
@@ -0,0 +1,63 @@
...
+describe('utils', () => {
+  describe('getDataHubParams', () => {
+    it('should parse url with no params and no hash correctly', () => {
+      //Arrange
+      //Act
+      const dataHubParams = getDataHubParams('/login');
+      //Assert
+      expect(dataHubParams.navigatedPage).toBe('/login');
+      expect(dataHubParams.search).toBe('');
+      expect(dataHubParams.hash).toBe('');
```
andresmoschini: Is it fine? or it should be null? @leoslopez what do you think?

yarn.lock
```diff
@@ -12857,7 +12857,7 @@ url-parse-lax@^1.0.0:
   dependencies:
     prepend-http "^1.0.1"
 
-url-parse@^1.4.3:
+url-parse@^1.4.3, url-parse@^1.4.7:
```
andresmoschini: Oh, we already had this dependency, interesting...

src/utils.test.js
leoslopez: It would be the same for the Datahub [interpretation](https://github.com/MakingSense/Customer-Data-Hub/blob/develop/public/dhapi.js#L117-L119): 
image

It means that, in case that the values are missing (null or empty too) in options, the values will be obtained from url.

But this makes me doubt about the proper interpretation of Datahub.
Maybe Datahub should do:
```
if (enableCustomTracking) {
  // consider values received
} else {
  // consider values from window.location
}
```
What do you think?

andresmoschini: Oh, you are right, good catch @leoslopez !

I like the change you are proposing.

yarn.lock
cbernat: Isn't yarn lock supposed to be automatically generated?

andresmoschini: Yes, yes, I wanted to say that we were already depending on _url-parse_ (in an indirect way), it was only a comment, nothing change.

PR #314, opened 2019-07-22
Add shopify http client 

cbernat:
- Add HttpClient for shopify, calling only shopify qa for now.
- The rest of the configs were added but are not linked. 
- Add error handling so cors issue should not be a problem.

image


src/services/shopify-client.ts
```diff
@@ -21,3 +24,45 @@ export type ShopifyErrorResult = { cannotConnectToAPI: true };
...
+    const { jwtToken } = this.getShopifyConnectionData();
+    const response = await this.axios.request<Result<ConnectedShop[], ShopifyErrorResult>>({
+      method: 'GET',
+      url: `/me/shops`,
+      headers: { Authorization: `token ${jwtToken}` },
+    });
+    return response.data;
+  }
+
+  // GET https://sfy.fromdoppler.com:4433/me/shops
```
andresmoschini: What is this comment?

src/services/shopify-client.ts
```diff
@@ -21,3 +24,45 @@ export type ShopifyErrorResult = { cannotConnectToAPI: true };
...
+    const connectionData = this.connectionDataRef.current;
+    if (!connectionData || connectionData.status !== 'authenticated' || !connectionData.jwtToken) {
+      throw new Error('Shopify connection data is not available');
+    }
+    return connectionData;
+  }
+
+  public async getShopifyData(): Promise<Result<ConnectedShop[], ShopifyErrorResult>> {
+    const { jwtToken } = this.getShopifyConnectionData();
+    const response = await this.axios.request<Result<ConnectedShop[], ShopifyErrorResult>>({
```
andresmoschini: I think that the type is wrong, is the response is successful you have to map it to a `ConnectedShop[]`, if it is not (or if there is an exception), you have to map it to a `ShopifyErrorResult`

src/services/shopify-client.ts
```diff
@@ -21,3 +24,45 @@ export type ShopifyErrorResult = { cannotConnectToAPI: true };
...
+  private getShopifyConnectionData(): ShopifyConnectionData {
+    const connectionData = this.connectionDataRef.current;
+    if (!connectionData || connectionData.status !== 'authenticated' || !connectionData.jwtToken) {
+      throw new Error('Shopify connection data is not available');
+    }
+    return connectionData;
+  }
+
+  public async getShopifyData(): Promise<Result<ConnectedShop[], ShopifyErrorResult>> {
+    const { jwtToken } = this.getShopifyConnectionData();
```
andresmoschini: Should the potential exception here be wrapped in a `ShopifyErrorResult`?

src/services/pure-di.tsx
```diff
@@ -118,7 +117,15 @@ export class AppCompositionRoot implements AppServices {
   }
 
   get shopifyClient() {
-    return this.singleton('shopifyClient', () => new HardcodedShopifyClient());
+    return this.singleton(
+      'shopifyClient',
+      () =>
+        new HttpShopifyClient({
+          axiosStatic: this.axiosStatic,
+          baseUrl: 'https://sfy.fromdoppler.com:4433',
```
andresmoschini: Remember to move it to the configuration file.

Also, I have already merge it to production, you can remove the port.

src/services/shopify-client.ts
cbernat: Perhaps :) This as in Datahub gets the jwtToken and that's it.

andresmoschini: Oh, _DataHub Client_ is not exception free, I think that you should use _Doppler Client_ as example: https://github.com/FromDoppler/doppler-webapp/blob/57af926f370aeee1c7bb2ad83bce813078ab6179/src/services/doppler-legacy-client.ts#L369-L373

src/services/shopify-client.ts
cbernat: Added happy path. Pending error.

src/services/pure-di.tsx
cbernat: I've added configurations but prefer to leave it hardcoded for testing purposes.

PR #320, opened 2019-07-26
feat(shopify): add table with list data and state

cbernat:
- Add table with list data:

image

- Add synchonizing state

image

- Updated UI library version

### DIFF
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -47,7 +47,7 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
       <thead>
         <tr>
           <th>{_('shopify.table_list')} </th>
-          <th> {_('shopify.table_subscribers_count')}</th>
+          <th> {_('shopify.table_shopify_customers_count')}</th>
         </tr>
       </thead>
       <tbody>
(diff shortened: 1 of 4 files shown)


src/i18n/en.js
```diff
@@ -185,7 +186,7 @@ export default {
     list_subtitle: `You can synchronize the data manually whenever you want.`,
     list_title: `Synchronized List`,
     table_list: `List Name`,
-    table_subscribers_count: `Subscribers`,
+    table_subscribers_count: `Shopify customers`,
```
andresmoschini: What do you think about also changing the key? maybe:

```suggestion
    table_shopify_customers_count: `Shopify customers`,
```

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -39,6 +41,33 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+          <tr>
+            <td>{list.name}</td>
+            <td>{list.amountSubscribers}</td>
+          </tr>
+        ) : (
+          <tr className="sync">
+            <td>{list.name}</td>
+            <td className="text-sync">
+              <span className="ms-icon icon-clock"></span>
+              Synchronizing...
```
andresmoschini: This is new, right? what do you think about adding a test?

src/i18n/en.js
cbernat: I thought that maybe in the future we will reuse this as subscribers, but ... YAGNI.

src/components/Integrations/Shopify/Shopify.js
cbernat: Also I've missed the resourced text here. Sure a test can help.

PR #326, opened 2019-07-30
Shopify: connect and wait for changes

cbernat:
Connect and wait for changes using hooks for polling.
- Move code fragment up
- Add hook useInterval 
- Add polling always 20 seconds, works for connect disconnect and any modifications in data.

When connected and it is not connected to a list, added new message.
image

So far the living code: http://cdn.fromdoppler.com/doppler-webapp/build1199/#/integrations/shopify


andresmoschini:
test this please

cbernat:
@andresmoschini This is working awesomly great! But since is pretty much behind master branch I will make another PR.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -68,24 +70,46 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+  }
+
   useEffect(() => {
     const getData = async () => {
       const result = await shopifyClient.getShopifyData();
       if (!result.success) {
         setError(<FormattedHTMLMessage id="validation_messages.error_unexpected_HTML" />);
       } else if (result.value.length) {
         setIsConnected(true);
+        setDoPoll(0);
```
fcoronelmakingsense: it's not better use and object and update component state only one time and don't four times?

cbernat: An object with what exactly? because I don't use all variables all the time...

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -61,35 +62,66 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+      <hr />
+    </>
+  );
+
+  function useInterval(callback, delay) {
+    const savedCallback = useRef();
+
+    useEffect(() => {
+      savedCallback.current = callback;
+    }, [callback]);
```
andresmoschini: What is the difference in the behavior of simply using this:

```diff
-    useEffect(() => {
-      savedCallback.current = callback;
-    }, [callback]);
+ savedCallback.current = callback;
```

The last assignment will only have effect when `savedCallback.current != callback`

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -61,35 +62,66 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+
+    useEffect(() => {
+      function tick() {
+        savedCallback.current();
+      }
+      if (delay !== null) {
+        let id = setInterval(tick, delay);
```
andresmoschini: Why did you use `let` in place of `const`? the `id` does not change, right?

src/components/Integrations/Shopify/Shopify.js
andresmoschini: What @fcoronelmakingsense said is reasonable: if a set of values change always together, it is useful to store the in only one state variable. 

Probably React has optimizations to avoid renders when the state will change soon, but, we could help and also the code could be cleaner.

@fcoronelmakingsense you have applied that approach in another component could you link it in this conversation?

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -61,35 +62,66 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
-        <p>
-          <FormattedHTMLMessage id="shopify.header_subtitle" />
-        </p>
-      </div>
-      <hr />
-    </>
+  useInterval(
+    () => {
+      if (isLoading) {
+        setDoPoll((doPoll) => doPoll + 1);
```
andresmoschini: What is the meaning of this counter?

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -61,35 +62,66 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+        <h2>{_('shopify.header_title')}</h2>
+        <p>
+          <FormattedHTMLMessage id="shopify.header_subtitle" />
+        </p>
+      </div>
+      <hr />
+    </>
+  );
+
+  function useInterval(callback, delay) {
```
andresmoschini: Could you move this to another place to allow sharing it?

src/components/Integrations/Shopify/Shopify.js
fcoronelmakingsense: i apply in `ReportsTrafficsources`, you can check this component @cbernat

src/components/Integrations/Shopify/Shopify.js
fcoronelmakingsense: i told her in `src/utils` looks better

src/components/Integrations/Shopify/Shopify.js
andresmoschini: Here you are:

https://github.com/FromDoppler/doppler-webapp/blob/58986eaeb82e3d02714f8994949c4f236a207e40/src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js#L35-L41

---

BTW, see more about this GitHub feature at:

* https://github.blog/2017-08-15-introducing-embedded-code-snippets/ 
* https://help.github.com/en/articles/creating-a-permanent-link-to-a-code-snippet

image

src/components/Integrations/Shopify/Shopify.js
cbernat: As I've understanded, the callback needs to be new each time, to allow accessing the current states variables, else they get stuck in the first variable.

andresmoschini: Mmmm, but the `useInterval` function is called each time.

When you have time, let us talk about it, I think that it would be better to understand for me.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -61,35 +62,66 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+
+    useEffect(() => {
+      savedCallback.current = callback;
+    }, [callback]);
+
+    useEffect(() => {
+      function tick() {
+        savedCallback.current();
+      }
+      if (delay !== null) {
```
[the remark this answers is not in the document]
cbernat: `delay` in null is set to stop de interval. That's why it is like this.

src/components/Integrations/Shopify/Shopify.js
cbernat: So it seems, I will change this...

src/components/Integrations/Shopify/Shopify.js
cbernat: It's a counter to trigger the effect of getShops data. See useEffect is dependant on this counter.

src/components/Integrations/Shopify/Shopify.js
andresmoschini: Maybe there is a reason, I only want to understand.

src/components/Integrations/Shopify/Shopify.js
andresmoschini: Mmm, it is difficult to understand. 

What do you think about doing the request in the callback?

cbernat: Seems a good idea, I'll see to it.

cbernat: It worked wonders! Thanks @andresmoschini !

src/components/Integrations/Shopify/Shopify.js
cbernat: Changed!

src/components/Integrations/Shopify/Shopify.js
cbernat: Fixed.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -99,6 +81,31 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
     </>
   );
 
+  const getData = useCallback(async () => {
```
andresmoschini: Very minor, but I am not confortable with the name, `getData` seems to be the name of a method to get information, in general that kind of methods returns a value and do not have side effects. 

In place of that, this method has side effects (updating the state) and do not return nothing.

What do you think about `updateData`, `update`, `pull` or `refresh`?

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -99,6 +81,31 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+    const result = await shopifyClient.getShopifyData();
+    if (!result.success) {
+      setError(<FormattedHTMLMessage id="validation_messages.error_unexpected_HTML" />);
+    } else if (result.value.length) {
+      setConnectionData({ shops: result.value, isConnected: true });
+      setIsConnecting(false);
+      setError(null);
```
andresmoschini: What do you think about convert _error_ and _is connecting_ states in properties like _shops_ and _is connected_?

src/components/Integrations/Shopify/Shopify.js
andresmoschini: Why is _useCallback_ necessary?

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -99,6 +81,31 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+    } else {
+      setConnectionData({ shops: result.value, isConnected: false });
+      setError(null);
+    }
+    if (!isConnecting) {
+      setIsLoading(false);
+    }
+  }, [isConnecting, shopifyClient]);
+
+  useEffect(() => {
```
andresmoschini: Why is _useEffect_ necessary?

src/utils.js
```diff
@@ -12,3 +13,21 @@ export function getDataHubParams(partialUrl) {
     search: parsedUrl.query,
   };
 }
+
+export function useInterval(callback, delay) {
+  const savedCallback = useRef();
+
+  useEffect(() => {
```
andresmoschini: I will try to understand why is this _useEffect_ necessary.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -99,6 +81,31 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
     </>
   );
 
+  const getData = useCallback(async () => {
+    const result = await shopifyClient.getShopifyData();
```
andresmoschini: I think that, before update the status, we need to determine if it has changed or not, in that way, we would avoid unnecessary and problematic rendering like this:

src/utils.js
```diff
@@ -12,3 +13,21 @@ export function getDataHubParams(partialUrl) {
     search: parsedUrl.query,
   };
 }
+
+export function useInterval(callback, delay) {
```
andresmoschini: Do you think that could be possible testing this hook?

Maybe there is some useful things here:
* https://github.com/testing-library/react-hooks-testing-library
* https://medium.com/@nitinpatel_20236/unit-testing-custom-react-hooks-caa86f58510

Besides the problem of how testing the hooks, we have another problem here, how to test when _setInterval_ is involved. Maybe this could help on this: https://jestjs.io/docs/en/timer-mocks

src/components/Integrations/Shopify/Shopify.js
andresmoschini: Thanks to _useEffect_, the content is executed after the first rendering, probably it is fine, but, why?

src/components/Integrations/Shopify/Shopify.js
cbernat: I've talk about this also with @fcoronelmakingsense this states do not update all at once. What difference this would make? This is to make code more readable and that's it?

src/components/Integrations/Shopify/Shopify.js
cbernat: If I use the function without useCallback I get this warning:
```  Line 84:  The 'getData' function makes the dependencies of useEffect Hook (at line 103) change on every render. To fix this, wrap the 'getData' definition into its own useCallback() Hook  react-hooks/exhaustive-deps ```
Since I use this in two places at the same time I've wrapped this up into useCallback hook.
https://github.com/facebook/react/issues/14920

File: src/components/Integrations/Shopify/Shopify.js (hunk 9 above)
Comment (cbernat) [reply src/components/Integrations/Shopify/Shopify.js]: As you've said I need this to execute the first time, else I have to wait for 20 seconds with a blank screen. O_o

File: src/utils.js (hunk 10 above)
Comment (cbernat) [reply src/utils.js]: Because you need some way of making the callback update, else you get the state frozen like the first time you've entered `useInverval` hook. The mechanism behind this, I'm not so sure.

File: src/components/Integrations/Shopify/Shopify.js (hunk 7 above)
Comment (andresmoschini) [reply src/components/Integrations/Shopify/Shopify.js]: Fixed since we removed the function.

Comment (andresmoschini) [reply src/components/Integrations/Shopify/Shopify.js]: `useCallback` was recommended in the warning, but it was not a good idea, because it memoizes the function. It is fixed since we do not need `getData` more than a time.

File: src/utils.js (hunk 10 above)
Comment (andresmoschini) [reply src/utils.js]: No, creo que la respuesta no va por ahí, porque en ese caso, esto también funcionaría:

```diff
- useEffect(() => {
   savedCallback.current = callback;
- }, [callback]);
```

El código dentro del `useEffect` se ejecuta después del código principal, pero aún no entiendo cual es la ventaja.

src/components/Integrations/Shopify/Shopify.js
andresmoschini: It is not happening anymore, we could check it again before the merge. Resolving the conversation...

src/components/Integrations/Shopify/Shopify.js
cbernat: We've talked this over, I will be uploading a fix shortly.

src/components/Integrations/Shopify/Shopify.js
cbernat: We fixed this changing the behaviour for `useInterval` hook adding the possibility to execute first time.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -1,22 +1,28 @@
...
-  const [isConnected, setIsConnected] = useState(null);
-  const [error, setError] = useState(null);
-  const [isLoading, setIsLoading] = useState(true);
+  const [shopifyState, setShopifyState] = useState({
+    shops: [],
+    isConnected: false,
+    isConnecting: false,
+    error: null,
+    isLoading: true,
+  });
```
andresmoschini: I think that this could be enough for default value:

```diff
  const [shopifyState, setShopifyState] = useState({
-    shops: [],
-    isConnected: false,
-    isConnecting: false,
-    error: null,
    isLoading: true,
  });
```

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -99,37 +86,61 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+          isLoading: false,
+        });
+      } else if (result.value.length) {
+        setShopifyState({
+          error: null,
+          shops: result.value,
+          isConnected: true,
+          isConnecting: false,
+          isLoading: false,
+        });
```
andresmoschini: I think that this should be enough:

```diff
        setShopifyState({
-          error: null,
          shops: result.value,
          isConnected: true,
-          isConnecting: false,
-          isLoading: false,
        });
```

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -99,37 +86,61 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+      } else {
+        setShopifyState({ error: null, shops: result.value, isConnected: false, isLoading: false });
```
```suggestion
        setShopifyState({ shops: [] });
```

cbernat: `IsLoading: false`, I need it to make the spinner dissapear, the same in the previous case.

andresmoschini: We have talked, the confusion is that @cbernat assumes that this works as class components _setState_ function, and I assume that it simply updates a value.

She will try and then let us know.

cbernat: @andresmoschini It works without merging state. So I'll keep the change done. Thanks!

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -99,37 +82,57 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+  useInterval({
+    runOnStart: true,
+    delay: 20000,
+    callback: async () => {
+      const result = await shopifyClient.getShopifyData();
+      if (!result.success) {
+        setShopifyState({
+          error: <FormattedHTMLMessage id="validation_messages.error_unexpected_HTML" />,
+        });
+      } else if (result.value.length) {
```
fcoronelmakingsense: dont need if `result.value` exist?

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -191,14 +194,26 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+                    onClick={() => {
+                      setShopifyState({ isConnecting: true });
+                    }}
+                  >
+                    {_('common.connect')}
+                  </a>
+                ) : (
+                  <button
+                    disabled
+                    className="dp-button button-medium primary-green button--loading"
```
fcoronelmakingsense: `primary-green` is a generic class? or is only for buttons? because the name means it's generic dont?

fcoronelmakingsense: why `button-medium` have only one `-` and `button--loading` have 2? both are modifiers maybe

src/utils.js
```diff
@@ -12,3 +13,24 @@ export function getDataHubParams(partialUrl) {
...
+
+export function useInterval({ callback, delay, runOnStart }) {
+  const savedCallback = useRef();
+
+  useEffect(() => {
+    savedCallback.current = callback;
+  }, [callback]);
+
+  useEffect(() => {
+    function tick() {
```
fcoronelmakingsense: it's right define the function inside? because all times state change the function define again maybe

src/utils.js
andresmoschini: Maybe they can help us: https://github.com/gaearon/overreacted.io/pull/346#issuecomment-518311802

src/utils.js
andresmoschini: That is the idea, postpouse the evaluation of _savedCallback.current_ to each tick.

This could be an alternative:

```js
const id = setInterval(() => savedCallback.current(), delay);
```

But this will not work fine:

```js
const id = setInterval(savedCallback.current, delay);
```

cbernat: That's the idea of it, I want to get the current state here.

src/components/Integrations/Shopify/Shopify.js
cbernat: I got this classes from ui library. I will check again to see if they are correct.

src/utils.js
cbernat: For more information check https://overreacted.io/making-setinterval-declarative-with-react-hooks/

src/components/Integrations/Shopify/Shopify.js
cbernat: If `result.success` the idea is that there should be a `value` in result. Maybe I am missing something?

src/components/Integrations/Shopify/Shopify.js
cbernat: They are as in ui library, please refer your concern there. http://cdn.fromdoppler.com/doppler-ui-library/latest/index.html

cbernat: I think that we agreed that modifiers will be added into classes using -- so yes it should be `button--medium` and --`button--primary` instead of `primary--green` right? @GusBaamonde @fcoronelmakingsense thoughts?

fcoronelmakingsense: i think you are right, but if change need to check in all places that use this classes

GusBaamonde: I agree, but I think that in a more detailed refactor for another pull request

cbernat: Si por eso decia, tal vez cambiemos el nombre de la clase a uno copado, manteniendo la clase vieja, hasta que cambiemos todo lo de webapp y ahi deprecamos la clase vieja de la libreria de ui.

src/services/shopify-client.doubles.ts
```diff
@@ -44,7 +44,16 @@ export class HardcodedShopifyClient implements ShopifyClient {
...
+    return { success: true, value: oneShop }; // Shopify connected to one shop
+    /*
+    //Test state change behaviour
+    if (new Date().getTime() % 2 !== 0) {
+      oneShop[0].list.amountSubscribers++;
+      return { success: true, value: oneShop };
+    } else {
+      return { success: true, value: [] }; 
+    }
+    */
```
andresmoschini: It is very arguable, so, please, take it as a simple suggestion.

What do you think about using _line comments_ (`// comment...`), in place of _block comments_ ( `/* comment ... */`).

They are more friendly with VS Code, you can use `CTRL+K CTRL+C` to comment and `CTRL+K CTRL+U` to uncomment.

cbernat: Fixed. `Ctrl + / ` in vs Code to toggle inline comments.

cbernat:
@andresmoschini I've added some tests for useInteval hook, can you think of a different scenario that would be interesting to test?

package.json
```diff
@@ -103,5 +103,9 @@
     "fail": [
       "@semantic-release/github"
     ]
+  },
+  "devDependencies": {
+    "@testing-library/react-hooks": "^1.1.0",
+    "react-test-renderer": "^16.8.6"
```
andresmoschini: We are not using dev dependencies because our build tool only includes the used code in production.

src/utils.test.js
```diff
@@ -60,4 +62,59 @@ describe('utils', () => {
...
+      renderHook(() =>
+        useInterval({
+          runOnStart: true,
+          delay: 100,
+          callback: () => {
+            counter++;
+          },
+        }),
+      );
+      jest.advanceTimersByTime(200);
```
andresmoschini: What do you think about also verify before moving in time?

```suggestion
      expect(counter).toBe(1);
      jest.advanceTimersByTime(200);
```

andresmoschini:
Nice tests! are we now missing an integration test of how Shopify component does the pulling? Maybe it is too much to ask...

cbernat:
> Nice tests! are we now missing an integration test of how Shopify component does the pulling? Maybe it is too much to ask...

I will try and think about this, if it gets too dificult perhaps I can make them in another PR, in order not to delay this feature longer.

src/utils.js
cbernat: Implemented.

package.json
cbernat: Fixed.

src/utils.test.js
cbernat: Fixed.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -51,11 +53,20 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+                <td>{list.name}</td>
+                <td className="text-sync">
+                  <span className="ms-icon icon-clock"></span>
+                  {_('common.synchronizing')}
+                </td>
+              </>
+            ) : (
+              <>
+                <td className="text-sync">{_('shopify.no_list_available')}</td>
+                <td className="text-sync"></td>
```
fcoronelmakingsense: empty column??

cbernat: yes, what do you suggest? leave one column with column span?

src/services/shopify-client.doubles.ts
```diff
@@ -15,6 +15,19 @@ const oneShop = [
   },
 ];
 
+// const oneShopNoList = [
```
fcoronelmakingsense: Maybe can add comment to know why is commented all code

src/utils.test.js
```diff
@@ -60,4 +62,60 @@ describe('utils', () => {
...
+          delay: 100,
+          callback: () => {
+            counter++;
+          },
+        }),
+      );
+
+      jest.runOnlyPendingTimers();
+      expect(counter).toBe(1);
+    });
```
fcoronelmakingsense: nice! maybe can add the comments to separate sections like `Arrange, Act, Assert`

src/components/Integrations/Shopify/Shopify.js
fcoronelmakingsense: you can delete

cbernat: As we've talked in slack I will remove this extra column here.

src/utils.test.js
cbernat: I will try, this test is very still very alien to me.

cbernat:
Test this please.

cbernat:
Test this please.

src/utils.test.js
cbernat: @fcoronelmakingsense solved!

cbernat:
@andresmoschini @fcoronelmakingsense @GusBaamonde  I will merge this in a moment if there are no objections.

src/services/shopify-client.ts
```diff
@@ -62,9 +63,11 @@ export class HttpShopifyClient implements ShopifyClient {
         name: response.dopplerListName,
         amountSubscribers: response.importedCustomersCount,
         state:
-          response.syncProcessInProgress === true
+          !!response.syncProcessInProgress && response.syncProcessInProgress !== 'false'
```
andresmoschini: Mmm, I am not sure about this, it seems that the server is returning not well-formed data.

Let me think it a little more before merging.

src/utils.test.js
```diff
@@ -60,4 +62,71 @@ describe('utils', () => {
...
+          runOnStart: true,
+          delay: 100,
+          callback: () => {
+            counter++;
+          },
+        }),
+      );
+      // Assert
+      expect(counter).toBe(1);
+      // Act
```
andresmoschini: I think that in this test is not an _act_ or maybe the `renderHook` is the _act_.

src/services/shopify-client.ts
andresmoschini: Oh, you have merged it just when I was commenting. 

Could you review if it will continue working fine when the backend result be ok?

src/utils.test.js
cbernat: Yes I had my doubts here.

src/services/shopify-client.ts
cbernat: Please let me know @andresmoschini when it's done.

cbernat: I've tested this and whether it's boolean or text it will work just fine.

andresmoschini: Great!

PR #328, opened 2019-07-31
Shopify fixes

cbernat:
- Fix disconnected warning as in confluence text
- Fix open external urls in a new tab

http://cdn.fromdoppler.com/doppler-webapp/build1122/#/integrations/shopify


andresmoschini:
👍

PR #331, opened 2019-07-31
feat(SignupConfirmation): add route for datahub tracking

fcoronelmakingsense:
- Add redirect to signup when access for url and signup isnt completed

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -13,6 +13,7 @@ import RedirectWithQuery from './components/RedirectWithQuery';
 import { Helmet } from 'react-helmet';
 import { availableLanguageOrNull } from './i18n/utils';
 import Shopify from './components/Integrations/Shopify/Shopify';
+import SignupConfirmation from './components/Signup/SignupConfirmation';
 
 class App extends Component {
   /**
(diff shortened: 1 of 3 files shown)


cbernat:
Very nice @fcoronelmakingsense !

andresmoschini:
But, I cannot see the redirect to `/signup/confirmation` after submit signup form. Am I missing something?

andresmoschini:
> But, I cannot see the redirect to `/signup/confirmation` after submit signup form. Am I missing something?

Also, having a test for that behavior would be nice.

cbernat:
> > But, I cannot see the redirect to `/signup/confirmation` after submit signup form. Am I missing something?
> 
> Also, having a test for that behavior would be nice.

Yes but how? I thought that too 😄

fcoronelmakingsense:
> But, I cannot see the redirect to `/signup/confirmation` after submit signup form. Am I missing something?

yes you are right, my bad

cbernat:
@fcoronelmakingsense The general idea of this task was to redirect to a new route after signup is ok, in order to allow tracking the access of this confirmation landing url (tracking is done by datahub and analytics that listen to changes in routing, this logic is in `index.js`)

fcoronelmakingsense:
test this please

cbernat:
test this please

andresmoschini:
Nice @fcoronelmakingsense ! it is weird but it is relatively simple and seems to work.

Having a test would be nice, but I understand that it is not so easy. So, if you want, merge as it is, but please keep thinking about how it (or a similar scenario) could be tested in the future.

fcoronelmakingsense:
> Nice @fcoronelmakingsense ! it is weird but it is relatively simple and seems to work.
> 
> Having a test would be nice, but I understand that it is not so easy. So, if you want, merge as it is, but please keep thinking about how it (or a similar scenario) could be tested in the future.

Yes, thanks I will consider it for possible scenarios in the future

PR #333, opened 2019-08-01
fix: change shopify button sizes

GusBaamonde:
- `button-medium` is for 95% of user actions.
- `button-big` is for promotional uses (as the case may be)
- `button-small` is for very specific cases (as the case may be)

sizes-buttons


cbernat:
👍 if you checked this with @santiagopaolucci I suppose it is correct.

GusBaamonde:
test this please

cbernat:
test this please.

santiagopaolucci:
> 👍 if you checked this with @santiagopaolucci I suppose it is correct.

It's ok :)

PR #334, opened 2019-08-02
fix(shopify): fix back button redirect when error

cbernat:
Added redirect to control panel for back button when there is an error.

http://cdn.fromdoppler.com/doppler-webapp/build1147/#/integrations/shopify

### DIFF
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -61,6 +61,13 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
       </tbody>
     </table>
   );
+
+  const backButton = (
+    <a href={_('common.control_panel_url')} className="dp-button button-medium primary-grey">
+      {_('common.back')}
+    </a>
+  );
+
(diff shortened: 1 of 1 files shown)


andresmoschini:
👍

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -118,9 +118,12 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
               </div>
               <footer className="dp-integration__actions">
-                <button type="button" className="dp-button button-medium primary-grey">
+                <a
+                  href={_('common.control_panel_url')}
+                  className="dp-button button-medium primary-grey"
+                >
                   {_('common.back')}
-                </button>
+                </a>
```
andresmoschini: Is it not possible to avoid coding duplication for this button?

cbernat: I can add a fragment for this, yes 😄

cbernat: Fixed

PR #335, opened 2019-08-02
fix: footer position

GusBaamonde:
related to: https://github.com/FromDoppler/Doppler-UI-System/pull/155

footer--position

Let me know what you think: @mvillaseco @santiagopaolucci @andresmoschini

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -12,7 +12,7 @@
     -->
 
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
-    <link rel="stylesheet" type="text/css" href="https://cdn.fromdoppler.com/doppler-ui-library/build470/css/styles.css">
+    <link rel="stylesheet" type="text/css" href="https://cdn.fromdoppler.com/doppler-ui-library/v3.9.5/css/styles.css">
     <link rel="stylesheet" type="text/css" href="zoho-chat.css">
     <!--
       Notice the use of %PUBLIC_URL% in the tags above.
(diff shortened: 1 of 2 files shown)


andresmoschini:
Estuvimos charlando con @GusBaamonde, porque en resoluciones un poco más pequeñas de lo normal, se veía un scroll, incluso, en mi máquina, con el browser maximisado también. 

Él va a revisar alternativas, por ejemplo usando `calc` o un alto mínimo para intentar evitar ese problema.

santiagopaolucci:
I saw this scene with Gus. I agree.

public/index.html
```diff
@@ -30,7 +30,7 @@
     <noscript>
       You need to enable JavaScript to run this app.
     </noscript>
-    <div id="root">
+    <div id="root" class="dp-container">
```
fcoronelmakingsense: it's not more clear `app-container` ? or the same

fcoronelmakingsense: or `webapp-container`

GusBaamonde: The nomenclature of classes that we have been using is `dp` -` component name`, `dp` comes from` doppler`

andresmoschini: I understand your concern @fcoronelmakingsense.

I think that you are right in that it is not something _related to our brand_. I saw this class as a utility one, something not related to our brand that could be used by a library client to make things easy.

In that case, what do you think about using a name related to the functionality or behavior related to this element?

The question is: what makes this _container_ different than another container, for example, a simple _div_?

fcoronelmakingsense: i talk with @GusBaamonde and we think can use something like `dp-app-container` to mantein the prefix

PR #336, opened 2019-08-05
feat(google adwords): add in login confirmation email

fcoronelmakingsense:
- Add google adwords when come from email confirmation and have in url the parameter `activationInProgress%20=%20true`

- Reference link https://bernardodiasdacruz.com/2018/04/10/integrate-google-analytics-with-react-router-v4/

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -49,6 +49,15 @@
     -->
     <script type="text/javascript" src="https://cdn.fromdoppler.com/doppler-ui-library/v3.9.5/js/app.js"></script>
     <script type="text/javascript" src="zoho-chat.js"></script>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=AW-1065197040/ZA62CKv_gZEBEPC79vsD"></script>
+    <script>
+      if (window.location.hostname !== 'localhost') {
+        window.dataLayer = window.dataLayer || [];
+        function gtag(){dataLayer.push(arguments);}
+        gtag('js', new Date());
+        gtag('config', 'AW-1065197040/ZA62CKv_gZEBEPC79vsD');
+      }
+    </script>
     <link rel="stylesheet" type="text/css" href="zoho-chat.css">
   </body>
 </html>
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -4,6 +4,7 @@ import '@testing-library/jest-dom/extend-expect';
 import App from './App';
 import { AppServicesProvider } from './services/pure-di';
 import { MemoryRouter as Router, withRouter } from 'react-router-dom';
+import Login from './components/Login/Login';
 
 function createDoubleSessionManager(appSessionRef) {
   const double = {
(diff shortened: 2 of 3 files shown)


src/components/Login/Login.js
```diff
@@ -82,6 +89,13 @@ const Login = ({ intl, location, dependencies: { dopplerLegacyClient, sessionMan
 
   const formMessage = useMemo(() => getForgotErrorMessage(location), [location]);
 
+  const activationInProgress = useMemo(() => isActivactionInProgress(location), [location]);
+
+  if(activationInProgress && typeof(window.gtag) === 'function') {
```
andresmoschini:
```suggestion
  if (activationInProgress && typeof(window.gtag) === 'function') {
```

andresmoschini: Why `window.gtag` could be not a function?

src/components/Login/Login.js
```diff
@@ -82,6 +89,13 @@ const Login = ({ intl, location, dependencies: { dopplerLegacyClient, sessionMan
...
+    const gtag = window.gtag;
+    gtag('event', 'conversion', {'send_to': 'AW-1065197040/ZA62CKv_gZEBEPC79vsD'});
```
andresmoschini: why not directly this?

```diff
-    const gtag = window.gtag;
-    gtag('event', 'conversion', {'send_to': 'AW-1065197040/ZA62CKv_gZEBEPC79vsD'});
+    window.gtag('event', 'conversion', {'send_to': 'AW-1065197040/ZA62CKv_gZEBEPC79vsD'});
```

src/components/Login/Login.js
andresmoschini: Is it possible testing it in any way?

andresmoschini: I mean, automatically.

src/components/Login/Login.js
andresmoschini: Dis you tested it manually? Does it generate a remote request?

src/components/Login/Login.js
fcoronelmakingsense: only validation if library doesnt load propertly

andresmoschini: 👍 I understand. We do not want to break our application if the library's CDD is down.

src/components/Login/Login.js
fcoronelmakingsense: yes, it's generated a remote request

src/components/Login/Login.js
fcoronelmakingsense: automatically? you said unit testing?

andresmoschini: right

fcoronelmakingsense: i really don't know, because the library use in the index.html. Maybe only we can simulate the response?

public/index.html
```diff
@@ -49,6 +49,15 @@
     -->
     <script type="text/javascript" src="https://cdn.fromdoppler.com/doppler-ui-library/v3.9.5/js/app.js"></script>
     <script type="text/javascript" src="zoho-chat.js"></script>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=AW-1065197040/ZA62CKv_gZEBEPC79vsD"></script>
+    <script>
+      if (window.location.hostname !== 'localhost') {
```
andresmoschini: Why did you apply this check? 

I am asking because it will avoid tracking localhost, but it will still track our navigation in the CDN, is it fine?

src/components/Login/Login.js
andresmoschini: Index.html cannot be tested, but I think that _Login.js_ could be tested, assuming that _Index.html_ works fine.

let us talk to discuss it if you want.

src/components/Login/Login.js
andresmoschini: Great!

public/index.html
fcoronelmakingsense: Maybe yes, because us use CDN for testing too, right?

andresmoschini: I do not know, I asked first!! :P

src/components/Login/Login.js
fcoronelmakingsense: i added the test

cbernat:
Great tests!

cbernat:
Instead of fix, the second commit should be perhaps chore? what do you think?

src/App.test.js
```diff
@@ -715,4 +716,50 @@ describe('App component', () => {
...
     });
   });
+
+  describe('activation in progress parameter', () => {
+    it('should call google adwords', () => {
+      // Arrange
+      const dependencies = {
+        window: {
+          gtag: jest.fn(),
+          Menubutton: MenubuttonFake,
```
andresmoschini: Is this line required?

fcoronelmakingsense: yes, because use in `LanguageSelector` component

src/App.test.js
```diff
@@ -715,4 +716,50 @@ describe('App component', () => {
...
+            <App locale="en" />
+          </Router>
+        </AppServicesProvider>,
+      );
+
+      // Assert
+      expect(dependencies.window.gtag).toBeCalledTimes(1);
+    });
+
+    it('should not to be call google adwords', () => {
```
andresmoschini: Could you include the condition?

```suggestion
    it('should not to be call google adwords when `activationInProgress` is not present in the query string', () => {
```

fcoronelmakingsense:
> Instead of fix, the second commit should be perhaps chore? what do you think?

Good catch! i changed. Thanks!

src/App.test.js
```diff
@@ -715,4 +716,50 @@ describe('App component', () => {
       expect(localStorageItems['dopplerFirstOrigin.value']).toBeUndefined();
     });
   });
+
+  describe('activation in progress parameter', () => {
+    it('should call google adwords', () => {
```
andresmoschini: Could you include the condition?

```suggestion
    it('should call google adwords when query string contains `activationInProgress%20=%20true`', () => {
```

src/App.test.js
```diff
@@ -715,4 +716,50 @@ describe('App component', () => {
       expect(localStorageItems['dopplerFirstOrigin.value']).toBeUndefined();
     });
   });
+
+  describe('activation in progress parameter', () => {
```
andresmoschini: I think that it is not the right subject, try to replace the it with this value:

* activation in progress parameter should call google adwords
* activation in progress parameter  should not to be call google adwords

I think that _activation in progress parameter_ is not who calls google adwords

src/App.test.js
fcoronelmakingsense: changed

src/App.test.js
```diff
@@ -715,4 +716,50 @@ describe('App component', () => {
...
+          <Router initialEntries={['/login']}>
+            <App locale="en" />
+          </Router>
+        </AppServicesProvider>,
+      );
+
+      // Assert
+      expect(dependencies.window.gtag).not.toBeCalled();
+    });
+  });
```
andresmoschini: Could you please add some more scenarios:

* [x] google adwords should be called when query string contains `activationInProgress%20=%20true`.

    Done.

* [x] google adwords should not be called when query string does not contain `activationInProgress` parameter.

    Done.

* [ ] google adwords should be called when query string contains `activationInProgress=true`.

    `/login?activationInProgress=true`

* [ ] google adwords should not be called when query string `activationInProgress` parameter is not `true`.

    `/login?activationInProgress=another`

* [ ] google adwords should not be called when query string `activationInProgress` has no value.

    `/login?activationInProgress`

fcoronelmakingsense: ready

andresmoschini:
🚀

PR #345, opened 2019-08-09
fix(app): unit testing console log errors

fcoronelmakingsense:
- Update react, react-dom, and react-testing-library versions
- Fix `App.test` and `Signup.test`


src/App.test.js
```diff
@@ -53,18 +54,30 @@ class MenubuttonFake {
...
+const emptyResponse = { success: false, error: new Error('Dummy error') };
+
+const dopplerSitesClientDouble = {
+  getBannerData: async () => {
+    await timeout(0);
+    return emptyResponse;
+  },
+};
+
+const defaultDependencies = {
```
andresmoschini: What do you think about converting it into a function to avoid possible dirty dependencies between tests?

```suggestion
const createDefaultDependencies = () => ({
```

src/App.test.js
```diff
@@ -813,6 +821,7 @@ describe('App component', () => {
           gtag: jest.fn(),
           Menubutton: MenubuttonFake,
         },
+        dopplerSitesClient: dopplerSitesClientDouble,
```
andresmoschini: In these cases where we need the default dependencies with few differences, I think that spread operator could work very well.

src/App.test.js
andresmoschini: We could also move this function to another place to allow us to reuse it easily.

andresmoschini:
Good work @fcoronelmakingsense !

I added some comments, but your work by the moment is very good, so, feel free to apply or not my proposals.

We can continue improving it later.

src/App.test.js
fcoronelmakingsense: Yes maybe we can add in another pr and merge this to fix the test comments if you agree with that

src/components/Signup/Signup.test.js
```diff
@@ -5,21 +5,33 @@ import DopplerIntlProvider from '../../i18n/DopplerIntlProvider';
...
+import { HardcodedDopplerSitesClient } from '../../services/doppler-sites-client.doubles';
+import { timeout } from '../../utils';
 
 class MenubuttonFake {
   init() {}
 }
 
+const emptyResponse = { success: false, error: new Error('Dummy error') };
+
+const defaultDependencies = {
```
cbernat: Perhaps this default dependencies could be in a separate file to be reused by multiple tests. What do you think?

cbernat:
This is looking very nice! Good job @fcoronelmakingsense 😄

src/components/Signup/Signup.test.js
fcoronelmakingsense: Yes is the same comment to said @andresmoschini we can add in another pr to reuse in more test and now in this pr only fix the errors

cbernat:
🚀

PR #354, opened 2019-08-12
chore(shopify): add httpClient service tests

cbernat:


src/services/shopify-client.test.ts
```diff
@@ -0,0 +1,162 @@
...
+  data: [
+    {
+      connectedOn: '2019-08-09T15:13:52.262Z',
+      dopplerListId: '27311899',
+      dopplerListName: 'Shopify Contacto',
+      importedCustomersCount: '6',
+      shopName: 'dopplerdevshop.myshopify.com',
+      shopifyAccessToken: '2741255a41f37341d0fa6d64d58e5c86',
+      syncProcessDopplerImportSubscribersTaskId: 'import-100544891',
+      syncProcessInProgress: 'false',
```
andresmoschini: Nice! 

Could you also add a test with a boolean false value?

It would be to prevent bad behaviors when the response is fixed.

cbernat: Sure.

src/services/shopify-client.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    create: () => axios,
+  } as AxiosStatic;
+  const connectionDataRef = {
+    current: {
+      status: 'authenticated',
+      jwtToken: 'jwtToken',
+      userData: {} as DopplerLegacyUserData,
+    },
+  } as RefObject<AppSession>;
+  const baseUrl = 'http://shopify.test';
```
fcoronelmakingsense: maybe dont need and use in `new HttpShopifyClient({ axiosStatic, 'http://shopify.test', connectionDataRef })`

src/services/shopify-client.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+const notConnected = {
+  data: [],
+};
+
+describe('HttpShopifyClient', () => {
+  it('should set connected shop without list correctly', async () => {
+    // Arrange
+    const request = jest.fn(async () => connectedWithoutListResponse);
+
```
fcoronelmakingsense: maybe can remove this space

src/services/shopify-client.test.ts
```diff
@@ -0,0 +1,162 @@
...
+      jwtToken: 'jwtToken',
+      userData: {} as DopplerLegacyUserData,
+    },
+  } as RefObject<AppSession>;
+  const baseUrl = 'http://shopify.test';
+  const shopifyClient = new HttpShopifyClient({ axiosStatic, baseUrl, connectionDataRef });
+  return shopifyClient;
+}
+
+const connectedWithListReadyResponse = {
```
fcoronelmakingsense: maybe if you dont use in more test, can define in `Arrange`

src/services/shopify-client.test.ts
```diff
@@ -0,0 +1,162 @@
...
+    const shopifyClient = createHttpShopifyClient({ request });
+
+    // Act
+    const result = await shopifyClient.getShopifyData();
+
+    // Assert
+    expect(request).toBeCalledTimes(1);
+    expect(result).not.toBe(undefined);
+    expect(result.success).toBe(true);
+    expect(result.value[0].shopName).toEqual(connectedWithListReadyResponse.data[0].shopName);
```
fcoronelmakingsense: dont need to check with `connectedWithListSyncResponse` ?

src/services/shopify-client.test.ts
cbernat: Good point.

src/services/shopify-client.test.ts
```diff
@@ -0,0 +1,162 @@
...
+  it('should set error connecting shopify', async () => {
+    // Arrange
+    const request = jest.fn(async () => {});
+
+    request.mockImplementation(() => {
+      throw new Error();
+    });
+
+    const shopifyClient = createHttpShopifyClient({ request });
+    console.error = jest.fn(); // silence console error for this test run only
```
fcoronelmakingsense: 😮

andresmoschini: Are you sure that this is not affecting other tests?

cbernat: In the real service when it throws an error, the console shows the error, here it's very ugly to get the console for tests with not needed messages (remember the act warnings)

andresmoschini: I know, but I think that this change is not being cleaned. Is it?

src/services/shopify-client.test.ts
cbernat: @andresmoschini Done!

src/services/shopify-client.test.ts
cbernat: Fixed

src/services/shopify-client.test.ts
cbernat: Fixed.

src/services/shopify-client.test.ts
cbernat: I will test this issue further to not get side effects here. Thanks.

src/services/shopify-client.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    // Act
+    const result = await shopifyClient.getShopifyData();
+
+    // Assert
+    expect(request).toBeCalledTimes(1);
+    expect(result).not.toBe(undefined);
+    expect(result.success).toBe(false);
+    expect(result.error).not.toBe(undefined);
+    console.error = consoleError; // Restore console error logs
```
cbernat: @andresmoschini @fcoronelmakingsense I fixed this issue with `console.error`. Now if I use `console.error` bellow it works as it should.

andresmoschini: It should be in `aftereach` because if the assertion fails, it is never executed.

PR #360, opened 2019-08-13
chore(react-intl): update version to 3.1.6 and fix compatibility

fcoronelmakingsense:
- Fix test
- Add full-icu to can test language local
- Remove unused things
- Remove `addLocaleData` because doesn't use in version 3


src/i18n/en.js
```diff
@@ -12,7 +12,7 @@ const urlShopify = process.env.REACT_APP_SHOPIFY_URL;
 // Common URLs
 const urlHelpAdvancedReports = `${urlHelp}/advanced-reports`;
 const urlPrivacy = `${urlSite}/en/legal/privacy`;
-const urlPrivacyFromSignup = `${urlPrivacy}?utm_source=app&utm_medium=landing&utm_campaign=signup`;
+const urlPrivacyFromSignup = `${urlPrivacy}?utm_source=app&amp;utm_medium=landing&amp;utm_campaign=signup`;
 const urlPrivacyFromLogin = `${urlPrivacy}?utm_source=app&utm_medium=landing&utm_campaign=login`;
```
andresmoschini: I am not sure if it is a good idea encoding the value of `urlPrivacyFromSignup` because it is only a string, it is not part of HTML yet, in the other hand, notice that `urlPrivacyFromLogin` is not encoded.

src/i18n/en.js
```diff
@@ -234,10 +234,7 @@ export default {
     placeholder_email: `Your Email will be your Username`,
     placeholder_password: `Enter your secret key`,
     placeholder_phone: `9 11 2345-6789`,
-    privacy_policy_consent_HTML: `
-      I accept Doppler's
-      <a target="_blank" href="${urlPrivacyFromSignup}" ${patchForBlank}>Privacy Policy</a>.
-      `,
+    privacy_policy_consent_HTML: `I accept Doppler's <a target="_blank" href="${urlPrivacyFromSignup}" ${patchForBlank}>Privacy Policy</a>.`,
```
andresmoschini: In general, we try to have short lines in this file.

src/i18n/en.js
fcoronelmakingsense: Because it's use in the footer with Markdown

src/i18n/en.js
fcoronelmakingsense: i can change

src/i18n/en.js
andresmoschini: I know, but we cannot change the value depending on who uses the variable.

Imagine that someone wants to also use this variable in markdown content.

src/components/Reports/ReportsFilters/ReportsFilters.js
```diff
@@ -39,7 +39,9 @@ const ReportsFilters = ({
                 {domainSelected.verified_date ? (
                   <>
                     <FormattedMessage id="reports_filters.verified_domain" />{' '}
-                    <FormattedDate value={domainSelected.verified_date} timeZone="UTC" />
+                    <span>
```
cbernat: Out of curiosity, why do you need an extra span here?

fcoronelmakingsense: in previous versions the `span` is generated automatically now dont

src/i18n/en.js
fcoronelmakingsense: you suggest use in the key?

src/components/Reports/ReportsFilters/ReportsFilters.js
cbernat: And I asume you need it here because of styles or anything else, ok. Thanks!

src/i18n/en.js
fcoronelmakingsense: image
like this?

andresmoschini: No, maybe one of these options (or another):

* Using a sufix

    ```js
     const urlPrivacyFromSignup_HTMLEncoded = `${urlPrivacy}?utm_source=app&amp;utm_medium=landing&amp;utm_campaign=signup`;
    ```

* Converting it before including in the key. 

    ```js
    const urlPrivacyFromSignup = `${urlPrivacy}?utm_source=app&utm_medium=landing&utm_campaign=signup`;
     const urlPrivacyFromSignup_HTMLEncoded = htmlEncode(urlPrivacyFromSignup);
    ```

    or

    ```js
    const prueba_HTML = `<a target="_blank" href="${htmlEncode(urlPrivacyFromSignup)}" . . .`
    ```

* Changing it by markdown

* Other option?

cbernat:
👍

src/i18n/en.js
fcoronelmakingsense: changed

PR #368, opened 2019-08-15
chore(shopify): refactor use one map to fix test warning

cbernat:
@fcoronelmakingsense the issue you've mentioned earlier fixed here.

### DIFF
--- a/src/components/Integrations/Shopify/Shopify.js
+++ b/src/components/Integrations/Shopify/Shopify.js
@@ -138,60 +138,63 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
             </>
           ) : shopifyState.isConnected ? (
             <>
-              <div className="dp-integration__block">
-                {shopifyHeader}
-                {shopifyState.shops.map((shop) => (
-                  <div key={shop.shopName} className="block dp-integration__status">
-                    <div className="status__info">
-                      <div>
-                        <StyledShopifyLogo />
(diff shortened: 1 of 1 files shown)


src/components/Integrations/Shopify/Shopify.js
```diff
@@ -138,60 +138,64 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
-                                  <FormattedDate value={shop.synchronization_date} timeZone="UTC" />
-                                </strong>
-                              </p>
-                              <p>
-                                {_('shopify.header_store')} <strong> {shop.shopName} </strong>
-                              </p>
-                            </li>
-                          </ul>
+              {shopifyState.shops.map((shop) => (
+                <div key={shop.shopName}>
```
andresmoschini: Is the key duplicated?

andresmoschini: image

andresmoschini: On the other hand, are these keys safe? I mean, we cannot have duplicated names?

cbernat: Yes I've seen them thanks.

cbernat: I don't think we can, they are shopify urls to the store.

andresmoschini: 👍

PR #379, opened 2019-08-21
Doppler API client for Shopify

cbernat:
### Doppler API client
- Add client for subscribers count
- Add tests
- Add validation with experimental features toggle. 
- Add combined test for shopify with doppler api and experimental features.

To test or change values from local storage, we need to refresh the page
build url: https://cdn.fromdoppler.com/doppler-webapp/build1445/#/integrations/shopify

To test in browser console (listId is optional, to test in another enviroment, else it takes shopify list):
```
localStorage.setItem('dopplerExperimental', JSON.stringify({DopplerAPI: {apikey: 'myapikey', listId:1255411}}));
```
When jwtToken is ready, enabling the feature without apikey, would allow a quick test
```
localStorage.setItem('dopplerExperimental', JSON.stringify({DopplerAPI: {}}));
```


src/utils.js
```diff
@@ -34,3 +34,23 @@ export function useInterval({ callback, delay, runOnStart }) {
     }
   }, [delay, runOnStart]);
 }
+
+export function isFeatureEnabled(feature) {
+  const experimentalFeatures = localStorage.getItem('dopplerExpermiental')
+    ? JSON.parse(localStorage.getItem('dopplerExpermiental'))
```
andresmoschini: What do you think about getting and parsing this item only one time?

src/utils.js
```diff
@@ -34,3 +34,23 @@ export function useInterval({ callback, delay, runOnStart }) {
     }
   }, [delay, runOnStart]);
 }
+
+export function isFeatureEnabled(feature) {
```
andresmoschini: What do you think about moving these changes to his own PR?

src/utils.js
```diff
@@ -34,3 +34,23 @@ export function useInterval({ callback, delay, runOnStart }) {
...
+    return !!experimentalFeatures.features.find((item) => item === feature);
+  }
+  return false;
+}
+
+export function getFeatureValue(param) {
+  const experimentalFeatures = localStorage.getItem('dopplerExpermiental')
+    ? JSON.parse(localStorage.getItem('dopplerExpermiental'))
+    : null;
+  if (experimentalFeatures) {
```
andresmoschini: Why not simply:

```js
return experimentalFeatures && experimentalFeatures.data[param];
```

Or

```js
return experimentalFeatures && experimentalFeatures.data[param] || false;
```

src/utils.js
```diff
@@ -34,3 +34,23 @@ export function useInterval({ callback, delay, runOnStart }) {
     }
   }, [delay, runOnStart]);
 }
+
+export function isFeatureEnabled(feature) {
+  const experimentalFeatures = localStorage.getItem('dopplerExpermiental')
+    ? JSON.parse(localStorage.getItem('dopplerExpermiental'))
+    : null;
+  if (experimentalFeatures) {
+    return !!experimentalFeatures.features.find((item) => item === feature);
```
andresmoschini: I cannot imagine the structure of `experimentalFeatures` object, is it something like:

```json
{
  features: [ 'feature1', 'feature2'],
  data: {
    'feature1': {},
    'arbitraryName: 5
  }
}
```

What do you think about:

```json
{
  feature1: true,
  feature2: {
    customValue1: "Hola"
  },
  feature3: false,
  feature4: null,
  feature5: 0
}
```

(features4, features45, and features6 are the same than disabled)

src/utils.js
andresmoschini: What do you think about moving it to his own module/file?

And it would be nice injecting this using CI to allow bypass this in the tests.

src/utils.js
cbernat: Now that you mention .. I should move this. Thanks!

src/utils.js
cbernat: Yes I will move this out of this PR.

src/utils.js
cbernat: The structure is in the tests. I did it this way to make a simple initial approach. But maybe it is best to change it with params by feature.

andresmoschini: Do you not like a simple object where the property names are the features and the property values the feature data?

src/utils.js
cbernat: Moved to https://github.com/FromDoppler/doppler-webapp/pull/382

src/utils.js
cbernat: Here I return a value or null, not true/false.

andresmoschini:
Revisando...

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -90,16 +93,33 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+    }
+    return 0;
+  };
+
   useInterval({
     runOnStart: true,
     delay: 20000,
     callback: async () => {
-      const result = await shopifyClient.getShopifyData();
+      let result = await shopifyClient.getShopifyData();
```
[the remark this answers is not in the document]
andresmoschini: No, you do not change it.

You are modifying the same object, maybe that is not a good idea neither: if you change a value or modifying an object, the logic is difficult to read and understand, and it is also more error-prone.

src/components/Integrations/Shopify/Shopify.test.js
```diff
@@ -5,6 +5,11 @@ import DopplerIntlProvider from '../../../i18n/DopplerIntlProvider.double-with-i
 import { AppServicesProvider } from '../../../services/pure-di';
 import Shopify from './Shopify';
 import { SubscriberListState } from '../../../services/shopify-client';
+import {
+  experimentalFeatures,
+  ExperimentalFeatures,
```
andresmoschini: What are these both things with pretty similar names?

src/components/Integrations/Shopify/Shopify.test.js
```diff
@@ -152,4 +157,52 @@ describe('Shopify Component', () => {
     await waitForDomChange();
     expect(getByText('validation_messages.error_unexpected_HTML'));
   });
+
+  // Failing test indicating memory leak in shopify component
+  xit('should use DopplerAPI client when feature is enabled', async () => {
```
andresmoschini: xit?

src/components/Integrations/Shopify/Shopify.test.js
cbernat: A mistype or perhaps autocode fix, I will remove this right away thanks!

src/services/app-session.ts
```diff
@@ -20,6 +20,11 @@ export interface ShopifyConnectionData {
   jwtToken: string;
 }
 
+export interface DopplerAPIConnectionData {
```
andresmoschini: I think that the type `DopplerAPIConnectionData` is not necessary. 

Or, if it is, it should be a private type of DopplerApiClient.

src/components/Integrations/Shopify/Shopify.test.js
cbernat: To skip this test  -->https://jestjs.io/docs/en/api#testskipname-fn

src/services/doppler-api-client.ts
```diff
@@ -0,0 +1,78 @@
...
+    const connectionData = this.connectionDataRef.current;
+    if (
+      !connectionData ||
+      connectionData.status !== 'authenticated' ||
+      !connectionData.jwtToken ||
+      !connectionData.userData
+    ) {
+      throw new Error('Doppler API connection data is not available');
+    }
+    return { jwtToken: connectionData.jwtToken, userAccount: connectionData.userData.user.email };
```
andresmoschini: This method structure is different than other related implementations:

https://github.com/FromDoppler/doppler-webapp/blob/30cbe6fdcb9a4216360ba001025aa9dabb6b0982/src/services/shopify-client.ts#L49-L55

https://github.com/FromDoppler/doppler-webapp/blob/30cbe6fdcb9a4216360ba001025aa9dabb6b0982/src/services/datahub-client.ts#L68-L78

In that cases, the resulting object is the same connection data, but casted with the expected structure.

I understand the convenience of this method, so I am fine on using in this way.

Only two things:

* `DopplerAPIConnectionData` type should be scoped because it is not used in other places.

* Maybe you could use deconstruction here:

    ```suggestion
        return { connectionData: { jwtToken }, connectionData: { userData: { user: { userAccount: email } } } };
    ```

src/services/doppler-api-client.ts
```diff
@@ -0,0 +1,78 @@
...
+          : SubscriberListState.synchronizingContacts,
+    };
+  }
+
+  public async getListData(
+    listId: number,
+    apikey: string,
+  ): Promise<ResultWithoutExpectedErrors<SubscriberList>> {
+    try {
+      // eslint-disable-next-line
```
andresmoschini: Is it really necessary?

src/services/pure-di.tsx
```diff
@@ -131,6 +133,18 @@ export class AppCompositionRoot implements AppServices {
     );
   }
 
+  get dopplerAPIClient() {
+    return this.singleton(
+      'dopplerAPIClient',
+      () =>
+        new HttpDopplerAPIClient({
+          axiosStatic: this.axiosStatic,
+          baseUrl: this.appConfiguration.shopifyUrl,
```
andresmoschini: You should use another configuration here, right?

src/services/doppler-api-client.ts
cbernat: To avoid `jwtToken` warning of not used variable

andresmoschini: Thanks for the explanation, maybe you could add it to the comment.

Another option is that since it is a temporal code, you could keep the warning.

Another one could be avoiding deconstructing that value while you do not need it.

src/services/pure-di.tsx
```diff
@@ -131,6 +133,18 @@ export class AppCompositionRoot implements AppServices {
     );
   }
 
+  get dopplerAPIClient() {
```
andresmoschini: What do you think about using the c# convention of using uppercase only for the first letter of acronyms with more than two letters?

andresmoschini:
```suggestion
  get dopplerApiClient() {
```

src/components/Integrations/Shopify/Shopify.js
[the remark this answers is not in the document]
andresmoschini: What do you think about moving all this logic related to gathering the data (shopify, doppler, experimental feature, etc) in a function?

In that way, this code would be simpler.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -90,16 +93,34 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
     </>
   );
 
+  const getSubscribersAmountFromAPI = async (listId, apikey) => {
+    const resultAPI = await dopplerAPIClient.getListData(listId, apikey);
+    if (resultAPI.success) {
+      return resultAPI.value.amountSubscribers;
+    }
+    return 0;
```
andresmoschini: I think that showing zero when there is an error is not a good idea.

src/services/doppler-api-client.ts
```diff
@@ -0,0 +1,78 @@
...
+      const { jwtToken, userAccount } = this.getDopplerAPIConnectionData();
```
andresmoschini: What do you think about using experimental data value here too?

andresmoschini: So, if the experimental feature `DopplerAPI` is disabled, the JWT token is returned as _jwtToken_ in the contrary, the API key stored in experimental data is returned as _jwtToken_.

src/services/doppler-api-client.ts
```diff
@@ -0,0 +1,78 @@
...
+      const response = await this.axios.request({
+        method: 'GET',
+        url: `http://newapiint.fromdoppler.net/accounts/${userAccount}/lists/${listId}?api_key=${apikey}`,
+        //headers: { Authorization: `token ${jwtToken}` }, // This line will remain commented until jwtToken is enabled for dopplerAPI
```
andresmoschini: _apiKey_ and _jwtToken_  are interchangeable. You do not need comment code here.

src/components/Integrations/Shopify/Shopify.js
cbernat: It would be an api error and not a shopify error, perhaps in this case show shopify customers? or log the error...

src/services/doppler-api-client.ts
```diff
@@ -0,0 +1,78 @@
...
+      });
+      if (response.status === 200 && response.data) {
+        return { success: true, value: this.mapList(response.data) };
+      } else {
+        return { success: false, error: response.statusText };
```
andresmoschini: I think that in general Doppler API stores the error message in a very specific place.

Maybe `response.title` (if it is available), see an example of Relay API Client (very similar):

https://github.com/FromDoppler/relay-webapp/blob/2b4141cb7c276d96488aa12e9d2301c7bbc9985a/src/wwwroot/interceptors/errorHandlerInterceptor.js#L29-L35

src/services/doppler-api-client.ts
```diff
@@ -0,0 +1,78 @@
...
+      throw new Error('Doppler API connection data is not available');
+    }
+    return { jwtToken: connectionData.jwtToken, userAccount: connectionData.userData.user.email };
+  }
+
+  private mapList(data: any): SubscriberList {
+    return {
+      name: data.name,
+      id: data.listId,
+      amountSubscribers: data.subscribersCount,
```
andresmoschini: Why not?

```suggestion
      subscribersCount: data.subscribersCount,
```

src/services/doppler-api-client.ts
cbernat: I don't see this work as expected 😕

andresmoschini: Do not worry, I am not so much _canchero_ with deconstructing.

src/services/doppler-api-client.ts
cbernat: Yes I've seen the response for error and I thought it was more descriptive statusText than title, but sure I can change it.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -94,32 +94,38 @@ const Shopify = ({
...
+    const resultAPI = await dopplerApiClient.getListData(listId, apikey);
     if (resultAPI.success) {
       return resultAPI.value.amountSubscribers;
+    } else {
+      return null;
     }
-    return 0;
+  };
+
+  const updateSubscriberCount = async (list) => {
```
andresmoschini: Yo proponía unir esta lógica con `shopifyClient.getShopifyData`, para dejar toda la mugre en un solo lugar.

src/services/doppler-api-client.ts
```diff
@@ -66,13 +66,12 @@ export class HttpDopplerAPIClient implements DopplerAPIClient {
     apikey: string,
   ): Promise<ResultWithoutExpectedErrors<SubscriberList>> {
     try {
-      // until jwtToken is enabled disable use of variable rule
-      // eslint-disable-next-line
-      const { jwtToken, userAccount } = this.getDopplerAPIConnectionData();
+      const { userAccount } = this.getDopplerApiConnectionData();
+      const jwtToken = apikey ? apikey : this.getDopplerApiConnectionData().jwtToken;
```
andresmoschini: Que opinás acerca de no aceptar el parámetro apikey y directamentes sacarlo de experimentalFeatures.

src/components/Integrations/Shopify/Shopify.js
cbernat: O sea manejar todo desde el servicio 👀 no?

src/services/doppler-api-client.ts
cbernat: I thought about this, but then I chickened out

src/components/Integrations/Shopify/Shopify.js
andresmoschini: Mmm, en el servicio es dificil, porque nuestros servicios por ahora son clientes de un solo backend.

Y no se si vale la pena crear una nueva capa.

Pero se podría hacer una función local _getShopifyData_ que mezcle (o no, según el caso) al resultado de ambos servicios.

src/services/doppler-api-client.ts
andresmoschini: Haha, nah, let me know and we could pair programming it if you need some encouragement.

src/services/doppler-api-client.ts
cbernat: I'm not so sure, amount and count is debatable...

andresmoschini: Right, but at least it is the same than before :P

src/components/Integrations/Shopify/Shopify.js
cbernat: Bien pero en ese caso, todo el seteo de estados y demas estarían tambien dentro de esta nueva funcion, porque el caso donde me interesa actualizar los datos es solamente cuando tengo la lista, sino no.

andresmoschini: Yo propongo hacer una abstracción solo de la obtención de datos. Que esa función se encargue de obtener y mergear los datos de las fuentes que haga falta.

Luego, la actualización de estado, etc. quedaría afuera, y sería agnóstica del origen de los datos.

Tal vez haya algo que no estoy viendo, cualquier cosa llamame.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -90,6 +93,27 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+    const dopplerAPIFeature = experimentalFeatures && experimentalFeatures.getFeature('DopplerAPI');
+    if (list && dopplerAPIFeature && dopplerAPIFeature.apikey) {
+      const subscribersCount = await getSubscribersAmountFromAPI(
+        dopplerAPIFeature.listId ? dopplerAPIFeature.listId : list.id,
+        dopplerAPIFeature.apikey,
+      );
+      list.amountSubscribers = subscribersCount ? subscribersCount : list.amountSubscribers;
```
```suggestion
      list.amountSubscribers = subscribersCount || list.amountSubscribers;
```

Or better yet:

```suggestion
      list.amountSubscribers = subscribersCount != null ? subscribersCount : list.amountSubscribers;
```

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -90,16 +93,43 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
+  const getSubscribersAmountFromAPI = async (listId, apikey) => {
+    const resultAPI = await dopplerApiClient.getListData(listId, apikey);
+    if (resultAPI.success) {
+      return resultAPI.value.amountSubscribers;
+    } else {
+      return null;
+    }
+  };
+
+  const updateSubscriberCount = async (list) => {
```
andresmoschini: Personally, I would put this code directly in `getShopifyData`, but it is only a personal opinion, so hard opinion about it.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -90,16 +93,43 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
     </>
   );
 
+  const getSubscribersAmountFromAPI = async (listId, apikey) => {
```
andresmoschini: Maybe, I would also put this code in `getShopifyData`.

src/components/Integrations/Shopify/Shopify.js
```diff
@@ -90,16 +93,43 @@ const Shopify = ({ intl, dependencies: { shopifyClient } }) => {
...
     delay: 20000,
     callback: async () => {
-      const result = await shopifyClient.getShopifyData();
+      const result = await getShopifyData();
       if (!result.success) {
         setShopifyState({
           error: <FormattedHTMLMessage id="validation_messages.error_unexpected_HTML" />,
         });
-      } else if (result.value.length) {
+      } else if (result.value && result.value.length) {
```
andresmoschini: If `result.success` then `result.value` has to have value, am I missing something?

src/components/Integrations/Shopify/Shopify.js
cbernat: Debatable... so I'll pass.

src/components/Integrations/Shopify/Shopify.js
cbernat: Same here arguable.. 😄

src/components/Integrations/Shopify/Shopify.test.js
```diff
@@ -152,4 +169,98 @@ describe('Shopify Component', () => {
...
+          <Shopify />
+        </DopplerIntlProvider>
+      </AppServicesProvider>,
+    );
+
+    // Assert
+    expect(container.querySelector('.loading-box')).toBeInTheDocument();
+    await waitForDomChange();
+    expect(container.querySelector('.dp-integration__status')).toBeInTheDocument();
+    expect(getByText(listExist.value.amountSubscribers.toString()));
```
andresmoschini: Here you are simply searching for `200`. I think that having false positives is too easy.

cbernat: What do you suggest?

andresmoschini: Mmm, maybe find the element and check the value? or, does it have related content? in that case, you could search for something like `200 subscribers`

cbernat: I don't have something like that O_o this is a table cell, with nothing subscribers amount as a number, like this:  `<td>200</td>`

andresmoschini: In that case, do not worry, we could deal with this issue later.

andresmoschini:
:shipit:

src/components/Integrations/Shopify/Shopify.js
cbernat: This is needed when shop is disconnected, value exists but is empty.

andresmoschini: Mmmm, maybe it is an issue in the client.

If the client returns a list of connected shops, and there are no shops, the list should be empty but should be a list.

PR #382, opened 2019-08-22
Experimental features

cbernat:
Add experimental feature management


src/utils.test.js
```diff
@@ -100,18 +100,18 @@ describe('utils', () => {
           }),
         { initialProps: { delay: 100 } },
       );
-      // Act
+      // Assert
       jest.advanceTimersByTime(600);
       rerender({ delay: null });
       jest.advanceTimersByTime(800);
-      // Assert
```
andresmoschini: these changes are not related to this feature, right?

cbernat: No but they are related to this file, it's a pity not to fix something so trivial.

src/experimental-features.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+  getExperimentalFeatures() {
+    return this.storage.getItem('dopplerExpermiental')
+      ? JSON.parse(this.storage.getItem('dopplerExpermiental'))
+      : null;
+  }
+  isFeatureEnabled(feature) {
+    const experimentalFeatures = this.getExperimentalFeatures();
+    if (experimentalFeatures) {
+      return !!experimentalFeatures[feature];
```
fcoronelmakingsense: i not sure but maybe can replace with `return experimentalFeatures && !!experimentalFeatures[feature];` and dont need return false

src/experimental-features.js
```diff
@@ -0,0 +1,26 @@
+export class ExperimentalFeatures {
```
fcoronelmakingsense: maybe can move to services folder, i don't know what you think

src/experimental-features.js
```diff
@@ -0,0 +1,26 @@
...
+    return false;
+  }
+  getFeatureValue(feature, param) {
+    const experimentalFeatures = this.getExperimentalFeatures();
+    if (experimentalFeatures) {
+      return !!experimentalFeatures[feature] && experimentalFeatures[feature][param]
+        ? experimentalFeatures[feature][param]
+        : null;
+    }
+    return null;
```
fcoronelmakingsense: maybe can replace with `return experimentalFeatures && !!experimentalFeatures[feature] && experimentalFeatures[feature][param]
        ? experimentalFeatures[feature][param]
        : null`

src/experimental-features.test.js
```diff
@@ -0,0 +1,52 @@
+import { ExperimentalFeatures } from './experimental-features';
+
+describe('experimental features', () => {
+  const experimentalFeatures = new ExperimentalFeatures(localStorage);
+  beforeEach(() => {
+    const experimentalFeaturesData = {
```
fcoronelmakingsense: why is inside beforeEach?

src/experimental-features.test.js
```diff
@@ -0,0 +1,52 @@
...
+  });
+  it('should return null when feature value does not exist', () => {
+    // Arrange
+    // Act
+    const value = experimentalFeatures.getFeatureValue('feature1', 'apikey1');
+
+    // Assert
+    expect(value).toBe(null);
+  });
+});
```
fcoronelmakingsense: dont need to test `getExperimentalFeatures` method?

src/experimental-features.js
cbernat: Yes and also convert this into typescript class would be ok too.

src/experimental-features.test.js
cbernat: This method was supposed to be private, perhaps I should change its access.

src/experimental-features.test.js
cbernat: I added this here because there is a afterEach that does the cleanup.

src/experimental-features.js
```diff
@@ -0,0 +1,26 @@
+export class ExperimentalFeatures {
+  constructor(storage) {
+    this.storage = storage;
+  }
+  getExperimentalFeatures() {
```
andresmoschini: What do you think about executing this and catching the value in the constructor?

I know that in that way if the local storage is updated we will not get the new values, but I think that in general, we do not want to do that.

On the other hand, reading the local storage is expensive and this code will be executed a lot of times, even when the feature is disabled (in production).

An alternative to read it in the constructor is using a lazy load pattern.

src/experimental-features.test.js
```diff
@@ -0,0 +1,52 @@
+import { ExperimentalFeatures } from './experimental-features';
+
+describe('experimental features', () => {
+  const experimentalFeatures = new ExperimentalFeatures(localStorage);
```
andresmoschini: What do you think about not using the "real" localStorage?

src/experimental-features.test.js
andresmoschini: If you do not use the "real" localStorage, you can prepare each test independently and you do not need to pay attention to the cleanup.

src/experimental-features.js
andresmoschini: What do you think about adding it to our [composition root](https://github.com/FromDoppler/doppler-webapp/blob/master/src/services/pure-di.tsx#L44) ?

cbernat: Yes, I thought that. I will do so.

src/services/experimental-features.js
```diff
@@ -1,26 +1,23 @@
...
-  }
-  getExperimentalFeatures() {
-    return this.storage.getItem('dopplerExpermiental')
+    this.features = this.storage.getItem('dopplerExpermiental')
       ? JSON.parse(this.storage.getItem('dopplerExpermiental'))
       : null;
   }
+  getExperimentalFeatures() {
+    return this.features;
+  }
```
andresmoschini: In my opinion, this method is not necessary.

src/services/experimental-features.js
```diff
@@ -0,0 +1,23 @@
...
+    this.features = this.storage.getItem('dopplerExpermiental')
+      ? JSON.parse(this.storage.getItem('dopplerExpermiental'))
+      : null;
+  }
+  getExperimentalFeatures() {
+    return this.features;
+  }
+  isFeatureEnabled(feature) {
+    const experimentalFeatures = this.getExperimentalFeatures();
+    return experimentalFeatures && !!experimentalFeatures[feature];
```
andresmoschini:
```suggestion
    return this.features && !!this.features[feature];
```

src/services/experimental-features.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+  getFeatureValue(feature, param) {
```
andresmoschini: Why not simply `getFeature(featureName)`, then you can do whatever you want with the result. You do not need `isFeatureEnabled(featureName)` neither.

src/services/experimental-features.test.js
```diff
@@ -0,0 +1,46 @@
...
+    feature2: true,
+    feature3: {
+      apikey: 'myApiKey',
+    },
+  };
+  localStorage.setItem('dopplerExpermiental', JSON.stringify(experimentalFeaturesData));
+  const experimentalFeatures = new ExperimentalFeatures(localStorage);
+
+  it('should validate if a feature is enabled', () => {
+    // Arrange
```
andresmoschini: What do you think about preparing the scenario of each test in place of sharing information?

cbernat: Why would I want that? I would be repeating code right? Maybe I'm missing something...

src/services/experimental-features.js
cbernat: I did it this way just on case one day we want to change the structure of features, that is why validating if it exists is one thing, and getting a value apart, I don't want to have to know how it is implemented in every place I use this, right?

src/services/experimental-features.test.js
andresmoschini: In order to read the test as a little story:

1. You arrange the initial state, the scenario.
2. You execute an action.
3. You verify the resulting state.

You want a clear and clean initial scenario to be sure about what is the test doing.

Of course, you could use `beforeEach` to set up a common scenario, but in general, it is not a good idea, because when the scenario is useful for more than a test, probably is not so clean as each test need.

andresmoschini: When setting up the scenario is difficult, you could use helpers to reuse code.

src/services/experimental-features.js
andresmoschini: In my opinion, each _experimental feature_ could have his own _experimental data_ without restriction about structure, it could be `true`, `5`, `"Any string"`, `{ testProperty: 1, another: "This is a string" }`. We do not need to pay attention to the payload format.

Regarding change structure of the features, it is not so much important if we do not define it. The experimental features will be dying and we can use an arbitrary structure for the new ones.

cbernat: A sound argument, thanks!

src/services/experimental-features.test.js
cbernat: I'll try to fine tune each scenario and see what I get.

src/experimental-features.test.js
cbernat: Removed this method.

src/services/local-storage-double.js
```diff
@@ -0,0 +1,9 @@
+export class FakeLocalStorage {
```
andresmoschini: I would use a simpler version in `experimental-features.test.js`, but this file looks nice and useful in other scenarios.

Maybe we should move it to a specific folder like `/test-utils` or something like that.

src/services/experimental-features.js
```diff
@@ -0,0 +1,10 @@
+export class ExperimentalFeatures {
+  constructor(storage) {
+    this.features = storage.getItem('dopplerExpermiental')
+      ? JSON.parse(storage.getItem('dopplerExpermiental'))
+      : null;
+  }
+  getFeature(featureName) {
+    return this.features && !!this.features[featureName] ? this.features[featureName] : null;
```
andresmoschini: What do you think about this:

```suggestion
    return this.features && this.features[featureName] || null;
```

src/services/experimental-features.test.js
```diff
@@ -0,0 +1,65 @@
...
+    const storage = new FakeLocalStorage();
+    storage.setItem('dopplerExpermiental', JSON.stringify(experimentalFeaturesData));
+    const experimentalFeatures = new ExperimentalFeatures(storage);
+    // Act
+    const value = experimentalFeatures.getFeature('feature3').apikey;
+
+    // Assert
+    expect(value).toBe('myApiKey');
+  });
+  it('should return null when feature value does not exist', () => {
```
andresmoschini: What do you think about adding some tests for _falsy_ feature values?

For example:

```js
// Arrange
const experimentalFeaturesData = {
  feature1: false // and also 0, '', null, undefined
};

// . . .

// Act
const value = experimentalFeatures.getFeature('feature1');

// Assert
expect(value).toBeNull;
```

andresmoschini: This [Jest functionality](https://jestjs.io/docs/en/api.html#2-describeeach-table-name-fn-timeout) looks great to test the same with different input values:

```js
describe.each`
  featureValue
  ${null}
  ${0}
  ${false}
  ${''}
`('should return null when feature value is falsy', ({featureValue}) => {
  test(`(featureValue is ${featureValue})`, () => {
    // Arrange
    const experimentalFeaturesData = {
      feature1: featureValue
    };

    // . . .

    // Act
    const value = experimentalFeatures.getFeature('feature1');

    // Assert
    expect(value).toBeNull;
  });
});

```

cbernat: This is very nice and work wonders. 😄

cbernat:
@andresmoschini @fcoronelmakingsense if there are no further suggestions this will be merged very soon.

PR #395, opened 2019-08-30
refactor(ReportsPageRanking): change service method

fcoronelmakingsense:
- add withEmail field for visitors with email
- fix test
- change service method to manage errors

### DIFF
--- a/src/services/datahub-client.ts
+++ b/src/services/datahub-client.ts
@@ -8,6 +8,13 @@ export interface DomainEntry {
   verified_date: Date | null;
 }
 
+export interface Page {
+  name: string;
+  totalVisitors: number;
+  url: string;
+  withEmail: number;
+}
+
 export interface TrafficSource {
   sourceName: string;
   quantity: number;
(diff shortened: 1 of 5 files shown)


andresmoschini:
❤️

src/services/datahub-client.ts
```diff
@@ -123,23 +132,37 @@ export class HttpDatahubClient implements DatahubClient {
...
+        {
+          startDate: dateFrom.toISOString(),
+          sortBy: 'visitors',
+        },
+      );
 
-    // By the moment we are hard-coding it because DataHub does not have this
-    // information. I am looking you Leo :P
-    const urlSchema = 'http://';
+      // By the moment we are hard-coding it because DataHub does not have this
```
cbernat: I think this comment does not apply anymore right?

fcoronelmakingsense: we dont have the schema because datahub dont provide us, then need to use this for now :(

andresmoschini: Related ticket: [DBR-271](http://jira.makingsense.com/browse/DBR-271)

cbernat:
🚀

PR #407, opened 2019-09-08
refactor(react-intl): update to use hook instead of injectintl

fcoronelmakingsense:
- Change injectintl for useIntl() hook the other is obsolete and don't need anymore

### DIFF
--- a/src/components/Login/LoginErrorAccountNotValidated.js
+++ b/src/components/Login/LoginErrorAccountNotValidated.js
@@ -1,46 +1,44 @@
 import React, { useState } from 'react';
 import { useCaptcha } from '../form-helpers/captcha-utils';
-import { injectIntl, FormattedHTMLMessage } from 'react-intl';
+import { useIntl, FormattedHTMLMessage } from 'react-intl';
 import { InjectAppServices } from '../../services/pure-di';
 
-export const LoginErrorAccountNotValidated = injectIntl(
-  InjectAppServices(
-    /**
-     * @param { Object } props
-     * @param { string } props.email
-     * @param { import('react-intl').InjectedIntl } props.intl
-     * @param { import('../../services/pure-di').AppServices } props.dependencies
-     */
-    ({ intl, email, dependencies: { dopplerLegacyClient } }) => {
-      const [resentTimes, setResentTimes] = useState(0);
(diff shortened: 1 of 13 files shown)


src/components/Login/LoginErrorAccountNotValidated.js
```diff
@@ -1,46 +1,45 @@
...
-     * @param { import('../../services/pure-di').AppServices } props.dependencies
-     */
-    ({ intl, email, dependencies: { dopplerLegacyClient } }) => {
-      const [resentTimes, setResentTimes] = useState(0);
-      const [Captcha, verifyCaptcha] = useCaptcha();
+export const LoginErrorAccountNotValidated = InjectAppServices(
+  /**
+   * @param { Object } props
+   * @param { string } props.email
+   * @param { import('react-intl').InjectedIntl } props.intl
```
andresmoschini: This parameter does not exist anymore, right?

cbernat:
Nice work @fcoronelmakingsense

andresmoschini:
Great work @fcoronelmakingsense !

Just to learn a little more, why did they prefer a _hook_ (`useIntl`) over a _hoc_ (a high order component `injectIntl`)? Maybe you have a link where they talk about the advantages.

andresmoschini:
There is a commentary here, but it does not talk about advantages or disadvantages, it seems to be almost the same.

`injectIntl` is not deprecated, right? I mean for our code-base style `useIntl` is nicer, but `injectIntl` is totally supported. Am I missing something?

fcoronelmakingsense:
> There is a commentary here, but it does not talk about advantages or disadvantages, it seems to be almost the same.
> 
> `injectIntl` is not deprecated, right? I mean for our code-base style `useIntl` is nicer, but `injectIntl` is totally supported. Am I missing something?

Sorry i forgot to put the link. It's not deprecated you right. https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md#new-useintl-hook-as-an-alternative-of-injectintl-hoc

src/components/Login/LoginErrorAccountNotValidated.js
fcoronelmakingsense: yes! i removed comment now

PR #410, opened 2019-09-09
refactor(datahub-client): add with email field traffic sources reports

fcoronelmakingsense:
- Add withEmail field for traffic sources method to prepare for full detail information with and without email

### DIFF
--- a/src/services/datahub-client.test.ts
+++ b/src/services/datahub-client.test.ts
@@ -33,10 +33,12 @@ const fullTrafficSourceResponse = {
       {
         sourceName: 'Email',
         quantity: 2000,
+        withEmail: 500,
       },
       {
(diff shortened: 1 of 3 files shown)


src/services/datahub-client.test.ts
```diff
@@ -33,10 +33,12 @@ const fullTrafficSourceResponse = {
       {
         sourceName: 'Email',
         quantity: 2000,
```
cbernat: Perhaps in this PR you could also add another commit that transforms this quantity into `withoutEmail` field?

fcoronelmakingsense: dont use other field. The idea is show in the view `quantity - withEmail` like in page ranking report

PR #411, opened 2019-09-09
refactor(styles): add new folder for styles

fcoronelmakingsense:
- Add new folder for styles to can add more files like fonts, mixins, etc..
- Change colors names to lowerCamelCase


cbernat:
🚀

PR #412, opened 2019-09-09
fix(shopify): prevent re render when data does not change

cbernat:
When state changes, react component is re rendered, in this case we are using polling to keep data up to date, and sometimes we received the same data.
- Added new feature in Utils,  `ResponseCache` to allow to cache responses when etag header is exposed. 
- Used this feature in shopify client
- Refactor shopify component to prevent re setting state when not needed.
- Fixed tests

https://cdn.fromdoppler.com/doppler-webapp/build1501/#/integrations/shopify

### DIFF
--- a/src/services/shopify-client.ts
+++ b/src/services/shopify-client.ts
@@ -2,6 +2,7 @@ import { ResultWithoutExpectedErrors } from '../doppler-types';
 import { AxiosInstance, AxiosStatic } from 'axios';
 import { AppSession, ShopifyConnectionData } from './app-session';
 import { RefObject } from 'react';
+import { ResponseCache } from './../utils';
 export enum SubscriberListState {
   ready,
   synchronizingContacts,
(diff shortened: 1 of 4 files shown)


src/services/shopify-client.ts
```diff
@@ -80,14 +82,17 @@ export class HttpShopifyClient implements ShopifyClient {
...
+      return this.responseCache.getCachedOrMap(this.getShopifyData, response, (r) => {
+        if (r.data && r.data.length) {
+          const connectedShops = r.data.map((shop: any) => {
+            return this.mapShop(shop);
+          });
+          return { success: true, value: connectedShops };
+        } else {
+          return { success: true, value: [] };
+        }
+      });
```
andresmoschini:
```js
return this.responseCache.getCachedOrMap(this.getShopifyData, response, (r) =>
  (r.data && r.data.length) ? {
    success: true, 
    value: r.data.map((shop: any) => {
      return this.mapShop(shop);
    }
  } : { success: true, value: [] }
});
```

Or better yet:

```js
return this.responseCache.getCachedOrMap(this.getShopifyData, response, (r) =>
  (r.data && r.data.length) ? { success: true, value: r.data.map(this.mapShop) }
    : { success: true, value: [] }
});
```

PR #414, opened 2019-09-10
refactor(utils): convert utils.ts to typescript

andresmoschini:
Hi team, now we have types in out _utils_ package :)


cbernat:
👍

fcoronelmakingsense:
Nice! the idea is in the future move to Typescript? or only the services and common stuff, what do you think?

andresmoschini:
> Nice! the idea is in the future move to Typescript? or only the services and common stuff, what do you think?

I have no preference. I am not sure if we need to define it.

PR #417, opened 2019-09-11
feat(reports): add format for tooltip date and data

cbernat:
Added format for name and value inside c3 tooltip configuration:

### In spanish:
image

### In English:
image


andresmoschini:
👍

PR #418, opened 2019-09-11
refactor(datahub-client): change daily visits to accept hours too

fcoronelmakingsense:
- Change names to use this method for days and for hours too

### DIFF
--- a/src/components/Reports/Reports.test.js
+++ b/src/components/Reports/Reports.test.js
@@ -49,12 +49,12 @@ describe('Reports page', () => {
   it('should render domains without pages', async () => {
     const datahubClientDouble = {
       getAccountDomains: async () => fakeData,
-      getVisitsByPeriod: async () => 0,
+      getTotalVisitsOfPeriod: async () => 0,
       getPagesRankingByPeriod: async () => {
         return { success: false, value: [] };
       },
       getTrafficSourcesByPeriod: async () => [],
-      getDailyVisitsByPeriod: async () => [],
+      getVisitsQuantitySummarizedByPeriod: async () => [],
     };
 
     const { getByText } = render(
@@ -87,7 +87,7 @@ describe('Reports page', () => {
     });
     const datahubClientDouble = {
       getAccountDomains: () => getAccountDomainsPromise,
-      getVisitsByPeriod: async () => 0,
+      getTotalVisitsOfPeriod: async () => 0,
       getPagesRankingByPeriod: async () => [],
       getPagesTrafficSourcesByPeriod: async () => [],
     };
--- a/src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
+++ b/src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
@@ -64,9 +64,10 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
   useEffect(() => {
     const fetchData = async () => {
       setState({ loading: true });
-      const dailyVisitsData = await datahubClient.getDailyVisitsByPeriod({
+      const dailyVisitsData = await datahubClient.getVisitsQuantitySummarizedByPeriod({
         domainName: domainName,
         dateFrom: dateFrom,
+        periodBy: 'days',
       });
       if (!dailyVisitsData.success) {
         setState({ loading: false });
--- a/src/services/datahub-client.doubles.ts
+++ b/src/services/datahub-client.doubles.ts
@@ -3,8 +3,9 @@ import {
   emailFilterOptions,
   DomainEntry,
   TrafficSourceResult,
-  DailyVisitsResult,
+  VisitsQuantitySummarizedResult,
   PageRankingResult,
+  filterByPeriodOptions,
 } from './datahub-client';
 import { timeout } from '../utils';
 
@@ -123,7 +124,7 @@ export class HardcodedDatahubClient implements DatahubClient {
     // return [];
   }
 
-  public async getVisitsByPeriod({
+  public async getTotalVisitsOfPeriod({
     domainName,
     dateFrom,
     emailFilter,
@@ -132,7 +133,7 @@ export class HardcodedDatahubClient implements DatahubClient {
     dateFrom: Date;
     emailFilter: emailFilterOptions;
   }): Promise<number> {
-    console.log('getVisitsByPeriod', { domainName, dateFrom, emailFilter });
+    console.log('getTotalVisitsOfPeriod', { domainName, dateFrom, emailFilter });
     await timeout(1500);
     const visits = Math.round(Math.random() * (100 - 1) + 1);
     return visits;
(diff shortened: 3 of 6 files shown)


src/services/datahub-client.doubles.ts
```diff
@@ -191,14 +191,16 @@ export class HardcodedDatahubClient implements DatahubClient {
...
+  public async getDailyAndHoursVisitsByPeriod({
     domainName,
     dateFrom,
+    periodByhours,
   }: {
     domainName: string;
     dateFrom: Date;
-  }): Promise<DailyVisitsResult> {
-    console.log('getDailyVisitsByPeriod', { domainName, dateFrom });
+    periodByhours: boolean;
```
andresmoschini: In general, using boolean parameters is not a good idea, because they are not so descriptive and they does not scale well. For example if `period` is not hour, is it day? is it month? week?.

Maybe you can use two public methods calling a common private one.

andresmoschini: Also, Relay API has a nice way to deal with that, I cannot find it now, but we can talk it later.

src/services/datahub-client.ts
```diff
@@ -199,23 +203,25 @@ export class HttpDatahubClient implements DatahubClient {
...
-  }): Promise<DailyVisitsResult> {
+    periodByhours: boolean;
+  }): Promise<DailyAndHoursVisitsResult> {
     try {
       const response = await this.customerGet<any>(
         `domains/${domainName}/events/quantity-summarized-by-period`,
         {
           startDate: dateFrom.toISOString(),
-          periodBy: 'days',
+          periodBy: periodByhours ? 'hours' : 'days',
```
andresmoschini: What do you think about exposing it as a union type and accept it as a parameter? you can also make one of these values as default using deconstruction:

```ts
type SummaryPerio = 'hours' | 'days';

// . . .

  public async getVisitsByPeriod({
    domainName,
    dateFrom,
    periodBy = 'days',
  }: {
```

Also `getVisitsByPeriod` seems to be more clear.

src/services/datahub-client.ts
```diff
@@ -199,23 +203,25 @@ export class HttpDatahubClient implements DatahubClient {
...
         },
       );
 
-      const dailyVisits = response.data.periods.map((x: any) => ({
+      const visits = response.data.periods.map((x: any) => ({
```
andresmoschini: What do you think about:

```suggestion
      const visitsByPeriod = response.data.periods.map((x: any) => ({
```

Or

```suggestion
      const periodVisits = response.data.periods.map((x: any) => ({
```

fcoronelmakingsense: fixed

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -44,9 +44,10 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
   useEffect(() => {
     const fetchData = async () => {
       setState({ loading: true });
-      const dailyVisitsData = await datahubClient.getDailyVisitsByPeriod({
+      const dailyVisitsData = await datahubClient.getDailyAndHoursVisitsByPeriod({
         domainName: domainName,
         dateFrom: dateFrom,
+        periodBy: 'days',
```
andresmoschini: I would prefer `groupBy` or `groupByPeriod` but DataHub API already accept the parameter `periodBy`, so, it seems that this one is the best name.

src/services/datahub-client.ts
```diff
@@ -199,23 +205,25 @@ export class HttpDatahubClient implements DatahubClient {
     }
   }
 
-  public async getDailyVisitsByPeriod({
+  public async getDailyAndHoursVisitsByPeriod({
```
andresmoschini: What do you think about a name more similar to DataHub API, for example: `getVisitsQuantitySummarizedByPeriod`?

fcoronelmakingsense: oh y didnt see that. Yes i can change. its more clear

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
fcoronelmakingsense: yes i saw Datahub api use this thats why use that

src/services/datahub-client.ts
```diff
@@ -30,9 +30,11 @@ export interface DailyVisits {
 
 export type emailFilterOptions = 'with_email' | 'without_email' | null;
 
+export type filterByPeriodOptions = 'days' | 'hours' | null;
```
andresmoschini:
```suggestion
export type filterByPeriodOptions = 'days' | 'hours';
```

src/services/datahub-client.ts
```diff
@@ -30,9 +30,11 @@ export interface DailyVisits {
...
+
 export interface DatahubClient {
   getAccountDomains(): Promise<DomainEntry[]>;
-  getVisitsByPeriod(query: {
+  getTotalVisitsOffPeriod(query: {
     domainName: number;
     dateFrom: Date;
     emailFilter: emailFilterOptions;
```
andresmoschini:
```suggestion
    emailFilter?: emailFilterOptions;
```

src/services/datahub-client.ts
```diff
@@ -45,12 +47,16 @@ export interface DatahubClient {
...
+    domainName: string;
+    dateFrom: Date;
+    periodBy: filterByPeriodOptions;
+  }): Promise<DailyAndHoursVisitsResult>;
 }
 
 export type TrafficSourceResult = ResultWithoutExpectedErrors<TrafficSource[]>;
 
-export type DailyVisitsResult = ResultWithoutExpectedErrors<DailyVisits[]>;
+export type DailyAndHoursVisitsResult = ResultWithoutExpectedErrors<DailyAndHoursVisits[]>;
```
andresmoschini: What do you think about also apply the new name criteria for this?

src/components/Reports/Reports.test.js
```diff
@@ -49,12 +49,12 @@ describe('Reports page', () => {
   it('should render domains without pages', async () => {
     const datahubClientDouble = {
       getAccountDomains: async () => fakeData,
-      getVisitsByPeriod: async () => 0,
+      getTotalVisitsOffPeriod: async () => 0,
```
cbernat: With this name `getTotalVisitsOffPeriod` you mean visits for this period? that would be with one f -->`getTotalVisitsOfPeriod`

cbernat: Else you are saying visits outside a period.

fcoronelmakingsense: thanks very much. i updated now

src/services/datahub-client.ts
fcoronelmakingsense: changed

src/services/datahub-client.ts
fcoronelmakingsense: fixed

cbernat:
🚀

andresmoschini:
❤️ 

> There are only two hard things in Computer Science: cache invalidation and naming things.
> 
> -- Phil Karlton

Or it was...

> there's two hard problems in computer science: we only have one joke and it's not funny.
> 
> -- clown puter scientist

PR #423, opened 2019-09-16
Daily visits withemail without

cbernat:
Report daily visit graphic without datahub new data, and with classification for email and no email. 

image

- Add with emails in datahub double
- Add values into graphic
- Hide graph for with Email and without
- Fine tune format for value to show a label for each case

### DIFF
--- a/src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
+++ b/src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
@@ -18,38 +18,52 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
       show: false,
     },
     tooltip: {
-      format: {
-        title: (date) => {
-          return intl.formatDate(date, {
+      contents: function(data) {
+        if (data.length) {
+          const date = intl.formatDate(data[0].x, {
             timeZone: 'UTC',
             day: 'numeric',
             month: 'long',
             year: 'numeric',
             weekday: 'long',
           });
-        },
-        value: (value, ratio, id) => {
-          switch (id) {
-            case 'quantity':
-              return (
-                intl.formatMessage({
-                  id: 'reports_daily_visits.tooltip_page_views',
-                }) + value
-              );
-            case 'withEmail':
-              return (
-                intl.formatMessage({
-                  id: 'reports_daily_visits.tooltip_with_email',
-                }) + value
-              );
-            case 'withoutEmail':
-              return (
-                intl.formatMessage({
-                  id: 'reports_daily_visits.tooltip_without_email',
-                }) + value
-              );
-          }
-        },
+
+          const tooltipData = data.reduce((accumulator, item) => {
+            accumulator[item.id] = item.value;
+            return accumulator;
+          }, {});
+
+          tooltipData.quantity_label = intl.formatMessage({
+            id: 'reports_daily_visits.tooltip_page_views',
+          });
+          tooltipData.withEmail_label = intl.formatMessage({
+            id: 'reports_daily_visits.tooltip_with_email',
+          });
+          tooltipData.withoutEmail_label = intl.formatMessage({
+            id: 'reports_daily_visits.tooltip_without_email',
+          });
+
+          const valueTemplate = (value) => {
+            return `<div class='tooltip-value'>${value}</div>`;
+          };
+
+          return `
+          <div class="c3-tooltip">
+            <div class='tooltip-title'>${date}</div>
+            ${
+              tooltipData.withEmail === null
+                ? valueTemplate(tooltipData.quantity_label + tooltipData.quantity)
+                : ''
+            }
+            ${
+              tooltipData.withEmail !== null
+                ? valueTemplate(tooltipData.withEmail_label + tooltipData.withEmail) +
+                  valueTemplate(tooltipData.withoutEmail_label + tooltipData.withoutEmail)
+                : ''
+            }
+          </div>`;
+        }
+        return '';
       },
     },
     axis: {
--- a/src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
+++ b/src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
@@ -28,12 +28,27 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
             weekday: 'long',
           });
         },
-        value: (value) => {
-          return (
-            intl.formatMessage({
-              id: 'reports_daily_visits.tooltip_page_views',
-            }) + value
-          );
+        value: (value, ratio, id) => {
+          switch (id) {
+            case 'quantity':
+              return (
+                intl.formatMessage({
+                  id: 'reports_daily_visits.tooltip_page_views',
+                }) + value
+              );
+            case 'withEmail':
+              return (
+                intl.formatMessage({
+                  id: 'reports_daily_visits.tooltip_with_email',
+                }) + value
+              );
+            case 'withoutEmail':
+              return (
+                intl.formatMessage({
+                  id: 'reports_daily_visits.tooltip_without_email',
+                }) + value
+              );
+          }
         },
       },
     },
@@ -78,7 +93,11 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
             json: dailyVisitsData.value,
             keys: {
               x: 'from',
-              value: ['quantity'],
+              value: ['quantity', 'withEmail', 'withoutEmail'],
+            },
+            classes: {
+              withEmail: 'hide-graph',
+              withoutEmail: 'hide-graph',
             },
           },
         });
--- a/src/services/datahub-client.ts
+++ b/src/services/datahub-client.ts
@@ -231,7 +231,7 @@ export class HttpDatahubClient implements DatahubClient {
         to: new Date(x.to),
         quantity: x.quantity,
         withEmail: x.withEmail,
-        withoutEmail: x.quantity - x.withEmail,
+        withoutEmail: x.withEmail !== undefined ? x.quantity - x.withEmail : x.withEmail,
       }));
 
       return {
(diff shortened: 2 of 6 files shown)


src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -18,23 +18,71 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
       show: false,
     },
     tooltip: {
-      format: {
-        title: (date) => {
-          return intl.formatDate(date, {
+      contents: function(data) {
```
fcoronelmakingsense: maybe can separate in other file or not? because it's too big the configuration

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -18,23 +18,71 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
...
+                        id: 'reports_daily_visits.tooltip_without_email',
+                      }) + item.value
+                    : '';
+                break;
+              default:
+                return '';
+            }
+          }
+          return `
+          <table class="c3-tooltip">
```
fcoronelmakingsense: Why use table?

src/services/datahub-client.ts
```diff
@@ -230,6 +230,8 @@ export class HttpDatahubClient implements DatahubClient {
         from: new Date(x.from),
         to: new Date(x.to),
         quantity: x.quantity,
+        withEmail: x.withEmail,
```
fcoronelmakingsense: what happen if withEmail is undefined or null?

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -18,23 +18,71 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
...
-          return (
-            intl.formatMessage({
-              id: 'reports_daily_visits.tooltip_page_views',
-            }) + value
-          );
-        },
+          let quantity = '',
+            withEmail = '',
+            withoutEmail = '';
+          for (const item of data) {
```
andresmoschini: This structure is difficult to follow. 

What is Data?

If it is an array, what do you think about using _reduce_ as @fcoronelmakingsense did?

What happens if there are more than a _quantity_ item or less than one?

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -18,23 +18,71 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
...
+          }
+          return `
+          <table class="c3-tooltip">
+            <tbody>
+              <tr>
+                <th colspan="2">${date}</th>
+              </tr>
+              <tr>
+                <td colspan="2">
+                ${withEmail === '' ? quantity : ''}
```
andresmoschini: What do you think about

```suggestion
                ${withEmail || quantity}
```

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
cbernat: There will always be three, this is associated to the columns defined in ln 129. 
```
 keys: {
              x: 'from',	              x: 'from',
              value: ['quantity'],	              value: ['quantity', 'withEmail', 'withoutEmail'],
            },
```

cbernat: It will try to use reduce here.

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
cbernat: Yes this configuration stands for title, and value in format section for tooltip.

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
cbernat: Because this is the structure already used by the library c3, some styles are tied to this. I could change them for this case, making a simple structure.

src/services/datahub-client.ts
cbernat: That's the idea. That this will be not defined, do you thinks it's best to set a value anyways?

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -18,23 +18,71 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
...
+                </td>
+              </tr>
+              <tr>
+                <td colspan="2">
+                ${withEmail}
+                </td>
+              </tr>
+              <tr>
+                <td colspan="2">
+                ${withoutEmail}
```
andresmoschini:
```suggestion
                <FormatMessage id="reports_daily_visits.tooltip_without_email" /><span>${withoutEmail}</span>
```

cbernat: But this html is not jsx O_o I can't do this here.

andresmoschini: oh!

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
andresmoschini: Uh, no estoy seguro de si esta opción es correcta, ya que va a usar _quantity_ incluso cuando _withEmail_ es cero, y probablemente eso esté bien.

src/services/datahub-client.ts
fcoronelmakingsense: how is the behavior when don't have value?

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
cbernat: Now it reduced its size a lot, perhaps it can be kept here.

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
cbernat: Condition validation changed.

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -18,23 +18,54 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
...
+                      }) + tooltipData.quantity
+                    : ''
+                }
+            </div>
+            <div class='tooltip-value'>
+                ${
+                  !!tooltipData.withEmail
+                    ? intl.formatMessage({
+                        id: 'reports_daily_visits.tooltip_with_email',
+                      }) + tooltipData.withEmail
```
andresmoschini: What do you think about adding the space here in place of the language file?
```suggestion
                      }) + ' ' + tooltipData.withEmail
```

I Feel that the spaces are error-prone: if change the layout we should also change the language file, that is not fine.

image

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -18,23 +18,54 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
...
-          return (
-            intl.formatMessage({
-              id: 'reports_daily_visits.tooltip_page_views',
-            }) + value
-          );
-        },
+
+          const tooltipData = data.reduce((accumulator, item) => {
+            accumulator[item.id] = item.value;
+            return accumulator;
```
andresmoschini: ❤️

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -18,23 +18,54 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
...
+            <div class='tooltip-title'>${date}</div>
+            <div class='tooltip-value'>
+                ${
+                  !tooltipData.withEmail
+                    ? intl.formatMessage({
+                        id: 'reports_daily_visits.tooltip_page_views',
+                      }) + tooltipData.quantity
+                    : ''
+                }
+            </div>
```
andresmoschini: Is it fine having this element empty?

```html
            <div class='tooltip-value'>
            </div>
```

Maybe the _div_ should not exist. What do you think about nesting HTML inside template entries or pre-define them before?

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
cbernat: I'm not so sure about this, take a look at the line above, we used the same criteria: leave a blank space in the text. I think we should be consistent in the way we handle this issues.

src/services/datahub-client.ts
cbernat: This value should be undefined, and it's handled in the component as in the other components.

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
andresmoschini: Ok, it is just that I would prefer being consistent in a different way, I think that if something is a label should not have space after and if it is a sentence it should not be concatenated. 

I think that it would be a better model of our intention. But, it is not necessary go deeper by the moment, it is not so important.

Thanks for the patience.

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -18,23 +18,56 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
...
+          tooltipData.withEmail_label = intl.formatMessage({
+            id: 'reports_daily_visits.tooltip_with_email',
+          });
+          tooltipData.withoutEmail_label = intl.formatMessage({
+            id: 'reports_daily_visits.tooltip_without_email',
+          });
+
+          const valueTemplate = (value) => {
+            return `<div class='tooltip-value'>${value}</div>`;
+          };
```
andresmoschini: Nice!

PR #430, opened 2019-09-19
fix: adjust margins error messages in traffic source

GusBaamonde:
messageError

### DIFF
--- a/src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
+++ b/src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
@@ -3,6 +3,7 @@ import { InjectAppServices } from '../../../services/pure-di';
 import { FormattedMessage, FormattedNumber } from 'react-intl';
 import Loading from '../../Loading/Loading';
 import * as S from './ReportsTrafficSources.styles';
+import { BoxMessage } from '../../styles/messages';
 
 const ReportsTrafficSources = function({ domainName, dateFrom, dependencies: { datahubClient } }) {
   const [state, setState] = useState({ loading: true });
@@ -127,11 +128,11 @@ const ReportsTrafficSources = function({ domainName, dateFrom, dependencies: { d
               </S.ListItem>
             ))
           ) : (
-            <div className="dp-msj-error bounceIn">
+            <BoxMessage className="dp-msj-error bounceIn">
               <p>
                 <FormattedMessage id="trafficSources.error" />
               </p>
-            </div>
+            </BoxMessage>
           )}
         </S.ListContainer>
       </div>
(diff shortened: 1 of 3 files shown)


src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
```diff
@@ -127,11 +128,11 @@ const ReportsTrafficSources = function({ domainName, dateFrom, dependencies: { d
               </S.ListItem>
             ))
           ) : (
-            <div className="dp-msj-error bounceIn">
+            <ContentError className="dp-msj-error bounceIn">
               <p>
                 <FormattedMessage id="trafficSources.error" />
               </p>
-            </div>
+            </ContentError>
```
andresmoschini: Should you use the `S.`convention here?

CC: @fcoronelmakingsense

src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
```diff
@@ -127,11 +128,11 @@ const ReportsTrafficSources = function({ domainName, dateFrom, dependencies: { d
               </S.ListItem>
             ))
           ) : (
-            <div className="dp-msj-error bounceIn">
+            <ContentError className="dp-msj-error bounceIn">
```
andresmoschini: These classes should be also part of the styled component, right?

src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
```diff
@@ -127,11 +128,11 @@ const ReportsTrafficSources = function({ domainName, dateFrom, dependencies: { d
               </S.ListItem>
             ))
           ) : (
-            <div className="dp-msj-error bounceIn">
+            <BoxMessage className="dp-msj-error bounceIn">
```
cbernat: This styled component should be named with prefix like `S.BoxMessage` I think we agreed on that some time ago.

fcoronelmakingsense: Same

src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
```diff
@@ -3,6 +3,7 @@ import { InjectAppServices } from '../../../services/pure-di';
 import { FormattedMessage, FormattedNumber } from 'react-intl';
 import Loading from '../../Loading/Loading';
 import * as S from './ReportsTrafficSources.styles';
+import { BoxMessage } from '../../styles/messages';
```
fcoronelmakingsense: import * as S from '../../styles/messages';

andresmoschini: There is already an `S`:

image

We have talked with @GusBaamonde that in the long term we need more reusable styles, for example, there should not be `ReportsTrafficSources.styles`, I mean styles specific for traffic sources.

PR #435, opened 2019-09-23
feat(datahub): add protocol if enabled

cbernat:
For page ranking data, add protocol if it is comming from datahub to complete full url.

### DIFF
--- a/src/services/datahub-client.ts
+++ b/src/services/datahub-client.ts
@@ -153,7 +153,7 @@ export class HttpDatahubClient implements DatahubClient {
 
       // By the moment we are hard-coding it because DataHub does not have this
       // information. I am looking you Leo :P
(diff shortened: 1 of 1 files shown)


src/services/datahub-client.ts
```diff
@@ -153,7 +153,7 @@ export class HttpDatahubClient implements DatahubClient {
 
       // By the moment we are hard-coding it because DataHub does not have this
       // information. I am looking you Leo :P
-      const urlSchema = 'http://';
+      const urlSchema = response.data.protocol ? response.data.protocol : 'http://';
```
andresmoschini: I prefer this syntax, but it is the same:

```suggestion
      const urlSchema = response.data.protocol || 'http://';
```

PR #439, opened 2019-09-23
fix: chart adjustment of reports daily visits

GusBaamonde:
Every time a new `key` is added (example: `witoutEmail`), c3 automatically generates a `class` with the new `key` (example: `c3-target-witoutEmail`) to which we will have to add the corresponding color of the data we want to paint.
The code example is in the image.
Design adjustments were made for a better visualization of the grouped data.
Design validated with @santiagopaolucci 

**Linear graph example**
example-01

**Example area chart**
example-02

**Example code**
c3

### DIFF
--- a/src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
+++ b/src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
@@ -81,7 +81,7 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
       pattern: ['#B58FC1'],
     },
     point: {
-      r: 5,
+      r: 3,
     },
     grid: {
       y: {
--- a/src/components/shared/C3Chart/C3Chart.styles.js
+++ b/src/components/shared/C3Chart/C3Chart.styles.js
@@ -27,9 +27,9 @@ export const C3ChartStyled = styled.div`
   }
 
   .c3-line {
-    stroke-width: 4px;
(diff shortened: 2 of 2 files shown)


santiagopaolucci:
It is cool!
For the moment, we do not use graphics to areas.
When we need to use these graphics we should check that the visualization is appropriate according to the scenario.
But it looks great!

src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -81,7 +81,7 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
       pattern: ['#B58FC1'],
     },
     point: {
-      r: 5,
+      r: 3,
```
cbernat: What is this change in parameter about?

GusBaamonde: It's the radius of the point on the graph, it just got smaller

andresmoschini:
Estoy viendo que las series WithEmail y WithoutEmail no se están apilando, creo que estaría bueno que se apilen ya que juntas reprerssentan el total de visitas. Entonces, abajo de todo se puede poner WithEmail y ensima la otra y el area total representaría la cantidad de visitas totales.

Por otro lado, podrás poner un ejemplo donde la cantidad de visitas con email sea mayor que la de visitas sin email?

PR #441, opened 2019-09-24
DW-7 - add routing without hash 

cbernat:
- Change `HashRouter` by `BrowserRouter` to allow urls without hash.

Issue in jira --> https://makingsense.atlassian.net/browse/DW-7

**Pending**
- [x]  Redirect or fix for urls with /#/ 
- [ ]  Server changes for this to work --> this will be applied outside this PR.
- [x] Make it work in cdn enviroments, it seems to be replacing all location pathname by hash.

CDN now works as before! https://cdn.fromdoppler.com/doppler-webapp/demo-build1675/#/integrations/shopify


src/index.js
```diff
@@ -4,7 +4,7 @@ import './index.css';
 import App from './App';
 import * as serviceWorker from './serviceWorker';
 import { AppServicesProvider } from './services/pure-di';
-import { HashRouter as Router } from 'react-router-dom';
+import { BrowserRouter as Router } from 'react-router-dom';
```
andresmoschini: 😂

andresmoschini: Nice! it was a simple change.

Did you validate if `#` based URLs do not work auto-magically?

cbernat: Yes, but sadly no, they all get redirected to `/`

src/App.js
```diff
@@ -88,9 +88,17 @@ class App extends Component {
...
           <Switch>
-            <Route path="/" exact>
-              <RedirectToLegacyUrl to="/Campaigns/Draft" />
-            </Route>
+            <Route
+              exact
+              path="/"
+              render={({ location }) =>
+                location.hash.length ? (
+                  <Redirect to={location.hash.replace('#', '')} />
```
fcoronelmakingsense: it's means all url with hash redirect to same url without hash right?

andresmoschini: What do you think about removing the first character in place of replacing all `#`?

All `hash` values start with `#`, right? So, it is cheaper and more general because it will keep other `#` occurrences (if there are more than one)

andresmoschini: Does it keep the page in history or skip it?

I do not know what option is better, but we should test it to ensure good user experience.

cbernat: It works locally and that's all, it needs further testing. I should check history too..

cbernat: History seems to be working great! 😄

cbernat: > What do you think about removing the first character in place of replacing all `#`?
> 
> All `hash` values start with `#`, right? So, it is cheaper and more general because it will keep other `#` occurrences (if there are more than one)

Only gets to `/` when the hash is in the route like this `http://localhost:3000/#/integrations/shopify` but if we have `http://localhost:3000/integrations/shopify#hash1` it never gets to `/` rule.
If whe have `http://localhost:3000/#/integrations/shopify#hash1` seems to only detect as hash the first hash and then redirects to `http://localhost:3000/integrations/shopify#hash1` without issues.

andresmoschini: Ohh, _replace_ replace in javascript only takes the first occurrence, right?

cbernat: If we were to test in cdn, redirects are working in a very weird fashion, `http://cdn.fromdoppler.com/doppler-webapp/build1606/#/login` redirects to `https://cdn.fromdoppler.com/reports` O_o

andresmoschini: It is because `/reports` is a kind of default page.

andresmoschini: Anyway, testing it in the CDN has no sense because we need changes in the HTTP server. But it generates a new issue for us: what will we do with our CDN after this change?

cbernat: Seems to be replacing pathname with location.hash, in local is the same, but in cdn url is not the same.

andresmoschini: Tiene que ser porque localmente el root path es `/`, pero en el CDN es algo como `/doppler-webapp/build76/`.

andresmoschini:
We have talked with @cbernat about this: https://github.com/FromDoppler/doppler-webapp/pull/441#discussion_r327788514

> Anyway, testing it in the CDN has no sense because we need changes in the HTTP server. But it generates a new issue for us: what will we do with our CDN after this change?

It seems that we cannot make our CDN working in the required way. Maybe we can use one router or another based on a build parameter.

So, in our CDN we would continue using `HashRouter` and in real environments, we would use `BrowserRouter`.

src/index.js
```diff
@@ -59,11 +60,15 @@ history.listen((location) => {
   trackNavigation(location);
 });
 
+// Choose hash router for cdn only
+const Router = process.env.REACT_APP_CDN_BUILD === 'true' ? CdnRouter : ServerRouter;
```
andresmoschini: What do you think about renaming this variable to the specific behavior, something like `REACT_APP_HASH_ROUTER`, the same for the router names: `BrowserRouter` and `HashRouter`.

andresmoschini: Or another option could be using a name in place of `'true'`, for example:

```suggestion
const Router = process.env.REACT_APP_ROUTER === 'hash' ? HashRouter : BrowserRouter;
```

cbernat: Yes I think you are about right, second option fits best.

PR #448, opened 2019-09-25
refactor(ForgotPassword): apply styled components

fcoronelmakingsense:
- Remove forgot password style-sheet
- Add styled component file and create one component for the message and add the animation.


src/components/ForgotPassword/ForgotPassword.styles.js
```diff
@@ -0,0 +1,53 @@
+import styled, { keyframes } from 'styled-components';
+import colors from '../styles/colors';
+
+const bouncedIn = keyframes`
```
GusBaamonde: Maybe we should have the animation in a separate file, since the same animation is used on other screens.
This way we could reuse it.

fcoronelmakingsense: Ok thanks!

fcoronelmakingsense: ready

PR #452, opened 2019-09-26
fix(ReportsFilters): add link when domain is non verified

fcoronelmakingsense:
- Add link to sitetracking when domain is non verified
- Fix styles to can show propertly with `FormattedMessageMarkdown`

### DIFF
--- a/src/i18n/es.js
+++ b/src/i18n/es.js
@@ -167,7 +167,7 @@ export default {
       </p>
       `,
     domain: `Dominio`,
-    domain_not_verified: `Sin visitas registradas`,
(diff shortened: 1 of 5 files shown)


src/i18n/es.js
```diff
@@ -167,7 +167,7 @@ export default {
       </p>
       `,
     domain: `Dominio`,
-    domain_not_verified: `Sin visitas registradas`,
+    domain_not_verified: `Sin visitas registradas [Verificar dominio](${urlSiteTracking}).`,
```
andresmoschini: Remember the suffix.

```suggestion
    domain_not_verified_MD: `Sin visitas registradas [Verificar dominio](${urlSiteTracking}).`,
```

fcoronelmakingsense: fixed

PR #467, opened 2019-10-03
feat(datahub-client): refactor to add pagination in page ranking

fcoronelmakingsense:
- Refactor method to can add pagination. Add page size and page number
- Add test for Ranking pages by period
- Refactor double for local

### DIFF
--- a/src/services/datahub-client.doubles.ts
+++ b/src/services/datahub-client.doubles.ts
@@ -160,11 +160,15 @@ export class HardcodedDatahubClient implements DatahubClient {
   public async getPagesRankingByPeriod({
     domainName,
     dateFrom,
+    pageSize,
+    pageNumber,
   }: {
-    domainName: number;
+    domainName: string;
     dateFrom: Date;
+    pageSize: number;
+    pageNumber: number;
   }): Promise<PageRankingResult> {
-    console.log('getPagesRankingByPeriod', { domainName, dateFrom });
+    console.log('getPagesRankingByPeriod', { domainName, dateFrom, pageSize, pageNumber });
     await timeout(1500);
     const pages = fakePagesData.map((x) => ({
       name: x.name,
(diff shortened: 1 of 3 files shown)


src/services/datahub-client.doubles.ts
```diff
@@ -173,9 +177,19 @@ export class HardcodedDatahubClient implements DatahubClient {
       withEmail: x.withEmail,
     }));
 
+    let pagesSubArray = [];
+
+    if (pageSize) {
+      const indexStart = pageSize * pageNumber - 1;
```
andresmoschini: Is it fine? If not, could you add a test to take into account this?

src/services/datahub-client.ts
```diff
@@ -140,16 +142,22 @@ export class HttpDatahubClient implements DatahubClient {
...
+    pageNumber: number;
   }): Promise<PageRankingResult> {
     try {
       const response = await this.customerGet<any>(
         `domains/${domainName}/events/summarized-by-page`,
         {
           startDate: dateFrom.toISOString(),
           sortBy: 'visitors',
+          pageSize: pageSize || '',
+          pageNumber: pageNumber || '',
```
andresmoschini: These values should not be numbers?

```suggestion
          pageNumber: pageNumber || 0,
```

Or

```suggestion
          pageNumber: pageNumber || null,
```

andresmoschini:
Nice!, does it looks fine even when DataHub does not have pagination option ready?

src/services/datahub-client.doubles.ts
fcoronelmakingsense: it's only for local purposes the real BE it's no like this

fcoronelmakingsense:
> Nice!, does it looks fine even when DataHub does not have pagination option ready?

Yes, now dont have pagination and works fine

src/services/datahub-client.doubles.ts
andresmoschini: Oh, I understand, but I have two observations:

* Please verify this line, because, if I am not wrong it should be:

    ```suggestion
          const indexStart = pageSize * pageNumber - 1;
    ```

* I have checked [the related demo page](https://cdn.fromdoppler.com/doppler-webapp/demo-build1658/#/reports), and I cannot see anything related to the pagination. Maybe is this code not necessary?

fcoronelmakingsense: Thanks i changed the parenthesis

PR #468, opened 2019-10-03
fix(ReportsTrafficSources): when don't have quantity

fcoronelmakingsense:
- Show empty data when don't have values

image

### DIFF
--- a/src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
+++ b/src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
@@ -5,6 +5,10 @@ import { Loading } from '../../Loading/Loading';
 import * as S from './ReportsTrafficSources.styles';
 import { BoxMessage } from '../../styles/messages';
 
+const SafeDivide = (number, quantity) => {
+  return quantity ? number / quantity : 0;
+};
+
 const ReportsTrafficSources = function({ domainName, dateFrom, dependencies: { datahubClient } }) {
   const [state, setState] = useState({ loading: true });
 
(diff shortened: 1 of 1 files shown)


src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
```diff
@@ -90,7 +94,9 @@ const ReportsTrafficSources = function({ domainName, dateFrom, dependencies: { d
                           <S.Bar
                             primary
                             style={{
-                              width: (trafficSource.withEmail / trafficSource.quantity) * 100 + '%',
+                              width: trafficSource.quantity
+                                ? (trafficSource.withEmail / trafficSource.quantity) * 100 + '%'
```
cbernat: I've noticed you have always a division, how about making a function for example `SafeDivide(trafficSource.withEmail,trafficSource.quantity )` that returns the division result or zero if the divisor is zero? Then you can reuse and avoid having an inline if in the html. What do you think?

fcoronelmakingsense: Ready! good catch. thanks!

src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
```diff
@@ -5,6 +5,10 @@ import { Loading } from '../../Loading/Loading';
 import * as S from './ReportsTrafficSources.styles';
 import { BoxMessage } from '../../styles/messages';
 
+const SafeDivide = (number, quantity) => {
```
cbernat: ❤️

cbernat:
🚀

src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.js
```diff
@@ -5,6 +5,10 @@ import { Loading } from '../../Loading/Loading';
...
+  return quantity ? number / quantity : 0;
+};
```
andresmoschini: Nice! since this is a very simple function you can rewrite it in this way:

```js
const SafeDivide = (number, quantity) => quantity ? number / quantity : 0;
```

PR #481, opened 2019-10-08
fix(ReportsDailyVisits): fix width style in the graphic

fcoronelmakingsense:
image

### DIFF
--- a/src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
+++ b/src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
@@ -129,19 +129,20 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
         <small className="title-reports-box">
           <FormattedMessage id="reports_daily_visits.title" />
         </small>
-        <S.ContentContainer>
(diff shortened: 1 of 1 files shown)


src/components/Reports/ReportsDailyVisits/ReportsDailyVisits.js
```diff
@@ -129,19 +129,20 @@ const ReportsDailyVisits = ({ domainName, dateFrom, dependencies: { datahubClien
...
                 <FormattedMessage id="trafficSources.error" />
               </p>
             </BoxMessage>
-          ) : (
-            <C3Chart config={chartConfig} dataOptions={chartDataOptions} data={state.chartData} />
-          )}
-        </S.ContentContainer>
+          </S.ContentContainer>
+        ) : (
+          <C3Chart config={chartConfig} dataOptions={chartDataOptions} data={state.chartData} />
```
andresmoschini: Just to understand, what is the rationale for removing the container in this chart?

fcoronelmakingsense: the container it's only for the box message. The chart has your container

PR #482, opened 2019-10-08
fix: adjust width of week chart and hours

GusBaamonde:
- Hover effect adjustment on each circle

week-hours-100

### DIFF
--- a/src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.styles.js
+++ b/src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.styles.js
@@ -5,7 +5,6 @@ import fonts from '../../styles/fonts';
 
 export const WrapperBoxContainer = styled.div`
   .reports-box {
-    width: 900px;
     overflow: visible;
   }
 `;
@@ -53,7 +52,7 @@ export const Row = styled.div`
 `;
 
 export const Column = styled.div`
-  width: 31px;
+  width: 46px;
   height: 31px;
   position: relative;
   display: flex;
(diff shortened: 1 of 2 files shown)


src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.styles.js
```diff
@@ -68,7 +67,7 @@ export const Circle = styled.span`
   width: 11px;
   height: 11px;
   background-color: ${colors.purple1};
-  box-shadow: 0p 0p 0 30px rgba(255, 255, 255, 0);
+  box-shadow: 0px 0px 0 30px rgba(255, 255, 255, 0);
```
andresmoschini: Maybe:

```suggestion
  box-shadow: 0 0 0 30px rgba(255, 255, 255, 0);
```

src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.styles.js
```diff
@@ -86,19 +85,19 @@ export const Circle = styled.span`
       height: 31px;
     `}
 
-   &:hover {
+    &:hover {
```
andresmoschini: 👍

fcoronelmakingsense:
i have only a question when happens is resize the the screen? it's broken or fluid?

GusBaamonde:
> i have only a question when happens is resize the the screen? it's broken or fluid?

It is fluid up to 1024, but when we have the grid we will adjust what is missing

GusBaamonde:
test this please

PR #494, opened 2019-10-16
fix(reports): return full date with days included

cbernat:
image

Fix: https://cdn.fromdoppler.com/doppler-webapp/build1717/#/reports

### DIFF
--- a/src/utils.test.js
+++ b/src/utils.test.js
@@ -1,5 +1,5 @@
 import 'jest';
-import { getDataHubParams, useInterval } from './utils';
+import { getDataHubParams, useInterval, addDays } from './utils';
 import { renderHook } from '@testing-library/react-hooks';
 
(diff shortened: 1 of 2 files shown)


src/utils.test.js
```diff
@@ -128,4 +128,16 @@ describe('utils', () => {
...
+    it('should return a valid date, without mutating the base date', () => {
+      // Arrange
+      const initialDate = new Date();
+      // Act
+      const resultDate = addDays(initialDate, 3);
+      // Assert
+      expect(resultDate.toISOString()).toBeDefined();
+      expect(initialDate).not.toEqual(resultDate);
+    });
+  });
```
andresmoschini: ❤️

cbernat: Done.

PR #499, opened 2019-10-18
feat(ReportsHoursVisits): add dynamic ranges values

fcoronelmakingsense:
image

image

### DIFF
--- a/src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.js
+++ b/src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.js
@@ -43,19 +43,27 @@ const ReportsHoursVisits = ({ domainName, dateFrom, dateTo, dependencies: { data
       if (!hoursVisitsdata.success) {
         setState({ loading: false });
       } else {
-        const visitsByWeekDayAndHour = hoursVisitsdata.value.reduce((accumulator, item) => {
-          const weekDay = item.from.getUTCDay();
-          const hour = item.from.getUTCHours();
-          accumulator[weekDay][hour].quantity += item.quantity;
-          if (item.withEmail || item.withEmail === 0) {
-            accumulator[weekDay][hour].withEmail += item.withEmail;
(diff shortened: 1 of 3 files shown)


cbernat:
👍

src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.js
```diff
@@ -51,11 +52,16 @@ const ReportsHoursVisits = ({ domainName, dateFrom, dateTo, dependencies: { data
             accumulator[weekDay][hour].withEmail += item.withEmail;
             accumulator[weekDay][hour].withoutEmail += item.withoutEmail;
           }
+          if (item.quantity > max) {
+            max = item.quantity;
```
andresmoschini: _reduce_ should not have side effects.

What do you think about adding a new property to the accumulator with the maximum value?

```js
        const visitsByWeekDayAndHour = hoursVisitsdata.value.reduce((accumulator, item) => {
          const weekDay = item.from.getUTCDay();
          const hour = item.from.getUTCHours();
          accumulator[weekDay][hour].quantity += item.quantity;
          accumulator.max = item.quantity > accumulator.max ? item.quantity : accumulator.max;
          // . . .
```

andresmoschini: The name `visitsByWeekDayAndHour` would be wrong, maybe rename it as `processedVisits` or something like that?

You can also change the structure, for example:

```
{
    byWeekDayAndHour: [
        // . . . 
    ], [
        // . . . 
    ], 
    // . ..
    max: 50
}
```

andresmoschini: Better example:

```js
const processedVisits = hoursVisitsdata.value.reduce((accumulator, item) => {
  // . . .
}, { byWeekDayAndHour: createEmptyWeekDayHoursMatrix(), max: 0 });
```

PR #503, opened 2019-10-22
refactor(Reports): change component class for function

fcoronelmakingsense:


src/components/Reports/Reports.js
```diff
@@ -15,54 +15,47 @@ import { Helmet } from 'react-helmet';
 import { Loading } from '../Loading/Loading';
 import { addDays } from '../../utils';
 
-class Reports extends React.Component {
-  /**
-   * @param { Object } props
-   * @param { import('../../services/pure-di').AppServices } props.dependencies
```
andresmoschini: Maybe you would like to keep these annotations, they are useful for IntelliSense.

There should be already examples of function components with these annotations.

src/components/Reports/Reports.js
```diff
@@ -15,54 +15,47 @@ import { Helmet } from 'react-helmet';
...
-   */
-  constructor({ dependencies: { datahubClient, appConfiguration } }) {
-    super();
+const periodSelectedDaysDefault = 7;
 
-    this.datahubClient = datahubClient;
-    this.appConfiguration = appConfiguration;
```
andresmoschini: ❤️

src/components/Reports/Reports.js
fcoronelmakingsense: ohh, thanks didn't know!

src/components/Reports/Reports.js
```diff
@@ -15,54 +15,47 @@ import { Helmet } from 'react-helmet';
...
-    const now = new Date();
-    const dateFrom = addDays(now, days * -1);
-    this.setState({ periodSelectedDays: days, dateFrom: dateFrom, dateTo: now });
-  };
+  useEffect(() => {
+    const fetchData = async () => {
+      const domains = await datahubClient.getAccountDomains();
+      if (domains) {
+        const now = new Date();
+        const dateFrom = addDays(now, periodSelectedDaysDefault * -1);
```
andresmoschini: What do you think about adding _periodSelectedDays_ as a dependency and update it declaratively?

cbernat:
💯

andresmoschini:
❤️

PR #504, opened 2019-10-22
DW-84 server conf no hash

cbernat:
Add nginx conf files for each enviroment, and copy into image, for deploy. 

- Add config files as they are now
- Add copy config files into nginx image
- remove homepage as `.`

Pending
- [x] Add QA and prod to use browser router, modifying conf files
- [ ] Add local enviroment script to use nginx locally for testing (I will add this in another PR)


andresmoschini:
❤️ 

> Pending
> 
>     * Add QA and prod to use browser router, modifying conf files
>     * Add local enviroment script to use nginx locally for testing

Are these tasks required for merging this? or they are future tasks?

Right now, the server startup script is overriding `site.conf` file, right?

cbernat:
> ❤️
> 
> > Pending
> > ```
> > * Add QA and prod to use browser router, modifying conf files
> > * Add local enviroment script to use nginx locally for testing
> > ```
> 
> Are these tasks required for merging this? or they are future tasks?
> 
> Right now, the server startup script is overriding `site.conf` file, right?

It's overriding int qa and prod. But since we are using a secret to create the service, this will be overwritten with the secret. I'm not so sure how to proceed from here.

andresmoschini:
> > Right now, the server startup script is overriding `site.conf` file, right?
> 
> It's overriding int qa and prod. But since we are using a secret to create the service, this will be overwritten with the secret. I'm not so sure how to proceed from here.

I think that you can merge it (nothing will happen), and then change the environment configuration, one by one.

andresmoschini:
> > Pending
> > ```
> > * Add QA and prod to use browser router, modifying conf files
> > * Add local enviroment script to use nginx locally for testing
> > ```
> 
> Are these tasks required for merging this? or they are future tasks?

I think that this question has not been answered yet.

andresmoschini:
https://cdn.fromdoppler.com/doppler-webapp/build1738/ is not working, is it expected?

cbernat:
> > > Pending
> > > ```
> > > * Add QA and prod to use browser router, modifying conf files
> > > * Add local enviroment script to use nginx locally for testing
> > > ```
> > 
> > 
> > Are these tasks required for merging this? or they are future tasks?
> 
> I think that this question has not been answered yet.

The change in config file for QA and prod is needed before merge. And then in order to use browser router, a change in env file for each enviroment needs to be done. I'm not sure how to handle env change. If I'm not being clear enogh perhaps we can discuss this in a call.

cbernat:
> https://cdn.fromdoppler.com/doppler-webapp/build1738/ is not working, is it expected?

Now it works, it's alive! https://cdn.fromdoppler.com/doppler-webapp/demo-build1742/#/integrations/shopify

Dockerfile.BUILDS_AND_CDN
```diff
@@ -15,6 +15,7 @@ RUN yarn verify
 # Build project
 ARG environment
 ENV REACT_APP_ROUTER=hash
+ENV PUBLIC_URL="."
```
andresmoschini: You should add this in all files where `REACT_APP_ROUTER=hash`, right?

andresmoschini: On the contrary, if you merge it, you will break the production environment.

andresmoschini: 669df940dc56294fc180661f719b82896e043b11 Oh, I understand, you want to merge it and apply the changes to PROD  all in the same step. 

I was understanding that the scope of this PR was only fixing INT.

cbernat: This file only has effect in cdn build, for releases we use `DockerFile.RELEASES`. In all the rest of the builds, url will be static, since we don't use '.' in homepage. This can be with or without router change. 
If I merge this PR the changes over config files will not be applied because on service creation we are using secrets. It should be replaced for current secret in production and QA.

cbernat:
@swidzinskiMS and @fgchaio could you please double check the conf files for nginx please? Just in case 😄

swidzinskiMS:
The .conf files are looking good Ceci!

fgchaio:
Looks good to me too. 💯

PR #512, opened 2019-10-24
fix(deploy): add enviroment to use in copy

cbernat:
Move arg closer to copy command to make it work.


cbernat:
I will merge this tomorrow, just in case.

andresmoschini:
👍  I guess

PR #513, opened 2019-10-24
fix(Reports): change dates to local time

fcoronelmakingsense:
- change all utc dates to local time
- change startDate to pass datahub date with 0 minutes and 0 seconds

### DIFF
--- a/src/components/Reports/Reports.js
+++ b/src/components/Reports/Reports.js
@@ -110,6 +110,7 @@ const Reports = ({ dependencies: { datahubClient } }) => {
               <ReportsTrafficSources
                 domainName={state.domainSelected.name}
                 dateFrom={state.dateFrom}
+                dateTo={state.dateTo}
               />
               <ReportsHoursVisits
                 domainName={state.domainSelected.name}
--- a/src/components/Reports/Reports.js
+++ b/src/components/Reports/Reports.js
@@ -17,8 +17,8 @@ import { addDays } from '../../utils';
 
 const periodSelectedDaysDefault = 7;
 
-const getCurrentDate = (date) => {
-  return new Date(date.getMonth() + 1 + '/' + date.getDate() + '/' + date.getFullYear());
+const getStartOfDate = (date) => {
+  return new Date(date.getFullYear(), date.getMonth() + 1, date.getDate());
 };
 
 /**
@@ -27,7 +27,7 @@ const getCurrentDate = (date) => {
  */
 
 const Reports = ({ dependencies: { datahubClient } }) => {
-  const today = getCurrentDate(new Date());
+  const today = getStartOfDate(new Date());
   const [state, setState] = useState({
     periodSelectedDays: periodSelectedDaysDefault,
     dateFrom: addDays(today, periodSelectedDaysDefault * -1),
@@ -40,7 +40,7 @@ const Reports = ({ dependencies: { datahubClient } }) => {
   };
 
   const changePeriod = (days) => {
-    const today = getCurrentDate(new Date());
+    const today = getStartOfDate(new Date());
     const dateFrom = addDays(today, days * -1);
     setState((prevState) => ({
       ...prevState,
--- a/src/components/Reports/Reports.js
+++ b/src/components/Reports/Reports.js
@@ -17,17 +17,21 @@ import { addDays } from '../../utils';
 
 const periodSelectedDaysDefault = 7;
 
+const getCurrentDate = (date) => {
+  return new Date(date.getMonth() + 1 + '/' + date.getDate() + '/' + date.getFullYear());
+};
+
 /**
  * @param { Object } props
  * @param { import('../../services/pure-di').AppServices } props.dependencies
  */
 
 const Reports = ({ dependencies: { datahubClient } }) => {
-  const now = new Date();
+  const dateNow = getCurrentDate(new Date());
   const [state, setState] = useState({
     periodSelectedDays: periodSelectedDaysDefault,
-    dateFrom: addDays(now, periodSelectedDaysDefault * -1),
-    dateTo: now,
+    dateFrom: addDays(dateNow, periodSelectedDaysDefault * -1),
+    dateTo: dateNow,
   });
 
   const changeDomain = async (name) => {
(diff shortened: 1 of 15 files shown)


src/components/Reports/Reports.js
```diff
@@ -17,17 +17,28 @@ import { addDays } from '../../utils';
...
 
+const getDateNow = (intl) => {
+  return new Date(
+    intl.formatDate(Date.now(), {
+      year: 'numeric',
+      month: 'numeric',
+      day: 'numeric',
+    }),
+  );
+};
```
andresmoschini: I do not understand what is the meaning of this function?

src/components/Reports/ReportsBox/ReportsBox.js
```diff
@@ -87,9 +87,8 @@ class ReportsBox extends React.Component {
                 <FormattedMessage id="reports_box.visits_with_email" />
               </h6>
               <small className="date-range">
-                <FormattedDate value={dateFrom} timeZone="UTC" />{' '}
-                <FormattedMessage id="reports_box.to" />{' '}
-                <FormattedDate value={dateTo} timeZone="UTC" />
+                <FormattedDate value={dateFrom} /> <FormattedMessage id="reports_box.to" />{' '}
```
andresmoschini: It is out of the scope of this PR, but I would like to comment it for the future, there should not be a text entry for `to`, there should be something like `{dateFrom} to {dateTo}`.

andresmoschini:
Are there not affected tests in this PR?

src/components/Reports/Reports.js
fcoronelmakingsense: create the date only using year month and day with 0 minutes and 0 seconds

andresmoschini:
I think that we are not paying attention to `dateTo` in the service. For example https://github.com/FromDoppler/doppler-webapp/blob/master/src/services/datahub-client.ts#L187-L193

src/components/Reports/Reports.js
andresmoschini: There should be a better way without `intl` (because it has nothing to do with localization).

Also, this method could be along with `addDays`, maybe it could be called `beginingOfDay`. For example:

```js
var today = beginingOfDay(new Date());
var yesterday = addDays(today, -1);
```

fcoronelmakingsense:
> I think that we are not paying attention to `dateTo` in the service. For example https://github.com/FromDoppler/doppler-webapp/blob/master/src/services/datahub-client.ts#L187-L193

Yes because datahub not need, is optional. But we can pass to them

src/components/Reports/Reports.js
```diff
@@ -17,17 +17,28 @@ import { addDays } from '../../utils';
 
 const periodSelectedDaysDefault = 7;
 
+const getDateNow = (intl) => {
+  return new Date(
+    intl.formatDate(Date.now(), {
```
andresmoschini: We have talked about removing _intl_ dependency in this function.

Also about renaming the function to `getCurrentDate` or convert this to something like `extractPureDate(date)` or `getBeginingOfDay(date)`.

andresmoschini:
>> I think that we are not paying attention to `dateTo` in the service.
> Yes because datahub not need, is optional. But we can pass to them

Regarding this, we was talking with @fcoronelmakingsense, we should probably discuss about the expected behavior with @mfantinims

src/components/Reports/Reports.js
fcoronelmakingsense: i changed

src/components/Reports/Reports.js
```diff
@@ -17,17 +17,21 @@ import { addDays } from '../../utils';
 
 const periodSelectedDaysDefault = 7;
 
+const getCurrentDate = (date) => {
+  return new Date(date.getMonth() + 1 + '/' + date.getDate() + '/' + date.getFullYear());
```
andresmoschini: There should be a way of doing this without using string, let me research...

andresmoschini:
```suggestion
  return new Date(date.getFullYear(), date.getMonth(), date.getDate());
```

src/components/Reports/Reports.js
```diff
@@ -17,17 +17,21 @@ import { addDays } from '../../utils';
 
 const periodSelectedDaysDefault = 7;
 
+const getCurrentDate = (date) => {
```
andresmoschini: In this case, _current_ is wrong, because you are getting the _pure date_ or _begining of the date_ passed as parameter. Let me research for common names for this operation.

andresmoschini: In _Moment.js_ they use _[startOf](https://momentjs.com/docs/#/manipulating/start-of/)_, what do you think about using the same convention than they?

```suggestion
const getStartOfDate = (date) => {
```

src/components/Reports/Reports.js
```diff
@@ -17,17 +17,21 @@ import { addDays } from '../../utils';
...
+};
+
 /**
  * @param { Object } props
  * @param { import('../../services/pure-di').AppServices } props.dependencies
  */
 
 const Reports = ({ dependencies: { datahubClient } }) => {
-  const now = new Date();
+  const dateNow = getCurrentDate(new Date());
```
andresmoschini: I think that _now_ is not a good name. Maybe _today_ or _startOfToday_:

```suggestion
  const startOfToday = getCurrentDate(new Date());
```

```suggestion
  const today = getCurrentDate(new Date());
```

cbernat:
👍

andresmoschini:
@fcoronelmakingsense @mfantinims 

> We talk with @mfantinims and decided to use the last 7 days excluded the current date. We think for this case can add a option in period filter like "today" and show in that case the current date.

En base a eso, y a las consultas que se hacen a DataHub, creo que se están mostrando mal las fechas:

image

fcoronelmakingsense:
Fix warning in ReportsHoursVisits.test 

image

src/i18n/en.js
```diff
@@ -150,6 +150,7 @@ export default {
     visits_description_without_emails: `Total number of users who visited your Website and whose Email has not been identified by Doppler. If a user entered several times, only one will be counted.`,
     visits_with_email: `Users with Email`,
     visits_without_emails: `Users without Email`,
+    without_included: `(without included)`,
```
andresmoschini: Maybe `not included` is better.

fcoronelmakingsense: fixed

src/components/Reports/Reports.js
fcoronelmakingsense: changed

src/components/Reports/Reports.js
fcoronelmakingsense: changed

PR #526, opened 2019-10-30
fix: change default router in env, remove in other env

cbernat:
This allows to overwrite in docker file later, for cdn enviroments.

Issue: DW-92


.env
```diff
@@ -12,4 +12,4 @@ REACT_APP_USE_DOPPLER_LEGACY_LOGIN=false
 REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 NODE_ICU_DATA=node_modules/full-icu
-REACT_APP_ROUTER=hash
+REACT_APP_ROUTER=browser
```
andresmoschini: Me parece que esto va a afectar a todos los entornos.

andresmoschini: 👍, Joya entonces.

PR #527, opened 2019-10-30
fix: change name of prop state by status to avoid errors

cbernat:


andresmoschini:
👍 

Are there not more places using `state`?

What was the issue? `state` is wrong and for any reason works wit _hash router_ but not with _browsere router_?

cbernat:
> 👍
> 
> Are there not more places using `state`?
> 
> What was the issue? `state` is wrong and for any reason works wit _hash router_ but not with _browsere router_?
This is the error in local env
image

Yes it tries to replace the state and fails, I've seen that in SignUpConfirmation component, state is a prop, not a state. The reason why it fails, I'm not so sure.

PR #537, opened 2019-10-31
fix: return the right month in getStartOfDate

andresmoschini:
Fix this:

image

Notice that the month is wrong. More details: https://makingsense.slack.com/archives/CGJ72QFQX/p1572470224155300

I will merge without review.

### DIFF
--- a/src/components/Reports/Reports.js
+++ b/src/components/Reports/Reports.js
@@ -18,7 +18,7 @@ import { addDays } from '../../utils';
 const periodSelectedDaysDefault = 7;
 
 const getStartOfDate = (date) => {
-  return new Date(date.getFullYear(), date.getMonth() + 1, date.getDate());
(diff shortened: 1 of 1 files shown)


andresmoschini:
Tested (minimally) in https://cdn.fromdoppler.com/doppler-webapp/demo-build1787/#/reports

src/components/Reports/Reports.js
```diff
@@ -18,7 +18,7 @@ import { addDays } from '../../utils';
 const periodSelectedDaysDefault = 7;
 
 const getStartOfDate = (date) => {
-  return new Date(date.getFullYear(), date.getMonth() + 1, date.getDate());
+  return new Date(date.getFullYear(), date.getMonth(), date.getDate());
```
cbernat: 👍

fcoronelmakingsense:
Thanks very much for the fix @andresmoschini

PR #542, opened 2019-10-31
refactor(utils): add test to check start of day function

fcoronelmakingsense:


src/utils.test.js
```diff
@@ -149,4 +149,45 @@ describe('utils', () => {
...
+  describe('getStartOfDate function', () => {
+    it('should return a valid date', () => {
+      // Arrange
+      const initialDate = new Date();
+
+      // Act
+      const resultDate = getStartOfDate(initialDate);
+
+      // Assert
+      expect(resultDate.toISOString()).toBeDefined();
```
andresmoschini: You can add an assert here to verify if the result date has 0 in hour, minute, second, and ms.

andresmoschini: Oh, you already did it: https://github.com/FromDoppler/doppler-webapp/pull/542/files#diff-38c429a60240c92cfd160bcc46f1a81bR177

Sorry.

Removing the comment.

src/utils.test.js
```diff
@@ -149,4 +149,45 @@ describe('utils', () => {
...
+    it('should return start of date with 0 minutes and 0 seconds', () => {
+      // Arrange
+      const initialDate = new Date();
+
+      // Act
+      const resultDate = getStartOfDate(initialDate);
+
+      // Assert
+      expect(resultDate.toISOString()).toBeDefined();
+      expect(resultDate.getMinutes()).toEqual(0);
```
andresmoschini: You can also add milliseconds.

src/utils.ts
```diff
@@ -76,3 +76,9 @@ export function addDays(date: Date, days: number) {
   newDate.setDate(date.getDate() + days);
   return newDate;
 }
+
+export function getStartOfDate(date: Date) {
+  return typeof date.getMonth === 'function'
+    ? new Date(date.getFullYear(), date.getMonth(), date.getDate())
+    : '';
```
andresmoschini: Mmm, I am not sure about it. 

I would prefer returning `undefined`, `null` or throwing an exception.

src/utils.test.js
```diff
@@ -149,4 +149,45 @@ describe('utils', () => {
...
+    it('should return empty value when pass type different to Date', () => {
+      // Arrange
+      const initialDate = '20-13-2018';
+
+      // Act
+      const resultDate = getStartOfDate(initialDate);
+
+      // Assert
+      expect(resultDate).toBeDefined();
+      expect(resultDate).toEqual('');
```
andresmoschini: I do not like this behavior, what is the rationale behind this?

src/utils.ts
fcoronelmakingsense: fixed, i return and undefined

fcoronelmakingsense:
> Maybe you can add another test to ensure that the new date has the same year, month, and day as the first one (what the issue was).
> 
> You can do it using arbitrary hardcoded input date and checking the result against a hardcoded value.

I already testing this. i put now a harcoded date to more specific

andresmoschini:
> I already testing this. i put now a harcoded date to more specific

You are right, thanks!

PR #543, opened 2019-10-31
feat: add grid system in traffic sources

GusBaamonde:
grid-system-in-traffic-sources

### DIFF
--- a/src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.styles.js
+++ b/src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.styles.js
@@ -23,20 +23,27 @@ export const ContentContainer = styled.div`
 
 export const ListContainer = styled.div`
   display: flex;
-  flex-flow: wrap;
+  flex-wrap: wrap;
   & > div:nth-child(odd) {
     background: ${colors.smoothGrey};
   }
   & > div:nth-child(n + 4) {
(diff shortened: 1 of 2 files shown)


src/components/Reports/ReportsTrafficSources/ReportsTrafficSources.styles.js
```diff
@@ -23,20 +23,27 @@ export const ContentContainer = styled.div`
...
-  flex-flow: wrap;
+  flex-wrap: wrap;
   & > div:nth-child(odd) {
     background: ${colors.smoothGrey};
   }
   & > div:nth-child(n + 4) {
-    margin-top: ${spacings.spacesLvl2};
+    margin-top: ${spacings.spacesLvl0};
+    @media (min-width: 768px) {
+      margin-top: ${spacings.spacesLvl2};
```
cbernat: Why do you need this here? I thought with just the grid it should adapt right?

GusBaamonde: That margin is the intermediate and necessary space to read the data fluently, according to what Santi told me.
But in mobile it does not work and can be removed.
The grid adapts the boxes is true, but there are certain things that we will have to touch them in the elements.

GusBaamonde:
test this please

PR #544, opened 2019-11-01
Dw 101 upgradepopup

cbernat:
For monthly subscriber or demo enviroment, click on buy and the popup did not show.

image

image

- Removed hook for intl
- Add test to validate correct render

### DIFF
--- a/src/components/UpgradePlanForm/UpgradePlanForm.js
+++ b/src/components/UpgradePlanForm/UpgradePlanForm.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { FormattedMessage, useIntl } from 'react-intl';
+import { FormattedMessage, injectIntl } from 'react-intl';
 import { Formik, Form, Field, ErrorMessage } from 'formik';
 import { InjectAppServices } from '../../services/pure-di';
 import { Loading } from '../../components/Loading/Loading';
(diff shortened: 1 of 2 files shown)


src/components/UpgradePlanForm/UpgradePlanForm.js
```diff
@@ -11,12 +11,12 @@ const fieldNames = {
...
 
 class UpgradePlanForm extends React.Component {
-  constructor({ dependencies: { dopplerLegacyClient } }) {
+  constructor({ dependencies: { dopplerLegacyClient }, intl }) {
     super();
 
     /** @type { import('../../services/doppler-legacy-client').DopplerLegacyClient } */
     this.dopplerLegacyClient = dopplerLegacyClient;
-    this.intl = useIntl();
+    this.intl = intl;
```
fcoronelmakingsense: why can't use hook?

src/components/UpgradePlanForm/UpgradePlanForm.test.js
```diff
@@ -0,0 +1,57 @@
...
+import '@testing-library/jest-dom/extend-expect';
+import IntlProvider from '../../i18n/DopplerIntlProvider.double-with-ids-as-values';
+import UpgradePlanForm from './UpgradePlanForm';
+import { AppServicesProvider } from '../../services/pure-di';
+
+describe('Upgrade plan form component', () => {
+  afterEach(cleanup);
+
+  it('renders upgrade popup for monthly subscribers without errors', async () => {
+    const dopplerLegacyClientDouble = {
```
fcoronelmakingsense: You can add `Arrange, Act, Assert` comments for more clear

src/components/UpgradePlanForm/UpgradePlanForm.js
andresmoschini: I have the same question, what is the rationale behind fixing the bug removing the hook?

cbernat: We cannot use hooks inside a class component, this was the error. It breaks the code.

cbernat: https://reactjs.org/docs/hooks-faq.html#should-i-use-hooks-classes-or-a-mix-of-both

PR #556, opened 2019-11-06
fix(unique page views): removed unnecessary classes

GusBaamonde:
fix-unique-pages-views

### DIFF
--- a/src/components/Reports/Reports.js
+++ b/src/components/Reports/Reports.js
@@ -16,7 +16,7 @@ import { Loading } from '../Loading/Loading';
 import { addDays, getStartOfDate } from '../../utils';
 
 // This value means the today date
-const periodSelectedDaysDefault = 0;
+const periodSelectedDaysDefault = 7;
 
 /**
  * @param { Object } props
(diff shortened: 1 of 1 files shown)


src/components/Reports/Reports.js
```diff
@@ -16,7 +16,7 @@ import { Loading } from '../Loading/Loading';
 import { addDays, getStartOfDate } from '../../utils';
 
 // This value means the today date
-const periodSelectedDaysDefault = 0;
+const periodSelectedDaysDefault = 7;
```
cbernat: With this the default is one week, is it intended?

cbernat:
subarashii 素晴しい ❤️

src/components/Reports/Reports.js
cbernat: @GusBaamonde told me it's intended.

cbernat:
For testing here in prod https://cdn.fromdoppler.com/doppler-webapp/build1837/#/reports

src/components/Reports/Reports.js
fcoronelmakingsense: yes, maybe need change the comment

PR #558, opened 2019-11-06
fix(test): remove warning for side effect null in react dom

cbernat:
Some of the lengthy console warning that we are getting in tests log, are from third party libraries that we use, that haven't updated ther dependencies, this is why yarn allows a configuration in `package.json` we could update dependencies for a particular library. This obviously is temporal. https://yarnpkg.com/lang/en/docs/selective-version-resolutions/  

In tests for: shopify.test ,signup.test, Reports.test, App.test
```
● Console

    console.warn node_modules/react-dom/cjs/react-dom.development.js:12338
      Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.

      * Move code with side effects to componentDidMount, and set initial state in the constructor.
      * Rename componentWillMount to UNSAFE_componentWillMount to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

      Please update the following components: SideEffect(NullComponent)
```

Pending 

- [x] add ^to version in resolution.
- [x] Check what we use in helmet is still funcional (title in pages, meta with description, zoho chat bubble logic, lang in html)

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,5 +108,9 @@
     "fail": [
       "@semantic-release/github"
     ]
+  },
+  "resolutions": {
+    "react-helmet/react-side-effect": ">=2.1.0",
+    "comment": "Resolution to avoid test warning: [...] Please update the following components: SideEffect(NullComponent)"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -5427,11 +5427,6 @@ execa@^1.0.0:
     signal-exit "^3.0.0"
     strip-eof "^1.0.0"
 
-exenv@^1.2.1:
-  version "1.2.2"
-  resolved "https://registry.yarnpkg.com/exenv/-/exenv-1.2.2.tgz#2ae78e85d9894158670b03d47bec1f03bd91bb9d"
-  integrity sha1-KueOhdmJQVhnCwPUe+wfA72Ru50=
-
 exit@^0.1.2:
   version "0.1.2"
   resolved "https://registry.yarnpkg.com/exit/-/exit-0.1.2.tgz#0632638f8d877cc82107d30a0fff1a17cba1cd0c"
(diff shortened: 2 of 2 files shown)


yarn.lock
```diff
@@ -11536,13 +11531,10 @@ react-scripts@^3.2.0:
...
 
-react-side-effect@^1.1.0:
-  version "1.1.5"
-  resolved "https://registry.yarnpkg.com/react-side-effect/-/react-side-effect-1.1.5.tgz#f26059e50ed9c626d91d661b9f3c8bb38cd0ff2d"
-  integrity sha512-Z2ZJE4p/jIfvUpiUMRydEVpQRf2f8GMHczT6qLcARmX7QRb28JDBTpnM2g/i5y/p7ZDEXYGHWg0RbhikE+hJRw==
-  dependencies:
-    exenv "^1.2.1"
-    shallowequal "^1.0.1"
+react-side-effect@2.1.0, react-side-effect@^1.1.0:
+  version "2.1.0"
```
andresmoschini: If I am understanding it fine, it says that we will use the version `2.1.0` even when the client asks for version `1.1.0`.

It could have breaking changes (major version number has changed), is it acceptable? do you know if it is only being used in tests?

cbernat: I will check, this was supposed to be only for react-helmet, not for all other packages.

cbernat: It's only used by `react-helmet` which we use to update html document properties like title and meta tags.

andresmoschini: The breaking change is related to the React version used, so, we are fine:

image

package.json
```diff
@@ -108,5 +108,8 @@
     "fail": [
       "@semantic-release/github"
     ]
+  },
+  "resolutions": {
+    "react-helmet/react-side-effect": "2.1.0"
```
fcoronelmakingsense: it's only in this library the problem?

cbernat: yes, I've searched through and found only helmet using it.

cbernat:
test this please.

package.json
```diff
@@ -108,5 +108,8 @@
     "fail": [
       "@semantic-release/github"
     ]
+  },
+  "resolutions": {
+    "react-helmet/react-side-effect": "^2.1.0"
```
andresmoschini: No le ponés un rango de versiones?

cbernat: Pero funciona eso? te entendi que me pasaste algo en beta O_o

andresmoschini:
```suggestion
    "react-helmet/react-side-effect@>=1.0.0 <2.1.0": "react-helmet/react-side-effect@^2.1.0"
```

cbernat: I've tried this and I get a warning:
```
yarn install v1.17.3
warning Resolution field "react-helmet/react-side-effect@^2.1.0" has an invalid version entry and may be ignored
``` 
So I'll leave this as is.

cbernat:
test this please.

package.json
andresmoschini: Did you try this?

```suggestion
    "react-helmet/react-side-effect@>=1.0.0 <2.1.0": "^2.1.0"
```

andresmoschini: Si no hacemos algo así, cuando `react-helmet/react-side-effect` se actualice a la versión 3, vamos a seguir con la `2.x.x`

andresmoschini: La otra opción es documentarnos mejor antes de hacer el cambio.

cbernat: Esto ultimo si que funciona, gracias <3

cbernat: Corri el test, recheckie el yarn lock y al final no funciona, sigo investigando. (en lo ultimo que subi se ve, en el lock)

cbernat:
Test this please.

cbernat:
@andresmoschini I've updated the resolution to allow taking more than the actual version, also I've added a comment in the form of a property. What do you think?

andresmoschini:
> @andresmoschini I've updated the resolution to allow taking more than the actual version, also I've added a comment in the form of a property. What do you think?

I Agree. If we cannot find a better solution after spend some time on this, I think that it is an acceptable solution.

PR #561, opened 2019-11-07
feat(Reports): change message domain non verified

fcoronelmakingsense:
- Remove message under of dropdown and put before reports boxes with other design
- Add test to check this state

image


src/components/Reports/ReportsFilters/ReportsFilters.js
```diff
@@ -50,9 +49,7 @@ const ReportsFilters = ({
                             <FormattedDate value={domainSelected.verified_date} />
                           </span>
                         </>
-                      ) : (
-                        <FormattedMessageMarkdown id="reports_filters.domain_not_verified_MD" />
-                      )}
+                      ) : null}
```
cbernat: Are you sure a null value here is ok?

fcoronelmakingsense: because the message now show before the boxes

cbernat:
🚀 I see this ok!

PR #571, opened 2019-11-12
fix(ReportsPageRanking.test): remove test because duplicated

fcoronelmakingsense:
- Remove the test to check page ranking without pages because we already have a test to test this scenario.
- When remove this fix 1 'act' warning 💃


cbernat:
👍 One warning less, thanks!

fcoronelmakingsense:
test this please

PR #572, opened 2019-11-12
chore: add jenkinsfile for testing

cbernat:
Add another jenkinsFile just for testing, this way I can change back and forth configuration without compromising working build. In jenkins it can be changed with jenkinsfile to use.

image

### DIFF
--- a/JenkinsfileTest
+++ b/JenkinsfileTest
@@ -1,0 +1,97 @@
+pipeline {
+    agent any
+    stages {
+        stage('Abort') {
+            input('Continue build?')
+        }
+        stage('Restore') {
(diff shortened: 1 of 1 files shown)


cbernat:
@andresmoschini since I didn't find a safe way to make my tests I've decided to add a file for tests purposes only. What do you think?

andresmoschini:
No entiendo bien como va a funcionar esto. 

Si necesitás hacer pruebas, tal vez directamente podés usar una rama en upstream con el Jenkinsfile modificado.

cbernat:
I will test this in a branch in upstream. Closing...

PR #576, opened 2019-11-13
refactor(Promotion):  remove unused functionality

fcoronelmakingsense:
- Remove disabled content because it's old and don't use anymore
- Refactor state like other components we have


cbernat:
👍

PR #578, opened 2019-11-13
Dw-117 jenkins pipeline conditional build for dependabot 

cbernat:
Added condition in jenkins pipeline, to prevent automatic build. 
If we would want to make a rebuild, we need to wait by blue view to accept the build (because it will timeout in 60 seconds window)
Else it will fail by timeout.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,17 @@
 pipeline {
     agent any
     stages {
+        stage('Confirmation') {
+            when {
+                beforeInput true
+                changeRequest author: 'dependabot-preview'
+            }
+            steps {
+                timeout(time: 180, unit: 'SECONDS') {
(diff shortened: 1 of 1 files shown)


cbernat:
image

Once timeout is finished, all stages are ignored, and build marked as failure. 🎉

andresmoschini:
Nice!

Jenkinsfile
```diff
@@ -1,6 +1,17 @@
 pipeline {
     agent any
     stages {
+        stage('Abort') {
```
andresmoschini:
```suggestion
        stage('Suspend') {
```

```suggestion
        stage('Pause') {
```

```suggestion
        stage('Confirmation') {
```

andresmoschini:
:shipit:

PR #582, opened 2019-11-15
fix: add grid system in unique page views

GusBaamonde:
- add grid system in unique page views

grid-system-page-views

- Add the error in dp-box-shadow

add-grid-system-and-error


andresmoschini:
Nide! does it also fix the issue with the spinners?

GusBaamonde:
> Nide! does it also fix the issue with the spinners?

In this case we don't have that problem. The problem is in the boxes above.
I will check it in another pr.

cbernat:
Very nice and clean! ❤️

PR #583, opened 2019-11-15
DW-114:chore: set default nginx version alpine

cbernat:
Set nginx to 1.16.1-alpine

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN yarn verify
 RUN yarn build
 
 # Host project in nginx
-FROM nginx
+FROM nginx:1.16-alpine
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN yarn verify
 RUN yarn build
 
 # Host project in nginx
-FROM nginx
+FROM nginx:1.16.1-alpine
```
andresmoschini: What do you think about using 1.16?

```suggestion
FROM nginx:1.16-alpine
```

In this way, if there as new patch version it will be updated.

andresmoschini:
:shipit:

PR #584, opened 2019-11-15
chore: add regex and log author data

cbernat:
- Add log for author in console (temporal for checking)
- Set regexp for author name in change request

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,10 +1,15 @@
 pipeline {
     agent any
     stages {
+        stage('Identification') {
+            steps {
+                sh 'echo Change request author: ${CHANGE_AUTHOR} '
+            }
+        }
         stage('Confirmation') {
             when {
                 beforeInput true
-                changeRequest author: 'dependabot-preview'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,10 +1,15 @@
...
+        stage('Identification') {
+            steps {
+                sh 'echo Change request author: ${CHANGE_AUTHOR} '
+            }
+        }
         stage('Confirmation') {
             when {
                 beforeInput true
-                changeRequest author: 'dependabot-preview'
+                changeRequest author: '[\\w]*dependabot[\\w]*', comparator: "REGEXP"
```
andresmoschini: Did you verify that this is not enough?

```suggestion
                changeRequest author: 'dependabot', comparator: "REGEXP"
```

andresmoschini: BTW, `-` is a word character (`\w`)

cbernat: Perhaps this suffice I'm not so sure. I guess I can test this approach with my user to be sure.

cbernat:
test this please.

PR #605, opened 2019-11-21
chore(App): change default language to spanish

fcoronelmakingsense:
- Because local when the browser is spanish in public pages change to english because the user has setting english language


andresmoschini:
👍

PR #612, opened 2019-11-25
feat: add component Subscriber History

GusBaamonde:
SubscriberHistory

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,6 +14,7 @@ import { Helmet } from 'react-helmet';
 import { availableLanguageOrNull } from './i18n/utils';
 import Shopify from './components/Integrations/Shopify/Shopify';
 import SignupConfirmation from './components/Signup/SignupConfirmation';
+import SubscriberHistory from './components/Reports/SubscriberHistory/SubscriberHistory';
 
 class App extends Component {
   /**
@@ -101,6 +102,7 @@ class App extends Component {
             />
             <PrivateRoute path="/reports/" exact requireSiteTracking component={Reports} />
             <PrivateRoute path="/integrations/shopify" exact component={Shopify} />
+            <PrivateRoute path="/reports/subscriber-history" exact component={SubscriberHistory} />
             <PublicRouteWithLegacyFallback exact path="/login" />
             <PublicRouteWithLegacyFallback exact path="/signup" />
             <PublicRouteWithLegacyFallback exact path="/login/reset-password" />
--- a/src/components/Reports/SubscriberHistory/SubscriberHistory.js
+++ b/src/components/Reports/SubscriberHistory/SubscriberHistory.js
@@ -1,0 +1,34 @@
+import React from 'react';
+import { FormattedMessage, useIntl } from 'react-intl';
+import { Helmet } from 'react-helmet';
+
+const SubscriberHistory = () => {
+  const intl = useIntl();
+  const _ = (id, values) => intl.formatMessage({ id: id }, values);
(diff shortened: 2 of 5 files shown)


fcoronelmakingsense:
Good!

src/App.js
```diff
@@ -14,6 +14,7 @@ import { Helmet } from 'react-helmet';
 import { availableLanguageOrNull } from './i18n/utils';
 import Shopify from './components/Integrations/Shopify/Shopify';
 import SignupConfirmation from './components/Signup/SignupConfirmation';
+import SubscriberHistory from './components/SubscriberHistory/SubscriberHistory';
```
cbernat: If we were saying that `SubscriberHistory` is a report, wouldn't it be better to add into reports folder?

src/components/SubscriberHistory/SubscriberHistory.js
```diff
@@ -0,0 +1,31 @@
...
+import { FormattedMessage } from 'react-intl';
+import { Helmet } from 'react-helmet';
+
+const SubscriberHistory = () => {
+  return (
+    <>
+      <FormattedMessage id="subscriber_history.page_title">
+        {(page_title) => (
+          <Helmet>
+            <title>{page_title}</title>
```
cbernat: Here I think we should add description as in the other pages.
```
<meta name="description" content={_('subscriberHistory_description')} />
```

src/components/SubscriberHistory/SubscriberHistory.test.js
```diff
@@ -0,0 +1,18 @@
...
+describe('SubscriberHistory component', () => {
+  afterEach(cleanup);
+
+  it('renders correctly', () => {
+    const { getByText } = render(
+      <IntlProvider>
+        <SubscriberHistory />
+      </IntlProvider>,
+    );
+    expect(getByText('subscriber_history.header_title')).toBeInTheDocument();
```
cbernat: How about adding the comments for each part of the test? Like this 
```
   // Arrange

   // Act
   const { getByText } = render(
      <IntlProvider>
        <SubscriberHistory />
      </IntlProvider>,
    );
   // Assert
    expect(getByText('subscriber_history.header_title')).toBeInTheDocument();
```

src/components/SubscriberHistory/SubscriberHistory.test.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import IntlProvider from '../../i18n/DopplerIntlProvider.double-with-ids-as-values';
+import SubscriberHistory from './SubscriberHistory';
+
+describe('SubscriberHistory component', () => {
```
cbernat: ❤️

cbernat:
Very nice @GusBaamonde! ❤️ 🎉

src/App.js
fcoronelmakingsense: Yes i think the same. I talked yesterday about it. But maybe need to refactor the reports too because we have a "reports" folder with `Reports.js` file that we use in Advanced reports. But maybe can change in this pr too

src/components/SubscriberHistory/SubscriberHistory.js
fcoronelmakingsense: Maybe need to define the content because in production don't have it
image

cbernat:
👍

PR #616, opened 2019-11-29
DW-129 clean workspace when finished

cbernat:
Cleanup workspace after build complete.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -88,6 +88,14 @@ pipeline {
             }
         }
     }
+    post { 
+        cleanup { 
+            cleanWs()
+            dir("${env.WORKSPACE}@tmp") {
+                deleteDir()
+            }
+        }
+    }
 }
 
 def boolean isVersionTag(String tag) {


andresmoschini:
Is it so simple?

cbernat:
> Is it so simple?

Seems to be 😄 only one way to know would be to test this.

cbernat:
image
For this PR that is #616 in workspace the foder has been removed. Except temp one.

andresmoschini:
>  For this PR that is #616 in workspace the foder has been removed. Except temp one.

Maybe temp folder is removed automatically in background.

I am pretty sure that @elsupergomez knows.

cbernat:
> > For this PR that is #616 in workspace the foder has been removed. Except temp one.
> 
> Maybe temp folder is removed automatically in background.
> 
> I am pretty sure that @elsupergomez knows.

With the last fix it removes temp file also
image

Jenkinsfile
```diff
@@ -88,6 +88,14 @@ pipeline {
             }
         }
     }
+    post { 
+        always {
```
elsupergomez: Maybe is better do the clean in the *cleanup* and leave the *always* for other post actions

```suggestion
        cleanup{ 
```

andresmoschini:
> I am pretty sure that @elsupergomez knows.

> Right now is the best solution, I will apply it to other projects.
> And we should follow this [JENKINS-41805](https://issues.jenkins-ci.org/browse/JENKINS-41805) to know if will handle by Jenkins in the future.

I knew! I knew!

PR #618, opened 2019-11-29
fix: outline function enabled for keyboard navigation in reports

GusBaamonde:
Related to: https://github.com/FromDoppler/Doppler-UI-System/pull/217

**No keyboard navigation**
outline-disabled

**With keyboard navigation**
outline-enabled

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -14,7 +14,7 @@
     <link
     rel="stylesheet"
     type="text/css"
-    href="https://cdn.fromdoppler.com/doppler-ui-library/build667/css/styles.css"
+    href="https://cdn.fromdoppler.com/doppler-ui-library/v3.25.1/css/styles.css"
   />
     <link rel="stylesheet" type="text/css" href="%PUBLIC_URL%/zoho-chat.css" />
     <!--
--- a/public/index.html
+++ b/public/index.html
@@ -14,7 +14,7 @@
     <link
     rel="stylesheet"
     type="text/css"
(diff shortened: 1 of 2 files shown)


public/index.html
```diff
@@ -14,7 +14,7 @@
     <link
     rel="stylesheet"
     type="text/css"
-    href="https://cdn.fromdoppler.com/doppler-ui-library/v3.25.0/css/styles.css"
+    href="https://cdn.fromdoppler.com/doppler-ui-library/build667/css/styles.css"
```
andresmoschini: Remember adding change it to the final version before merging.

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
```diff
@@ -9,11 +9,11 @@ const HeaderUserMenu = ({ user }) => {
 
   return (
     <div>
-      <span className="user-menu--open">
+      <a href="#" className="user-menu--open">
```
andresmoschini: If I am not wrong, this is not a good practice: if it is not a real link, it should be a button.

andresmoschini:
❤️

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
GusBaamonde: Yes!!! Thanks!

cbernat:
👍

PR #626, opened 2019-12-03
Cleanup for pipeline

cbernat:
Add echo to cleanup to verify if it's working for tag generation.


cbernat:
This test is concluded, I'm closing this.

PR #627, opened 2019-12-05
feat(doppler-api-client): get campaigns deliveries method

fcoronelmakingsense:
- [x] Add service method to get campaigns deliveries
- [ ] Add test to the service 
- [x] Add grid component to show campaigns deliveries and subscriber 
- [ ] Add test component 

https://cdn.fromdoppler.com/doppler-webapp/build1985/#/reports/subscriber-history

### DIFF
--- a/src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistoryCurrentSearch.js
+++ b/src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistoryCurrentSearch.js
@@ -2,9 +2,13 @@ import React, { useState, useEffect } from 'react';
 import { InjectAppServices } from '../../../../services/pure-di';
 import { Loading } from '../../../Loading/Loading';
 import { FormattedMessage } from 'react-intl';
+import SubscriberHistorySentCampaigns from './SubscriberHistorySentCampaigns/SubscriberHistorySentCampaigns';
 
 const SubscriberHistoryCurrentSearch = ({ email, dependencies: { dopplerApiClient } }) => {
   const [state, setState] = useState({ loading: true });
+
+  const showSubscriberCampaigns = () =>
+    setState((prevState) => ({ ...prevState, showSubscriberCampaigns: true }));
 
   useEffect(() => {
     const fetchData = async () => {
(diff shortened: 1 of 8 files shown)


src/services/doppler-api-client.double.ts
```diff
@@ -72,4 +101,22 @@ export class HardcodedDopplerApiClient implements DopplerApiClient {
     //   error: new Error('Dummy error'),
     // };
   }
+
+  public async getSubscriberSentCampaigns(
+    email: string,
+    apikey: string,
+  ): Promise<ResultWithoutExpectedErrors<CampaignDeliveryCollection>> {
+    console.log('getApiSubscriber');
```
cbernat: Here it should be `console.log('getSubscriberSentCampaigns');` right?

fcoronelmakingsense: thanks!

cbernat:
So far so good. How about showing this data in the component?

fcoronelmakingsense:
update: add component

image

src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistoryCurrentSearch.js
```diff
@@ -30,26 +34,33 @@ const SubscriberHistoryCurrentSearch = ({ email, dependencies: { dopplerApiClien
...
+                <th>nombre</th>
+                <th>apellido</th>
+                <th>ranking</th>
+                <th>status</th>
+              </tr>
+            </thead>
+            <tbody>
+              <tr>
+                <td>
+                  <span onClick={showSubscriberCampaigns}>{state.subscriber.email}</span>
```
cbernat: Herne perhaps you could add underline to style or green color to identify as link. What do you think?

src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistorySentCampaigns/SubscriberHistorySentCampaigns.js
```diff
@@ -0,0 +1,61 @@
+import React, { useState, useEffect } from 'react';
+import { InjectAppServices } from '../../../../../services/pure-di';
+import { Loading } from '../../../../Loading/Loading';
+import { FormattedMessage } from 'react-intl';
+
+const SubscriberHistorySentCampaigns = ({ subscriber, dependencies: { dopplerApiClient } }) => {
+  const [state, setState] = useState({ loading: true });
```
cbernat: Why not set state for loading inside fetch as in the previous component? This way all side effects are inside `useEffect`

src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistoryCurrentSearch.js
fcoronelmakingsense: yes i know! it's only for working. When apply the styles i add this. Thanks!

src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistorySentCampaigns/SubscriberHistorySentCampaigns.js
fcoronelmakingsense: Yes but need before too because when render the view before enter effect for a second it's broken

src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistoryCurrentSearch.js
andresmoschini: Maybe, it could be a real link. I mean, it would be great if we could have permalink for the different pages.

andresmoschini: Creo que este tema es importante, por favor si no se entiende lo quepropongo hablemosló.

Es el tipo de cosas que si no hacemos bien desde el comienzo, terminan sin arreglarse nunca y embrollandose.

fcoronelmakingsense: si lo entiendo, pero si es un link real deberia tener una url tambien, decis la misma url parametrizada?

fcoronelmakingsense:
we apply the styles in the https://makingsense.atlassian.net/browse/DW-126 task

src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistoryCurrentSearch.js
andresmoschini: claro, pienso que tendríamos que pensar en tener todas las URLs que podamos.

PR #644, opened 2019-12-19
Dw-142 redirect to whitelist domain

cbernat:
- Rename `RedirectToLegacyUrl` to `SafeRedirect`
- Detect full url in `SafeRedirect` and then redirect to this url or redirect to Doppler
- Add tests for `SafeRedirect` regarding partial and full urls
- Detect partial url
- Add whitelist and tests

### DIFF
--- a/src/components/SafeRedirect.js
+++ b/src/components/SafeRedirect.js
@@ -7,8 +7,7 @@ import RedirectToExternalUrl from './RedirectToExternalUrl';
  * @param { string } props.to
  * @param { import('../services/pure-di').AppServices } props.dependencies
  */
-function RedirectToLegacyUrl({ to, dependencies }) {
+function SafeRedirect({ to, dependencies }) {
   return <RedirectToExternalUrl to={dependencies.appConfiguration.dopplerLegacyUrl + to} />;
 }
-
-export default InjectAppServices(RedirectToLegacyUrl);
+export default InjectAppServices(SafeRedirect);
--- a/src/components/SafeRedirect.js
+++ b/src/components/SafeRedirect.js
@@ -8,6 +8,32 @@ import RedirectToExternalUrl from './RedirectToExternalUrl';
  * @param { import('../services/pure-di').AppServices } props.dependencies
  */
 function SafeRedirect({ to, dependencies }) {
-  return <RedirectToExternalUrl to={dependencies.appConfiguration.dopplerLegacyUrl + to} />;
+  const isPartialUrl = (url) => {
+    return /^\/[^/]/.test(url);
+  };
+
+  const isWhitelisted = (url) => {
+    const loginWhitelist = [
+      'https://academy.fromdoppler.com/',
+      'http://academy.fromdoppler.com/',
+      'http://prod.doppleracademy.com/',
+      'https://prod.doppleracademy.com/',
+      'https://doppleracademy.com/',
+      'http://doppleracademy.com/',
+    ];
+    return loginWhitelist.some((element) => url.startsWith(element));
+  };
+
+  if (isPartialUrl(to)) {
+    return <RedirectToExternalUrl to={dependencies.appConfiguration.dopplerLegacyUrl + to} />;
+  } else if (isWhitelisted(to)) {
+    return <RedirectToExternalUrl to={to} />;
+  } else {
+    return (
+      <RedirectToExternalUrl
+        to={dependencies.appConfiguration.dopplerLegacyUrl + '/Campaigns/Draft'}
+      />
+    );
+  }
 }
 export default InjectAppServices(SafeRedirect);
--- a/src/App.js
+++ b/src/App.js
@@ -8,7 +8,7 @@ import Reports from './components/Reports/Reports';
 import { InjectAppServices } from './services/pure-di';
 import queryString from 'query-string';
 import { OriginCatcher } from './services/origin-management';
-import RedirectToLegacyUrl from './components/RedirectToLegacyUrl';
+import SafeRedirect from './components/SafeRedirect';
 import RedirectWithQuery from './components/RedirectWithQuery';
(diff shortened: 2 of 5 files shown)


src/components/SafeRedirect.js
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * @param { Object } props
+ * @param { string } props.to
+ * @param { import('../services/pure-di').AppServices } props.dependencies
+ */
+function SafeRedirect({ to, dependencies }) {
+  const isCompleteUrl = (url) => {
+    return new RegExp(
+      /^(?:http(s)?:\/\/)?[\w.-]+(?:\.[\w\.-]+)+[\w\-\._~:/?#[\]@!\$&'\(\)\*\+,;=.]+$/,
```
andresmoschini: What do you think about something more simple? like `/^\w+:/` if it matches, it is not a relative URL. 

Of course, it should be refined, but in my opinion, the focus should be on differentiating paths from other things..

What is more, maybe we could verify URLs against the white list, and if it does not match, simply assume that it is a legacy URL and concatenate with `dopplerLegacyUrl`.

cbernat: I thought that too, verify against a whitelist and that's it, but this could provide a bit more information in case of errors in urls as well, for example you added a wrongly form url in whitelist without noticing, perhaps I'm overthinking this...

andresmoschini: Nah, you are right, the problem is that you should deal with the errors hehe

src/components/SafeRedirect.js
```diff
@@ -0,0 +1,37 @@
...
+  const loginWhitelist =
+    [
+      'https://academy.fromdoppler.com/',
+      'http://prod.doppleracademy.com/',
+      'https://doppleracademy.com/',
+    ] || whitelist;
+
+  const isPartialUrl = (url) => {
+    return RegExp(/^\/[\/\w*]*[\w\?\&\=]*$/).test(url); // eslint-disable-line
+  };
```
andresmoschini: What do you think about simply:

```js
const isPartialUrl = /^\/[\/\w*]*[\w\?\&\=]*$/.test;
```

andresmoschini: Why do you need an eslint exception?

andresmoschini: I am not sure about these criteria, what happens if the redirect destination starts with `?` or has any special character? I think that we can discuss it better in a quick call.

src/components/SafeRedirect.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param { string } props.to
+ * @param { import('../services/pure-di').AppServices } props.dependencies
+ */
+function SafeRedirect({ to, whitelist, dependencies }) {
+  const loginWhitelist =
+    [
+      'https://academy.fromdoppler.com/',
+      'http://prod.doppleracademy.com/',
+      'https://doppleracademy.com/',
+    ] || whitelist;
```
andresmoschini: The parameter `whitelist` will be never taken into account.

src/components/SafeRedirect.js
```diff
@@ -0,0 +1,37 @@
...
+
+  // whitelist = whitelist || loginWhitelist;
+
+  const whitelisted = (url, whitelist) => {
+    return whitelist && whitelist.find((element) => url.indexOf(url) !== -1);
+  };
+
+  if (isPartialUrl(to)) {
+    return <RedirectToExternalUrl to={dependencies.appConfiguration.dopplerLegacyUrl + to} />;
+  } else if (whitelisted(to, loginWhitelist)) {
```
andresmoschini: I think that making this function parametrizable is not necessary, I mean it has a very clear responsibility, it is not necessary making this reusable.

src/components/SafeRedirect.js
cbernat: > Why do you need an eslint exception?

image

andresmoschini: And is it wrong?

src/components/SafeRedirect.js
cbernat: I still think it add readability, it's quicker to understand.

andresmoschini: In that case, we should disable the rule globally because we do not want to take it into account.

src/components/SafeRedirect.js
cbernat: I don't like to have useless warnings... Im taking this characters as characters and not as quantifiers.

src/components/SafeRedirect.js
andresmoschini: Pedon, confundí el comentario, luego lo hablamos.

src/components/SafeRedirect.js
```diff
@@ -0,0 +1,39 @@
...
+  const isWhitelisted = (url) => {
+    const loginWhitelist = [
+      'https://academy.fromdoppler.com/',
+      'http://academy.fromdoppler.com/',
+      'http://prod.doppleracademy.com/',
+      'https://prod.doppleracademy.com/',
+      'https://doppleracademy.com/',
+      'http://doppleracademy.com/',
+    ];
+    return !!loginWhitelist.find((element) => url.startsWith(element));
```
andresmoschini: What do you think about using this?

```suggestion
    return loginWhitelist.some((element) => url.startsWith(element));
```

Or better yet:

```suggestion
    return loginWhitelist.some(url.startsWith);
```

cbernat: Second approach saddly did not work.

andresmoschini: Thanks!, when I have time I have to try it in order to understand what I am missing.

cbernat: So will I 😄

PR #649, opened 2019-12-23
DW-151 Add logger for log errors in login

cbernat:
- Added loggly in index
- Added function to log an entry
- Added log entry in login only for unexpected error with the following information:
   - account: user account for doppler
   - origin: href origin to know enviroment of origin
   - section: right now only login
   - browser: browser in which the error occur
   - error: callback result with error

[Loggly Setup](https://dopplerwebapp.loggly.com/sources/setup/js-app-setup) 
To see the log with correct credentials [here](https://dopplerwebapp.loggly.com/search#terms=&from=2019-12-20T17:52:33.937Z&until=2019-12-20T18:22:33.937Z&source_group=)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -12,10 +12,10 @@
     -->
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <link
-    rel="stylesheet"
-    type="text/css"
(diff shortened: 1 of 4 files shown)


andresmoschini:
👍

PR #650, opened 2019-12-23
fix: update build version

GusBaamonde:
**Update build version to correct select arrows**

bug-drpdown

bug-arrows-mac


cbernat:
🚀

PR #651, opened 2019-12-26
feat(SubscriberHistory): add service method to get subscribers

fcoronelmakingsense:
- Add the service method to get subscribers and the doubles and tests

### DIFF
--- a/src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistoryCurrentSearch.js
+++ b/src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistoryCurrentSearch.js
@@ -7,22 +7,41 @@ import SubscriberHistorySentCampaigns from './SubscriberHistorySentCampaigns/Sub
 const SubscriberHistoryCurrentSearch = ({ searchText, dependencies: { dopplerApiClient } }) => {
   const [state, setState] = useState({ loading: true });
 
-  const showSubscriberCampaigns = () =>
-    setState((prevState) => ({ ...prevState, showSubscriberCampaigns: true }));
+  const showSubscriberCampaigns = (subscriber) =>
+    setState((prevState) => ({
+      ...prevState,
+      showSubscriberCampaigns: true,
+      selectedSubscriber: subscriber,
+    }));
 
   useEffect(() => {
     const fetchData = async () => {
       setState({ loading: true });
-      const response = await dopplerApiClient.getSubscriber(searchText);
+      const response = await dopplerApiClient.getSubscribers(searchText);
       if (response.success) {
-        const subscriber = {
-          ...response.value,
-          firstName: response.value.fields.find((x) => x.name === 'FIRSTNAME'),
-          lastName: response.value.fields.find((x) => x.name === 'LASTNAME'),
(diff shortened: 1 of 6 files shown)


src/services/doppler-api-client.ts
```diff
@@ -11,6 +11,7 @@ export interface DopplerApiClient {
     email: string,
     apikey: string,
   ): Promise<ResultWithoutExpectedErrors<CampaignDeliveryCollection>>;
+  getSubscribers(email: string): Promise<ResultWithoutExpectedErrors<SubscriberCollection>>;
```
cbernat: Remember here instead of `email` we are using a searching text that can be part of: email, last name or name.

src/services/doppler-api-client.ts
```diff
@@ -205,4 +226,35 @@ export class HttpDopplerApiClient implements DopplerApiClient {
...
+
+  public async getSubscribers(
+    email: string,
+  ): Promise<ResultWithoutExpectedErrors<SubscriberCollection>> {
+    try {
+      const { jwtToken, userAccount } = this.getDopplerApiConnectionData();
+
+      const { data } = await this.axios.request({
+        method: 'GET',
+        url: `/accounts/${userAccount}/subscribers/`,
```
cbernat: Where do you plan on add the fallback for this method? since it's still not deployed in production.

fcoronelmakingsense: in the component. not in the service because not depends of that

src/services/doppler-api-client.ts
fcoronelmakingsense: ohh you right! thanks

src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistoryCurrentSearch.js
```diff
@@ -7,22 +7,41 @@ import SubscriberHistorySentCampaigns from './SubscriberHistorySentCampaigns/Sub
...
+          return {
+            ...subscriber,
+            firstName: subscriber.fields.find((x) => x.name === 'FIRSTNAME'),
+            lastName: subscriber.fields.find((x) => x.name === 'LASTNAME'),
+          };
+        });
+        setState({ loading: false, subscribers: subscribers });
       } else {
-        setState({ loading: false });
+        // Fallback for production
```
cbernat: Beware of this fallback, if the endpoint does not exist in production you will get a 404, not a success false. 
image

cbernat: I would add a try in `getSubscribers` call and as a catch a fallback to `getSubscriber`.

fcoronelmakingsense: ok i added. now im waiting for build to test in cdn

cbernat: We still have a problem... 
image

cbernat:
🚀

PR #653, opened 2019-12-27
feat(log): add trace and fullResponse in unexpected error to be logged

cbernat:
### Weird wrong captcha error in log
Logging unexpected errors we found some weird wrongCaptcha error, that needed more data to be understanded. Why? because it's nearly impossible to reproduce, the captcha does not allow to send a wrong captcha to BE and the message is sent from BE. 

In loggly:

image

Message from BE in doppler --> https://github.com/MakingSense/Doppler/blob/develop/Doppler.Presentation.MVC/Controllers/WebAppPublicController.cs#L312

### Data added:
- trace from FE
- full response from service call

### DIFF
--- a/src/doppler-types.ts
+++ b/src/doppler-types.ts
@@ -1,10 +1,4 @@
-export type UnexpectedError = {
-  success?: false;
-  message?: string;
-  trace?: any;
-  fullResponse?: any;
-  error?: any;
-};
+export type UnexpectedError = { success?: false } | { success?: false; [key: string]: any };
 export type ErrorResult<TError> = { success?: false; expectedError: TError } | UnexpectedError;
 export type Result<TResult, TError> = { success: true; value: TResult } | ErrorResult<TError>;
 export type ResultWithoutExpectedErrors<TResult> =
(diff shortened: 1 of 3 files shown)


src/doppler-types.ts
```diff
@@ -1,4 +1,10 @@
-export type UnexpectedError = { success?: false; message?: string | null; error?: any };
+export type UnexpectedError = {
+  success?: false;
+  message?: string;
+  trace?: any;
+  fullResponse?: any;
+  error?: any;
```
andresmoschini: Is it not possible adding properties dinamically? Altough maybe it is better having them here, I am not sure.

cbernat: Like having an object of properties for example? 
```
export type UnexpectedError = {
  success?: false;
  properties?: any;
};
```
Then in `properties`:
```
properties.trace= "...";
properties.fullResponse={...};
...
```

andresmoschini: I thought that it was possible without another object, but I do not know. I had not enough practice with TypeScript

cbernat: I think it's ok to have success explicit the rest of the properties can be dynamic. Like this:
```
export type UnexpectedError = {
  success?: false;
  [key: string]: any;
};
```

PR #662, opened 2020-01-09
feat(login): map max attemps error into message

cbernat:
Add message for login maximum attempts reached, instead of unexpected error message. 

image

Issue: DW-157

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -128,7 +128,11 @@ const Login = ({ location, dependencies: { dopplerLegacyClient, sessionManager, 
         setErrors({
           _error: <FormattedHTMLMessage id="validation_messages.error_account_is_canceled_HTML" />,
         });
-      } else if (result.expectedError && result.expectedError.blockedAccountInvalidPassword) {
+      } else if (
+        result.expectedError &&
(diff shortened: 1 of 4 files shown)


cbernat:
> for me its ok, i dont know if can add some test, to catch this in the future

This issue was detected with the log implementation, we missed in doppler - webapp migration, perhaps a service test mocking all responses like in doppler would help. We have some mocks in legacy service double but not all possible responses.

src/components/Login/Login.js
```diff
@@ -128,7 +128,11 @@ const Login = ({ location, dependencies: { dopplerLegacyClient, sessionManager,
         setErrors({
           _error: <FormattedHTMLMessage id="validation_messages.error_account_is_canceled_HTML" />,
         });
-      } else if (result.expectedError && result.expectedError.blockedAccountInvalidPassword) {
+      } else if (
+        result.expectedError &&
+        (result.expectedError.blockedAccountInvalidPassword ||
+          result.expectedError.maxLoginAttempts)
+      ) {
```
andresmoschini: Se podría agregar un test para tener en cuenta este nuevo comportamiento.

PR #673, opened 2020-01-13
Feat: Add SMS credits in user menu

lucianaDilizio:
Add SMS credits in user menu

image


src/components/Header/HeaderUserMenu/HeaderUserMenu.js
```diff
@@ -62,6 +63,26 @@ const HeaderUserMenu = ({ user }) => {
             ''
           )}
         </div>
+        {user.sms.description ? (
+          <div className="user-plan--container">
+            <div className="user-plan--type">
+              <p><strong className={user.sms.remainingCredits < 0 ? 'red' : ''}>US$ {' '}
```
fcoronelmakingsense: maybe its better a function to get the className to have the html more clean. what do you think @cbernat

src/components/Header/HeaderUserMenu/HeaderUserMenu.test.js
```diff
@@ -71,4 +77,26 @@ describe('Header user menu', () => {
     );
     expect(getByText('Upgrade')).toBeInTheDocument();
   });
+
+  it('user have SMS enabled', () => {
+    const { getByText } = render(
```
fcoronelmakingsense: Can add the comments like `// Arrange // Act // Assert` ?

lucianaDilizio: If I'm not wrong, I didn't see them in any test file. Maybe this could be part of a refactor.

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
cbernat: I agree, would you mind change this into a function @lucianaDilizio please?

src/components/Header/HeaderUserMenu/HeaderUserMenu.test.js
cbernat: When we remember we add this comments, you are more than welcome to remind us about this in the future too 😝

src/components/Header/HeaderUserMenu/HeaderUserMenu.test.js
```diff
@@ -71,4 +77,26 @@ describe('Header user menu', () => {
     );
     expect(getByText('Upgrade')).toBeInTheDocument();
   });
+
+  it('user have SMS enabled', () => {
```
cbernat:
```suggestion
  it('should show available SMS text when user have SMS enabled', () => {
```

src/components/Header/HeaderUserMenu/HeaderUserMenu.test.js
```diff
@@ -71,4 +77,26 @@ describe('Header user menu', () => {
...
+    const { getByText } = render(
+      <IntlProvider>
+        <HeaderUserMenu user={userData} />
+      </IntlProvider>,
+    );
+    expect(getByText('Available for SMS')).toBeInTheDocument();
+  });
+
+  it('user have SMS disabled', () => {
```
cbernat:
```suggestion
  it('should not show SMS text when user have SMS disabled', () => {
```

src/components/Header/HeaderUserMenu/HeaderUserMenu.test.js
```diff
@@ -71,4 +77,26 @@ describe('Header user menu', () => {
...
     expect(getByText('Upgrade')).toBeInTheDocument();
   });
+
+  it('user have SMS enabled', () => {
+    const { getByText } = render(
+      <IntlProvider>
+        <HeaderUserMenu user={userData} />
+      </IntlProvider>,
+    );
+    expect(getByText('Available for SMS')).toBeInTheDocument();
```
cbernat: How about here validating against the variable? `userData.sms.description`
```suggestion
    expect(getByText(userData.sms.description)).toBeInTheDocument();
```

src/components/Header/HeaderUserMenu/HeaderUserMenu.test.js
```diff
@@ -71,4 +77,26 @@ describe('Header user menu', () => {
...
+    userData.sms = {
+      remainingCredits: '',
+    };
+
+    const { queryByText } = render(
+      <IntlProvider>
+        <HeaderUserMenu user={userData} />
+      </IntlProvider>,
+    );
+    expect(queryByText('Available for SMS')).not.toBeInTheDocument();
```
cbernat:
```suggestion
    expect(queryByText(userData.sms.description)).not.toBeInTheDocument();
```

src/App.test.js
```diff
@@ -168,6 +168,7 @@ describe('App component', () => {
               lang: 'es',
               avatar: {},
               plan: {},
+              sms: {},
```
cbernat: In all scenarios are you sure back end always returns an sms object?

lucianaDilizio: I guess so. Tests fail if I don't add sms object.

src/components/Header/HeaderUserMenu/HeaderUserMenu.test.js
lucianaDilizio: In this case `userData.sms.description` doesn't have any value. I want to verify that there's no any element with this specific text.

cbernat: In which way do you think we can avoid this hardcoded value and still validate?

src/App.test.js
cbernat: But we always get sms object from BE? otherwise you should add a test to allow not having it. And also check if the app breaks.

lucianaDilizio: I didn't see any case without this info. Even in free accounts we receive this value empty. I can apply a test to verify what you say.

cbernat: Great, then no need. Thanks! ❤️

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
```diff
@@ -62,6 +69,27 @@ const HeaderUserMenu = ({ user }) => {
             ''
           )}
         </div>
+        {user.sms.description ? (
```
cbernat: Perhaps here instead of ask for a property this way, how about doing this?
```suggestion
        {Object.keys(user.sms).length ? (
```

src/services/doppler-legacy-client.ts
```diff
@@ -244,6 +259,7 @@ export function mapHeaderDataJson(json: any) {
       lang: json.user.lang,
       nav: (json.user.nav && json.user.nav.map(mapNavEntry)) || [],
       plan: mapPlanEntry(json.user.plan),
+      sms: mapSmsEntry(json.user?.sms),
```
cbernat: How about avoiding all inner mapping if there are no properties? Alongside with changes in `HeaderUserMenu` component line 72. 
```suggestion
      sms: Object.keys(Json.user.sms).length? mapSmsEntry(json.user.sms): {},
```

cbernat:
Please test this as we talked in production and in qa (both for sms user and not sms enabled user) and thten I suppose we are done! 😄

src/components/Header/HeaderUserMenu/HeaderUserMenu.js
```diff
@@ -62,6 +69,27 @@ const HeaderUserMenu = ({ user }) => {
             ''
           )}
         </div>
+        {Object.keys(user.sms).length ? (
```
andresmoschini: Is there no cleaner way than `Object.keys(user.sms)`?

src/services/doppler-legacy-client.ts
```diff
@@ -244,6 +259,7 @@ export function mapHeaderDataJson(json: any) {
       lang: json.user.lang,
       nav: (json.user.nav && json.user.nav.map(mapNavEntry)) || [],
       plan: mapPlanEntry(json.user.plan),
+      sms: json.user.sms?.description ? mapSmsEntry(json.user.sms) : {},
```
andresmoschini: Creo que aquí sería mejor devolver null cuando no hay info de sms, o en su defecto un objeto que contenga una propiedad booleana que indique que no tiene info de sms.

PR #676, opened 2020-01-14
fix(CampaignsHistory): add star icons in grid

fcoronelmakingsense:
- Add star icons in subscriber search results
- Add star icons in campaigns history results

image

image

### DIFF
--- a/src/components/Reports/CampaignsHistory/CampaignsHistory.js
+++ b/src/components/Reports/CampaignsHistory/CampaignsHistory.js
@@ -11,10 +11,10 @@ function extractEmail(location) {
   return (parsedQuery && parsedQuery['email']) || null;
 }
 
-const getDeliveryStatusCssClassName = (deliveryStatus) => {
-  let deliveryCssClass = '';
-  switch (deliveryStatus) {
-    case 'opened':
+const getStarCssClassName = (actualStarNumber, score) => {
+  return actualStarNumber <= score ? 'icon-star' : 'icon-star-filled';
+};
+
       deliveryCssClass = 'status--opened';
       break;
     case 'notOpened':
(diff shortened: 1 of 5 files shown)


src/components/Reports/CampaignsHistory/CampaignsHistory.js
```diff
@@ -11,6 +11,10 @@ function extractEmail(location) {
   return (parsedQuery && parsedQuery['email']) || null;
 }
 
+const getStarCssClassName = (actualStarNumber, score) => {
```
cbernat: How about refactoring this into utils?

src/components/Reports/SubscriberHistory/SubscriberHistoryCurrentSearch/SubscriberHistoryCurrentSearch.js
```diff
@@ -80,22 +84,17 @@ const SubscriberHistoryCurrentSearch = ({ searchText, dependencies: { dopplerApi
...
-                        ><span
-                          class="ms-icon icon-star"
-                          aria-label="Sin estrella"
-                        ></span>
-                        <span
-                          class="ms-icon icon-star"
-                          aria-label="Sin estrella"
-                        ></span>
-                      </div> */}
+                    <div className="dp-calification">
```
cbernat: Another possibility would be instead of refactoring a function into utils, add a new component for calification, with the function and jsx alltogether. What do you think?

fcoronelmakingsense: i added new component

cbernat:
The stars look very nice refactored into a component ❤️ 🚀

PR #678, opened 2020-01-16
fix(CampaignsHistory): grid empty message

fcoronelmakingsense:
- Change message
- Add tests to check the message


cbernat:
❤️  🚀

PR #680, opened 2020-01-17
DW-177 add redirect url signup

cbernat:
### Allow new registered user from academy to return to academy site
For doppler academy, we need to allow redirect after the user has signed up. 
This url we plan on use in doppler back end after user activation.

- Added redirect url parameter into signup
- Mapped redirect into legacy service to allow redirect into back end
- Before map check redirect url is whitelisted, otherwise map nothing


cbernat:
https://cdn.fromdoppler.com/doppler-webapp/int-build2080/#/signup?redirect=http://prod.doppleracademy.com/
https://cdn.fromdoppler.com/doppler-webapp/int-build2080/#/signup

PR #688, opened 2020-01-20
feat(CampaignsHistory): add pagination

fcoronelmakingsense:
- create new component to manage the pagination
- change service method to use the current page and the quantity per page
- change page when click in numbers

### DIFF
--- a/src/services/doppler-api-client.double.ts
+++ b/src/services/doppler-api-client.double.ts
@@ -49,34 +49,29 @@ const subscriber = {
   score: 2,
 };
 
-const campaignDeliveryCollection = {
-  items: [
-    {
-      campaignId: 1,
-      campaignName: 'Campaña estacional de primavera',
-      campaignSubject: '¿Como sacarle provecho a la primavera?',
-      deliveryStatus: 'opened',
-      clicksCount: 2,
-    },
-    {
-      campaignId: 2,
-      campaignName: 'Campaña calendario estacional 2019',
-      campaignSubject: 'El calendario estacional 2019 ya está aquí',
-      deliveryStatus: 'notOpened',
-      clicksCount: 23,
-    },
-    {
-      campaignId: 3,
-      campaignName: 'Emms 2019 preveento 1',
-      campaignSubject: 'Ya comienza el dia 2. Accede a las conferencias',
-      deliveryStatus: 'softBounced',
-      clicksCount: 100,
-    },
-  ],
-  currentPage: 0,
-  itemsCount: 3,
-  pagesCount: 1,
-};
+const campaignDeliveryItems = [
+  {
+    campaignId: 1,
+    campaignName: 'Campaña estacional de primavera',
+    campaignSubject: '¿Como sacarle provecho a la primavera?',
+    deliveryStatus: 'opened',
+    clicksCount: 2,
+  },
+  {
+    campaignId: 2,
+    campaignName: 'Campaña calendario estacional 2019',
+    campaignSubject: 'El calendario estacional 2019 ya está aquí',
+    deliveryStatus: 'notOpened',
+    clicksCount: 23,
+  },
+  {
+    campaignId: 3,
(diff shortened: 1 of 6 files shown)


src/services/doppler-api-client.ts
```diff
@@ -9,7 +9,8 @@ export interface DopplerApiClient {
   getSubscriber(email: string, apikey: string): Promise<ResultWithoutExpectedErrors<Subscriber>>;
   getSubscriberSentCampaigns(
     email: string,
-    apikey: string,
+    campaignsPerPage: number,
```
cbernat: How about adding some test for this mappings?

src/services/doppler-api-client.double.ts
```diff
@@ -49,34 +49,29 @@ const subscriber = {
...
-      campaignSubject: 'Ya comienza el dia 2. Accede a las conferencias',
-      deliveryStatus: 'softBounced',
-      clicksCount: 100,
-    },
-  ],
-  currentPage: 0,
-  itemsCount: 3,
-  pagesCount: 1,
-};
+const campaignDeliveryItems = [
```
cbernat: I think here perhaps you could add more data, for us to see how this goes, visually. What do you think?

cbernat:
So far so good. Very nice! 💪

src/services/doppler-api-client.ts
fcoronelmakingsense: i added

src/services/doppler-api-client.double.ts
```diff
@@ -171,11 +166,19 @@ export class HardcodedDopplerApiClient implements DopplerApiClient {
...
   ): Promise<ResultWithoutExpectedErrors<CampaignDeliveryCollection>> {
-    console.log('getSubscriberSentCampaigns');
+    console.log('getSubscriberSentCampaigns', email, campaignsPerPage, currentPage);
     await timeout(1500);
 
+    const campaignDeliveryCollection = {
+      items: campaignDeliveryItems,
+      currentPage: currentPage,
+      itemsCount: campaignDeliveryItems.length,
+      pagesCount: 2,
```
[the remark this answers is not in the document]
fcoronelmakingsense: added

src/services/doppler-api-client.double.ts
fcoronelmakingsense: yes i added now

src/services/doppler-api-client.double.ts
```diff
@@ -49,35 +49,49 @@ const subscriber = {
...
+  return {
+    campaignId: index,
+    campaignName: 'Campaña estacional de primavera',
+    campaignSubject: '¿Como sacarle provecho a la primavera?',
+    deliveryStatus: getDeliveryStatus(Math.round(Math.random() * (5 - 1) + 1)),
+    clicksCount: Math.round(Math.random() * (100 - 1) + 1),
+  };
+});
+
+// const campaignDeliveryItems = [
```
cbernat: Why do you have this commented code? are you going to use it for tests?

fcoronelmakingsense: i saw it, its old! i already deleted

src/services/doppler-api-client.double.ts
```diff
@@ -49,35 +49,29 @@ const subscriber = {
...
+    case 2:
+      return 'softBounced';
+    case 3:
+      return 'hardBounced';
+    default:
+      return 'notOpened';
+  }
 };
 
+const campaignDeliveryItems = [...Array(100)].map((_, index) => {
```
cbernat: 😄

cbernat:
I've tested this in cdn pointing to prod, and it's working great! 💪 https://cdn.fromdoppler.com/doppler-webapp/build2101/#/reports/campaigns-history?email=cbernat@makingsense.com&page=2

cbernat:
If a user has no campaigns I see this
image

Could you please recheck?

PR #690, opened 2020-01-21
fix(doppler-sites-client): get banner map data

fcoronelmakingsense:
Fix mapping when get banner data from doppler page


cbernat:
👍

PR #692, opened 2020-01-21
DW-158:feat: capture wrong captcha error from BE

cbernat:
When captcha fails in back end, the error is displayed in the logs, but we are not warning de user about this kind of error, and we do not know what migth be triggering it. 

Informing the user can help us if he/she contacts support. They can provide with steps to reproduce this issue. 

image

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -140,6 +140,16 @@ const Login = ({ location, dependencies: { dopplerLegacyClient, sessionManager, 
         });
       } else if (result.expectedError && result.expectedError.invalidLogin) {
         setErrors({ _error: 'validation_messages.error_invalid_login' });
+      } else if (result.expectedError && result.expectedError.wrongCaptcha) {
+        setErrors({ _error: 'validation_messages.error_invalid_captcha' });
+        addLogEntry({
+          account: values[fieldNames.user],
+          origin: window.location.origin,
+          section: 'Login',
+          browser: window.navigator.userAgent,
+          error: result,
+        });
+        console.log('invalid captcha', result);
       } else {
         console.log('Unexpected error', result);
         setErrors({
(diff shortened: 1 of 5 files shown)


src/components/Login/Login.js
```diff
@@ -140,6 +140,9 @@ const Login = ({ location, dependencies: { dopplerLegacyClient, sessionManager,
         });
       } else if (result.expectedError && result.expectedError.invalidLogin) {
         setErrors({ _error: 'validation_messages.error_invalid_login' });
+      } else if (result.expectedError && result.expectedError.wrongCaptcha) {
+        setErrors({ _error: 'validation_messages.error_invalid_captcha' });
+        console.log('invalid captcha', result);
```
andresmoschini: Is this also logged with _loggly_?

cbernat: Right now is logged in within loggly, this was an attempt to capture this as an expected error, but we need a message and an advise we could give to a user experimenting this issue. I'm not so sure about that...

cbernat: If I upload this code, It will stop showing in loggly and the user will see an special error for this scenario.

andresmoschini: Oh, it would be nice logging these errors with loggly also, right?

cbernat: Remember we would log only unexpected errors because of the log size? :trollface: Or do you suggest we do an exception just for this particular error? I'm not sure of merging this though...

andresmoschini: I was talking about logging the errors that could be making us troubles, without worry about if they are being shown to the user or not.

If you do not log this error, maybe you will have better information from the user when he calls support service, but you will not have any details about the request, response, etc.

cbernat: Ok, that seems nice!

cbernat: @andresmoschini fixed this issue.

andresmoschini: yo? yo no arreglé nada, a mi que me revisen...

cbernat: ??? Sume loggear el error tambien, eso 😄

PR #699, opened 2020-01-24
fix(SubscriberHistory): change component name to MasterSubscriber

fcoronelmakingsense:
- Change subscriber history to master subscriber

### DIFF
--- a/src/components/Reports/SubscriberHistory/SubscriberHistory.test.js
+++ b/src/components/Reports/SubscriberHistory/SubscriberHistory.test.js
@@ -2,11 +2,11 @@ import React from 'react';
 import { render, cleanup, wait } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 import IntlProvider from '../../../i18n/DopplerIntlProvider.double-with-ids-as-values';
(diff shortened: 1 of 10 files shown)


src/components/Reports/SubscriberHistory/SubscriberHistory.test.js
```diff
@@ -1,23 +1,219 @@
 import React from 'react';
-import { render, cleanup } from '@testing-library/react';
+import { render, cleanup, wait } from '@testing-library/react';
```
cbernat: Why did you add this new `wait` here? it was not here before.

cbernat: Forget this comment, this was marked as change in all files view, but in this commit it has no changes.

cbernat:
Seems ok, thanks for the small commits, it makes it a lot easier to understand. 🚀

src/components/Reports/SubscriberHistory/SubscriberHistory.test.js
fcoronelmakingsense: because in the old file before rename dont have and now yes in the new file

PR #700, opened 2020-01-24
fix(SubscriberHistory): add link preview in campaigns history

fcoronelmakingsense:
- Add links mappings in service

### DIFF
--- a/src/services/doppler-api-client.double.ts
+++ b/src/services/doppler-api-client.double.ts
@@ -69,6 +69,13 @@ const campaignDeliveryItems = [...Array(100)].map((_, index) => {
     campaignSubject: '¿Como sacarle provecho a la primavera?',
     deliveryStatus: getDeliveryStatus(Math.round(Math.random() * (5 - 1) + 1)),
     clicksCount: Math.round(Math.random() * (100 - 1) + 1),
+    links: [
+      {
+        href: 'http://dopplerfilesint.fromdoppler.net/Users/50018/Campaigns/33850437/33850437.png',
+        description: 'get preview',
+        rel: '/docs/rels/get-campaign-preview',
+      },
+    ],
   };
 });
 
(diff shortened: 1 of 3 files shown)


src/services/doppler-api-client.ts
```diff
@@ -121,13 +128,22 @@ export class HttpDopplerApiClient implements DopplerApiClient {
...
+  }
+
   private mapCampaignsDelivery(data: any): CampaignDelivery[] {
     return data.map((x: any) => ({
       campaignId: x.campaignId,
       campaignName: x.campaignName,
       campaignSubject: x.campaignSubject,
       clicksCount: x.clicksCount,
       deliveryStatus: x.deliveryStatus,
+      links: this.mapLinks(x._links),
```
cbernat: Have taken in account that maybe there could be no links?

fcoronelmakingsense: yes!

cbernat: In no links case it would be empty array. So it's ok.

src/services/doppler-api-client.double.ts
```diff
@@ -69,6 +69,13 @@ const campaignDeliveryItems = [...Array(100)].map((_, index) => {
     campaignSubject: '¿Como sacarle provecho a la primavera?',
     deliveryStatus: getDeliveryStatus(Math.round(Math.random() * (5 - 1) + 1)),
     clicksCount: Math.round(Math.random() * (100 - 1) + 1),
+    links: [
+      {
+        href: 'http://dopplerfilesint.fromdoppler.net/Users/50018/Campaigns/33850437/33850437.png',
```
cbernat: Here you can have several links that have nothing to do whith preview, perhaps add one like that, and a filter in map funcion. What do you think?

fcoronelmakingsense: maybe we can use in the future this links. for now we can keep it

cbernat:
🚀

PR #711, opened 2020-01-30
DW-186: axios retry for login only with log

cbernat:
### Login log unexpected errors
For network error, or timeout errors retry 3 times automatically. 

image

image

### DIFF
--- a/src/services/doppler-legacy-client.ts
+++ b/src/services/doppler-legacy-client.ts
@@ -1,6 +1,8 @@
-import { AxiosInstance, AxiosStatic } from 'axios';
+import { AxiosInstance, AxiosStatic, AxiosError } from 'axios';
 import { Color } from 'csstype';
 import { Result, EmptyResult, EmptyResultWithoutExpectedErrors } from '../doppler-types';
+import axiosRetry from 'axios-retry';
+import { addLogEntry } from '../utils';
 
 export interface DopplerLegacyClient {
   getUserData(): Promise<DopplerLegacyUserData>;
@@ -307,11 +309,33 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
   private readonly axios: AxiosInstance;
   private readonly baseUrl: string;
 
+  private isEnabledForRetry(error: AxiosError) {
+    return error.config && error.config.url === 'WebAppPublic/Login';
+  }
+
   constructor({ axiosStatic, baseUrl }: { axiosStatic: AxiosStatic; baseUrl: string }) {
     this.baseUrl = baseUrl;
     this.axios = axiosStatic.create({
       baseURL: baseUrl,
       withCredentials: true,
+    });
+
+    axiosRetry(this.axios, {
+      retries: 3,
+      shouldResetTimeout: true,
+      retryCondition: (error) => {
+        return this.isEnabledForRetry(error);
+      },
+      retryDelay: (retryCount: any, error: any) => {
+        addLogEntry({
+          origin: window.location.origin,
+          section: 'Login/legacyClient',
+          browser: window.navigator.userAgent,
+          error: error,
+          retryCount: retryCount,
+        });
+        return retryCount * 200;
+      },
     });
   }
 
--- a/src/services/doppler-legacy-client.ts
+++ b/src/services/doppler-legacy-client.ts
@@ -308,17 +308,27 @@ export interface DopplerLegacyUpgradePlanContactModel {
 export class HttpDopplerLegacyClient implements DopplerLegacyClient {
   private readonly axios: AxiosInstance;
   private readonly baseUrl: string;
+  private window: any;
 
   private isEnabledForRetry(error: AxiosError) {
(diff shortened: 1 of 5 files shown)


src/services/doppler-legacy-client.ts
```diff
@@ -313,6 +340,7 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
   }
 
   public async login(model: LoginModel): Promise<LoginResult> {
+    this.setRetryAxiosAndLog();
```
andresmoschini: I think that it is not a good solution because the _Axios_ instance is shared.

Maybe you could have more than an _Axios_ instance?

For example:

```js
  constructor({ axiosStatic, baseUrl }: { axiosStatic: AxiosStatic; baseUrl: string }) {
    this.baseUrl = baseUrl;
    this.axios = axiosStatic.create({
      baseURL: baseUrl,
      withCredentials: true,
    });
    this.axiosRetry = axiosRetry(axiosStatic.create({
      baseURL: baseUrl,
      withCredentials: true,
    }), {
    retries: 3,
    shouldResetTimeout: true,
    retryCondition: () => {
      return true;
    },
    retryDelay: (retryCount: any, error: any) => {
      addLogEntry({
        origin: window.location.origin,
        section: 'Login/legacyClient',
        browser: window.navigator.userAgent,
        error: error,
        retryCount: retryCount,
      });
      return retryCount * 1000;
    },
  });
}
```

andresmoschini: Another option could be using the same axios instance but modify _retryCondition_ to only work for login.

src/services/doppler-legacy-client.ts
```diff
@@ -300,7 +302,32 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
...
+  private setRetryAxiosAndLog() {
+    axiosRetry(this.axios, {
+      retries: 3,
+      shouldResetTimeout: true,
+      retryCondition: () => {
+        return true;
+      },
+      retryDelay: (retryCount: any, error: any) => {
+        addLogEntry({
+          origin: window.location.origin,
```
andresmoschini: We are trying to avoid using _window_ global object, maybe you could inject `window`, `location` or `navigator` using our DI.

src/services/doppler-legacy-client.ts
```diff
@@ -300,7 +302,32 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
...
+          section: 'Login/legacyClient',
+          browser: window.navigator.userAgent,
+          error: error,
+          retryCount: retryCount,
+        });
+        return retryCount * 1000;
```
andresmoschini: Do we wait for a second? is it not so much?

src/services/doppler-legacy-client.ts
cbernat: Good point!

src/services/doppler-legacy-client.ts
cbernat: I don't have a strong opinion on this, maybe less it's ok too. What do you suggest?

andresmoschini: In my opinion, wait up to 3 seconds is too much.
Maybe waiting only some few milli-seconds?

src/services/doppler-legacy-client.ts
cbernat: Injected window user DI

src/services/doppler-legacy-client.ts
```diff
@@ -306,13 +308,49 @@ export interface DopplerLegacyUpgradePlanContactModel {
...
+  private window: any;
 
-  constructor({ axiosStatic, baseUrl }: { axiosStatic: AxiosStatic; baseUrl: string }) {
+  private isEnabledForRetry(error: AxiosError) {
+    if (error.config.url === 'WebAppPublic/Login') {
+      return true;
+    } else {
+      return false;
+    }
+  }
```
andresmoschini:
```
  private isEnabledForRetry(error: AxiosError) {
    return error.config.url === 'WebAppPublic/Login';
  }
```

And, if you are not sure of having config:

```
  private isEnabledForRetry(error: AxiosError) {
    return error.config && error.config.url === 'WebAppPublic/Login';
  }
```

src/services/pure-di.tsx
```diff
@@ -108,6 +108,7 @@ export class AppCompositionRoot implements AppServices {
         new HttpDopplerLegacyClient({
           axiosStatic: this.axiosStatic,
           baseUrl: this.appConfiguration.dopplerLegacyUrl,
+          window: this.window,
```
andresmoschini: 👍

src/services/doppler-legacy-client.ts
cbernat: Changed.

cbernat:
Test this please

cbernat:
test this please

PR #712, opened 2020-01-30
feat: DW-188-login-bapro

cbernat:
Add more domains to whitelist for testing if needed for bapro enviroment, and other doppler enviroments, to allow use of redirect as we do in academy.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -151,6 +151,14 @@ export function isWhitelisted(url: string) {
     'https://prod.doppleracademy.com/',
     'https://doppleracademy.com/',
     'http://doppleracademy.com/',
+    'https://app2.bancoprovinciamail.com.ar/',
+    'https://app2.fromdoppler.com/',
+    'http://localhost:',
+    'http://localhost/',
+    'https://localhost:',
+    'https://localhost/',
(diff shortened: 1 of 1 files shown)


src/utils.ts
```diff
@@ -151,6 +151,8 @@ export function isWhitelisted(url: string) {
     'https://prod.doppleracademy.com/',
     'https://doppleracademy.com/',
     'http://doppleracademy.com/',
+    'https://app2.bancoprovinciamail.com.ar/',
+    'https://app2.fromdoppler.com/',
```
andresmoschini: If we change all our redirections to include the domain, I think that you will need to also add test environments: 
* `http://localhost/`, `http://localhost:` (I am not sure about the ports)
* `https://localhost/`, `https://localhost:` (I am not sure about the ports)
* `http://localhost` (I am not sure about the ports)
* `https://appint.fromdoppler.com/` 
* `https://appqa.fromdoppler.com/`
* https?

cbernat: I added `http://localhost` and `https://localhost` to allow more freedom (avoiding port)

andresmoschini: In that way, you are also allowing `http://localhost.malisioussite.com`

PR #717, opened 2020-02-04
fix(SubscriberHistory): redirect master subscriber 

fcoronelmakingsense:
- Redirect to master subscriber when you put a subscriber doesn't exist for this user


cbernat:
🚀

PR #721, opened 2020-02-06
fix(Environment): change protocol for api int and qa

fcoronelmakingsense:
- change http protocol to https for the doppler api


andresmoschini:
:shipit:

PR #727, opened 2020-02-10
fix(SubscriberHistory): pagination current page style

fcoronelmakingsense:
image

### DIFF
--- a/src/components/shared/Pagination/Pagination.js
+++ b/src/components/shared/Pagination/Pagination.js
@@ -9,7 +9,12 @@ export const Pagination = ({ currentPage, pagesCount, urlToGo }) => {
       <ul>
         {pageNumbers.map((number) => (
           <li key={number}>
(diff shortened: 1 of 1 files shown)


src/components/shared/Pagination/Pagination.js
```diff
@@ -9,7 +9,12 @@ export const Pagination = ({ currentPage, pagesCount, urlToGo }) => {
       <ul>
         {pageNumbers.map((number) => (
           <li key={number}>
-            <Link to={`${urlToGo}page=${number}`}>{number}</Link>
+            <Link
+              className={number === Number(currentPage) ? 'dp-active-page' : ''}
```
cbernat: Why `currentPage` is not a number at this point?

fcoronelmakingsense: because when pass from component is a string

PR #749, opened 2020-02-20
fix: set nginx config to only use TLSv1.2

elsupergomez:
This PR is for change the site config to force only accept TLSv1.2 in the web app.

The SSL Lab test right: https://www.ssllabs.com/ssltest/analyze.html?d=app.fromdoppler.com
and should be the same as https://www.ssllabs.com/ssltest/analyze.html?d=app2.fromdoppler.com

this change should solve this problem

### DIFF
--- a/conf/site-int.conf
+++ b/conf/site-int.conf
@@ -7,6 +7,7 @@ server{
         server_name             app.fromdoppler.com;
         ssl_certificate         /run/secrets/site.crt;
         ssl_certificate_key     /run/secrets/site.key;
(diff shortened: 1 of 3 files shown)


andresmoschini:
❤️

andresmoschini:
@elsupergomez could you amend your commit changing `chore` by `fix` in order to generate a new version and throw a release?

image

andresmoschini:
🚀

cbernat:
Working as expected 
image
Thanks @elsupergomez !

PR #763, opened 2020-02-28
fix: update build version

GusBaamonde:
- fix the icon in standby
- font hierarchy arrangement
- behavior icon alignment

fix-icon-sh-in-webapp


cbernat:
👍

PR #771, opened 2020-03-04
Dw-216 gdpr report with all permission fields for doppler user

cbernat:
Now we show all fields for the user, and when de subscriber never agreed nor rejected the permission we show a no response label. Showing all permissions here also allows a user to see the legal text (this cannot be seen if the permission is in use by at least one subscriber)

image

http://localhost:3000/reports/subscriber-gdpr?email=cbernat@makingsense.com

https://cdn.fromdoppler.com/doppler-webapp/build2205/#/reports/subscriber-gdpr?email=cbernat@makingsense.com

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -9,7 +9,9 @@ const Header = ({
   userData: { user, nav, alert, notifications, emptyNotificationText },
   location: { pathname },
 }) => {
-  const inactiveSection = pathname.match(/^\/integrations\/*/) !== null;
+  const inactiveSection =
+    pathname.match(/^\/integrations\/*/) !== null ||
+    pathname.match(/^\/reports\/subscriber-gdpr*/) !== null;
   return (
     <div>
       {alert ? <HeaderMessages alert={alert} user={user} /> : null}
--- a/src/components/Reports/GDPR/SubscriberGdpr.js
+++ b/src/components/Reports/GDPR/SubscriberGdpr.js
@@ -22,7 +22,8 @@ const SubscriberGdpr = ({ location, dependencies: { dopplerApiClient } }) => {
       setState({ loading: true });
       const email = replaceSpaceWithSigns(extractParameter(location, queryString.parse, 'email'));
       const responseSubscriber = await dopplerApiClient.getSubscriber(email);
-      if (responseSubscriber.success) {
+      const allFields = await dopplerApiClient.getUserFields();
+      if (responseSubscriber.success && allFields.success) {
         const subscriber = {
           firstName: responseSubscriber.value.fields.find((x) => x.name === 'FIRSTNAME'),
           lastName: responseSubscriber.value.fields.find((x) => x.name === 'LASTNAME'),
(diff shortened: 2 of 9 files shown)


cbernat:
Since these are trivial changes I will merge them. Feel free to make any comments later though.

GusBaamonde:
🙂  related to: https://github.com/FromDoppler/Doppler-UI-System/pull/262

src/components/Header/Header.js
```diff
@@ -9,7 +9,9 @@ const Header = ({
   userData: { user, nav, alert, notifications, emptyNotificationText },
   location: { pathname },
 }) => {
-  const inactiveSection = pathname.match(/^\/integrations\/*/) !== null;
+  const inactiveSection =
+    pathname.match(/^\/integrations\/*/) !== null ||
+    pathname.match(/^\/reports\/subscriber-gdpr*/) !== null;
```
fcoronelmakingsense: its temporal. i working in the a general solution

src/components/Reports/GDPR/SubscriberGdpr.js
```diff
@@ -22,20 +22,34 @@ const SubscriberGdpr = ({ location, dependencies: { dopplerApiClient } }) => {
       setState({ loading: true });
       const email = replaceSpaceWithSigns(extractParameter(location, queryString.parse, 'email'));
       const responseSubscriber = await dopplerApiClient.getSubscriber(email);
-      if (responseSubscriber.success) {
+      const allFields = await dopplerApiClient.getUserFields();
+      if (responseSubscriber.success && allFields.success) {
```
fcoronelmakingsense: how do you know when one fails who is it? or it doesnt matter?

cbernat: Here we need both, what do you suggest?

src/components/Reports/GDPR/SubscriberGdpr.js
```diff
@@ -22,20 +22,34 @@ const SubscriberGdpr = ({ location, dependencies: { dopplerApiClient } }) => {
...
+          (customField) => customField.type === 'permission' || customField.type === 'consent',
+        );
+
+        const fields = allPermissionFields.map((field) => {
+          const found = subscriberPermissionFields.find(
+            (subscriberField) => subscriberField.name === field.name,
+          );
+          field.value = found ? found.value : 'none';
+          return field;
+        });
```
fcoronelmakingsense: dont know if works but is another option :)
` return subscriberPermissionFields.find(
            (subscriberField) => subscriberField.name === field.name,
          ).value || 'none' ;

src/components/Reports/GDPR/SubscriberGdpr.test.js
```diff
@@ -180,4 +204,27 @@ describe('SubscriberGdpr report component', () => {
...
+          <SubscriberGdpr />
+        </IntlProvider>
+      </AppServicesProvider>,
+    );
+
+    await waitForDomChange();
+    // Assert
+    // expect to have at least one row in the grid
+    const tableNode = getByText('subscriber_gdpr.permission_name').closest('table');
+    //expect(document.querySelectorAll('tbody tr').length).toBe(subscriberFields.length);
```
fcoronelmakingsense: why comment the expect?

src/services/doppler-api-client.test.ts
```diff
@@ -475,4 +475,58 @@ describe('HttpDopplerApiClient', () => {
...
+      const dopplerApiClient = createHttpDopplerApiClient({ request });
+
+      // Act
+      const result = await dopplerApiClient.getUserFields();
+
+      // Assert
+      expect(request).toBeCalledTimes(1);
+      expect(result).not.toBe(undefined);
+      expect(result.success).toBe(true);
+      expect(result.value).not.toBe(null);
```
fcoronelmakingsense: you could add more checks for the correct mapping

src/components/Reports/GDPR/SubscriberGdpr.test.js
cbernat: I forgot about this thanks!

src/components/Reports/GDPR/SubscriberGdpr.js
fcoronelmakingsense: but you show the same message for both cases?

cbernat: Yes. Else it has no meaning to show incomplete information.

src/components/Reports/GDPR/SubscriberGdpr.js
cbernat: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find array.find returns `undefined` if no item is found.

src/services/doppler-api-client.test.ts
cbernat: Correct mapping was already tested in fields mapping for report for `subscriber-history`

PR #778, opened 2020-03-09
fix(Header): selected menu option with differents urls

fcoronelmakingsense:
- Refactor functionality of get the current selected menu option depend of the current url

Shopify
image

Reports
image

Subscriber History 
image

Master Subscriber 
image

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -4,22 +4,48 @@ import HeaderMessages from './HeaderMessages/HeaderMessages';
 import HeaderUserMenu from './HeaderUserMenu/HeaderUserMenu';
 import { FormattedMessage } from 'react-intl';
 import Notifications from './Notifications';
+import { getCurrentPageForUrl } from '../../utils';
+
+const getUpdatedSubNav = (currentSubNav, subMenuItem) => {
+  return currentSubNav.map((item) => {
+    return {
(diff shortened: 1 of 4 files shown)


src/components/Header/Header.js
```diff
@@ -4,22 +4,48 @@ import HeaderMessages from './HeaderMessages/HeaderMessages';
...
+      title: item.title,
+      url: item.url,
+      isSelected: item.idHTML === subMenuItem,
+    };
+  });
+};
+
+const getUpdateMenu = (currentUrl, nav) => {
+  const currentPage = getCurrentPageForUrl(currentUrl);
+  if (currentPage && currentPage.menu) {
```
cbernat: If the `currentPage` is not defined what would be the outcome? Reports --> advanced selected menu? How about having a fallback to avoid this? We would want the menu to be neutral in a not considered scenario.

cbernat: Ignore this comment, we will want to have the data from BE else we defined otherwise (because the item is not in the header, example: shopify)

fcoronelmakingsense: yes you can check http://cdn.fromdoppler.com/doppler-webapp/build2209/#/reports/master-subscriber

cbernat:
🚀 Green light!

PR #779, opened 2020-03-11
fix(Login): show promotions depend of the page parameter

fcoronelmakingsense:
image

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -19,12 +19,17 @@ import { FormattedMessageMarkdown } from '../../i18n/FormattedMessageMarkdown';
 import { connect } from 'formik';
 import Promotions from '../shared/Promotions/Promotions';
 import queryString from 'query-string';
-import { addLogEntry } from '../../utils';
+import { addLogEntry, extractParameter } from '../../utils';
 
 const fieldNames = {
   user: 'user',
   password: 'password',
 };
+
+/** Extract the page parameter from url*/
+function extractPage(location) {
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -33,7 +33,7 @@ const forcedServices =
   process.env.NODE_ENV === 'development' || process.env.REACT_APP_IS_DEMO_ENABLED
     ? {
         dopplerLegacyClient: new HardcodedDopplerLegacyClient(),
-        dopplerSitesClient: new HardcodedDopplerSitesClient(),
+        //dopplerSitesClient: new HardcodedDopplerSitesClient(),
```
cbernat: Here you need to remove the comment for it to work in demo and dev right?

andresmoschini: is it fine?

src/components/Login/Login.js
```diff
@@ -261,7 +266,7 @@ const Login = ({ location, dependencies: { dopplerLegacyClient, sessionManager,
           </p>
         </footer>
       </article>
-      <Promotions type="login" />
+      <Promotions type="login" page={extractPage(location)} />
```
andresmoschini: ❤️ 🎉

cbernat:
Nice!

src/index.js
fcoronelmakingsense: i correct, thanks

fcoronelmakingsense: there is wrong! thanks for the advice

PR #781, opened 2020-03-13
fix(Header): inactive section fallback

fcoronelmakingsense:
- fix header when dont have menu for the url.

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -38,7 +38,8 @@ const Header = ({
   location: { pathname },
 }) => {
   const updatedNav = getUpdateMenu(pathname, nav);
-  const isInactiveSection = !getCurrentPageForUrl(pathname).menu;
+  let isInactiveSection = getCurrentPageForUrl(pathname);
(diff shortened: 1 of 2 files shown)


cbernat:
👍

cbernat:
How about adding some tests about this?

src/components/Header/Header.js
```diff
@@ -38,7 +38,8 @@ const Header = ({
   location: { pathname },
 }) => {
   const updatedNav = getUpdateMenu(pathname, nav);
-  const isInactiveSection = !getCurrentPageForUrl(pathname).menu;
+  let isInactiveSection = getCurrentPageForUrl(pathname);
+  isInactiveSection = !isInactiveSection || !isInactiveSection.menu;
```
lucianaDilizio: What about:
```
let currentPageforUrl = getCurrentPageForUrl(pathname); 
let isInactiveSection = !currentPageforUrl || !currentPageforUrl .menu;
```
?
I mean, to be clearer.

PR #785, opened 2020-03-16
fix(SubscriberHistory): change reports url depend of the environment

fcoronelmakingsense:
- Use env variables to get the de domain for reports
- Change in header variable name to more clear

image

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -7,3 +7,4 @@ REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 REACT_APP_SHOPIFY_URL= https://sfy.fromdoppler.com:4433
 REACT_APP_DOPPLER_API_URL= http://newapiqa.fromdoppler.net
+REACT_APP_REPORTS_URL= https://reportsint.fromdoppler.net:400
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -38,8 +38,8 @@ const Header = ({
   location: { pathname },
 }) => {
   const updatedNav = getUpdateMenu(pathname, nav);
-  let isInactiveSection = getCurrentPageForUrl(pathname);
-  isInactiveSection = !isInactiveSection || !isInactiveSection.menu;
+  const currentPage = getCurrentPageForUrl(pathname);
+  const isInactiveSection = !currentPage || !currentPage.menu;
   return (
     <div>
       {alert ? <HeaderMessages alert={alert} user={user} /> : null}
--- a/src/components/Reports/SubscriberHistory/SubscriberHistory.js
+++ b/src/components/Reports/SubscriberHistory/SubscriberHistory.js
@@ -35,7 +35,13 @@ const getDeliveryStatusCssClassName = (deliveryStatus) => {
 
 const campaignsPerPage = 10;
 
-const SubscriberHistory = ({ location, dependencies: { dopplerApiClient } }) => {
+const SubscriberHistory = ({
+  location,
+  dependencies: {
+    dopplerApiClient,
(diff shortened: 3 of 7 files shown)


src/components/Header/Header.js
```diff
@@ -38,8 +38,8 @@ const Header = ({
   location: { pathname },
 }) => {
   const updatedNav = getUpdateMenu(pathname, nav);
-  let isInactiveSection = getCurrentPageForUrl(pathname);
-  isInactiveSection = !isInactiveSection || !isInactiveSection.menu;
+  const currentPage = getCurrentPageForUrl(pathname);
+  const isInactiveSection = !currentPage || !currentPage.menu;
```
andresmoschini: ❤️

src/components/Reports/SubscriberHistory/SubscriberHistory.js
```diff
@@ -217,7 +217,7 @@ const SubscriberHistory = ({ location, dependencies: { dopplerApiClient } }) =>
                               {campaign.urlImgPreview ? (
                                 <div className="dp-tooltip-container">
                                   <a
-                                    href={`https://reports2.fromdoppler.com/Dashboard.aspx?idCampaign=${campaign.campaignId}`}
+                                    href={`${process.env.REACT_APP_REPORTS_URL}/Dashboard.aspx?idCampaign=${campaign.campaignId}`}
```
andresmoschini: It would be better to use dependency injection for this kind of things.

By the moment, it looks the same, but eventually (not so far), we will need to refactor this kind of things and it will be a mess.

fcoronelmakingsense: how is it? for understand :)

andresmoschini: Well, you can check for other usages for `process.env...`

If it is not enough to make the idea clear, please call me.

src/components/Header/Header.js
lucianaDilizio: Nice, thanks! 😄

src/components/Reports/SubscriberHistory/SubscriberHistory.js
fcoronelmakingsense: fixed

andresmoschini: Great! it was simpler than I though 😀

andresmoschini:
🚀

src/components/Reports/SubscriberHistory/SubscriberHistory.js
```diff
@@ -217,7 +223,7 @@ const SubscriberHistory = ({ location, dependencies: { dopplerApiClient } }) =>
                               {campaign.urlImgPreview ? (
                                 <div className="dp-tooltip-container">
                                   <a
-                                    href={`https://reports2.fromdoppler.com/Dashboard.aspx?idCampaign=${campaign.campaignId}`}
+                                    href={`${reportsUrl}/Dashboard.aspx?idCampaign=${campaign.campaignId}`}
```
lucianaDilizio: ❤️

.env.development
```diff
@@ -7,3 +7,4 @@ REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 REACT_APP_SHOPIFY_URL= https://sfy.fromdoppler.com:4433
 REACT_APP_DOPPLER_API_URL= http://newapiqa.fromdoppler.net
+REACT_APP_REPORTS_URL= https://reportsint.fromdoppler.net:400
```
cbernat: Here I would leave doppler local configuration in case we need it. Just a thought 😄

PR #786, opened 2020-03-17
fix(SubscriberHistory): preview behind header

fcoronelmakingsense:
- Add some margin top to fix the first preview in grid sometimes put behind the header

image


GusBaamonde:
I think that all screens should have a gray header with the title, to have consistency like this in the mocks, https://projects.invisionapp.com/d/main#/console/18626800/399167847/preview  Otherwise we could ask Manu

fcoronelmakingsense:
updated!
image

PR #787, opened 2020-03-19
fix(SubscriberGdpr): permission styles in grid

fcoronelmakingsense:
- fix styles when have more than one permission.

image

### DIFF
--- a/src/components/Reports/GDPR/SubscriberGdpr.styles.js
+++ b/src/components/Reports/GDPR/SubscriberGdpr.styles.js
@@ -1,0 +1,8 @@
+import styled from 'styled-components';
+import fonts from '../../styles/fonts';
+
+export const TextColumn = styled.div`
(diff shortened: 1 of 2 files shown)


src/components/Reports/GDPR/SubscriberGdpr.styles.js
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components';
+import fonts from '../../styles/fonts';
+
+export const TextColumn = styled.div`
+  p {
+    font-size: ${fonts.fontLvl3};
```
GusBaamonde: looks great, i would just add a line-height: 18px; so it remains like the rest

fcoronelmakingsense: it's has now a line-height with that value

PR #788, opened 2020-03-19
fix(SubscriberGdpr): add missing breadcrumb

fcoronelmakingsense:
image

### DIFF
--- a/src/components/Reports/GDPR/SubscriberGdpr.js
+++ b/src/components/Reports/GDPR/SubscriberGdpr.js
@@ -17,6 +17,19 @@ const SubscriberGdpr = ({ location, dependencies: { dopplerApiClient } }) => {
   const _ = (id, values) => intl.formatMessage({ id: id }, values);
   const [state, setState] = useState({ loading: true });
 
+  const Breadcrumb = () => (
+    <nav className="dp-breadcrumb">
+      <ul>
+        <li>
+          <a href={_('subscriber_history.subscriber_breadcrumb_url')}>
+            {_('subscriber_history.subscriber_breadcrumb')}
+          </a>
+        </li>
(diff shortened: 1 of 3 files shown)


andresmoschini:
👍

src/components/Reports/GDPR/SubscriberGdpr.js
```diff
@@ -17,6 +17,19 @@ const SubscriberGdpr = ({ location, dependencies: { dopplerApiClient } }) => {
...
+      <ul>
+        <li>
+          <a href={_('subscriber_history.subscriber_breadcrumb_url')}>
+            {_('subscriber_history.subscriber_breadcrumb')}
+          </a>
+        </li>
+        <li>{_('subscriber_gdpr.gpdr_state_breadcrumb')}</li>
+      </ul>
+    </nav>
+  );
```
lucianaDilizio: Isn't any way to put this in one place and export it?

fcoronelmakingsense: i dont know if we can do a general component, because the texts and links changes. What you propose?

lucianaDilizio: Could't you pass that information by parameters? Like HeaderNav element for example

andresmoschini: Sorry, I have no enough context, @lucianaDilizio what is the repeated structure?

lucianaDilizio: I think that the breadcrumb is something common that we probably use in other places, so I just asked if there are any way to reuse this code for futures cases.

fcoronelmakingsense: yes maybe after we can think how can create reusable component

PR #790, opened 2020-03-20
fix(confirmation page): the relative position of the element was remo…

GusBaamonde:
- Update build version
- the relative position of the element was removed for better positioning of the backgrounds in library UI

related to: https://github.com/FromDoppler/Doppler-UI-System/pull/270

bug-thanks


andresmoschini:
Did you try in WebApp CDN against production?

In that case, :shipit:

PR #791, opened 2020-03-20
fix(Login): check redirect parameter behaviour like page parameter

fcoronelmakingsense:
- When has `redirect=https://doppleracademy.com` in the url call `getBanner` with the `page=doppler-academy` parameter

image

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -26,9 +26,17 @@ const fieldNames = {
   password: 'password',
 };
 
+function extractPageFromRedirect(location) {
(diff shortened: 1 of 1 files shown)


src/components/Login/Login.js
```diff
@@ -26,9 +26,14 @@ const fieldNames = {
   password: 'password',
 };
 
+function extractRedirect(location) {
```
andresmoschini: What do you think about 

```suggestion
function extractPageFromRedirect(location) {
```

You can also add a comment with the explanation that it is a patch, the reason, etc.

fcoronelmakingsense: fixed

PR #805, opened 2020-03-27
Add flag in phone input based in IP country

lucianaDilizio:
Show flag in phone element based on user ip using ipinfo.io

### DIFF
--- a/src/components/Signup/Signup.test.js
+++ b/src/components/Signup/Signup.test.js
@@ -16,6 +16,12 @@ const defaultDependencies = {
       return emptyResponse;
     },
   },
+  ipinfoClient: {
+    getCountryCode: async () => {
+      await timeout(0);
+      return 'PR';
+    },
+  },
 };
 
 describe('Signup', () => {
@@ -136,6 +142,28 @@ describe('Signup', () => {
     // Assert
     await waitFor(() =>
       expect(container.querySelector('#iti-item-io').nextElementSibling.id).toBe('iti-item-bn'),
+    );
+  });
+
+  it('should render flag based in ipinfoClient result', async () => {
+    // Arrange
+    const dependencies = defaultDependencies;
+    const location = { search: 'test', pathname: '/signup' };
+
+    // Act
+    const { container } = render(
+      <AppServicesProvider forcedServices={dependencies}>
+        <DopplerIntlProvider locale="en">
+          <Router>
+            <Signup location={location} />
+          </Router>
+        </DopplerIntlProvider>
+      </AppServicesProvider>,
+    );
+
+    // Assert
+    await waitFor(() =>
+      expect(container.querySelector('.iti__selected-flag').title).toBe('Puerto Rico: +1'),
     );
   });
 });
--- a/src/components/form-helpers/form-helpers.js
+++ b/src/components/form-helpers/form-helpers.js
@@ -20,6 +20,7 @@ import './form-helpers.css';
 import 'intl-tel-input/build/css/intlTelInput.min.css';
 import { useCaptcha } from './captcha-utils';
 import { FormattedMessageMarkdown } from '../../i18n/FormattedMessageMarkdown';
+import { InjectAppServices } from '../../services/pure-di';
 
 function concatClasses(...args) {
   return args.filter((x) => x).join(' ');
@@ -273,6 +274,7 @@ const _formatFieldValueAsInternationalNumber = (iti, fieldName, setFieldValue) =
  * @param { string } props.label - label
  * @param { string } props.placeholder - placeholder
  * @param { React.MutableRefObject<import('intl-tel-input').Plugin> } props.intlTelInputRef - intlTelInputRef
+ * @param { import('../../services/pure-di').AppServices } props.dependencies
  */
 const _PhoneFieldItem = ({
   className,
@@ -281,6 +283,7 @@ const _PhoneFieldItem = ({
   placeholder,
   required,
   formik: { values, handleChange, handleBlur, setFieldValue },
+  dependencies: { ipinfoClient },
   ...rest
 }) => {
   const intl = useIntl();
@@ -319,9 +322,9 @@ const _PhoneFieldItem = ({
       autoPlaceholder: 'aggressive',
       preferredCountries: ['ar', 'mx', 'co', 'es', 'ec', 'cl', 'pe', 'us'],
       initialCountry: 'auto',
-      geoIpLookup: (callback) => {
-        // TODO: determine current country using geolocation
-        callback('ar');
+      geoIpLookup: async (success) => {
+        const countryCode = await ipinfoClient.getCountryCode();
+        success(countryCode);
       },
     });
     intlTelInputRef.current = iti;
@@ -330,7 +333,7 @@ const _PhoneFieldItem = ({
       setEventListenerSet(false);
       iti.destroy();
     };
-  }, [intl.locale, fieldName, setFieldValue]);
+  }, [intl.locale, fieldName, setFieldValue, ipinfoClient]);
 
   if (!eventListenerSet && inputElRef.current && intlTelInputRef.current) {
     inputElRef.current.addEventListener('countrychange', handleChange);
@@ -359,7 +362,7 @@ const _PhoneFieldItem = ({
   );
 };
 
-export const PhoneFieldItem = connect(_PhoneFieldItem);
+export const PhoneFieldItem = connect(InjectAppServices(_PhoneFieldItem));
 
 export const InputFieldItem = ({
   className,
--- a/src/services/ipinfo-client.doubles.ts
+++ b/src/services/ipinfo-client.doubles.ts
@@ -1,0 +1,11 @@
+import { IpinfoClient } from './ipinfo-client';
+import { timeout } from '../utils';
+
+export const fakeResponse = 'AR';
+
+export class HardcodedIpinfoClient implements IpinfoClient {
+  public async getCountryCode(): Promise<string> {
+    await timeout(1500);
+    return fakeResponse;
+  }
+}
--- a/src/services/ipinfo-client.test.ts
+++ b/src/services/ipinfo-client.test.ts
@@ -1,0 +1,89 @@
+import { AxiosStatic } from 'axios';
+import { HttpIpinfoClient } from './ipinfo-client';
+import { exception } from 'react-ga';
+
+function createHttpIpinfoClient(axios: any) {
+  const axiosStatic = {
+    create: () => axios,
+  } as AxiosStatic;
+  const ipinfoClient = new HttpIpinfoClient({
+    axiosStatic,
+  });
+  return ipinfoClient;
+}
+
+describe('HttpIpinfoClient', () => {
+  it('should return the received country code', async () => {
+    // Arrange
+    const countryCode = 'FR';
+    const request = jest.fn(async () => ({ data: { country: countryCode } }));
+    const ipinfoClient = createHttpIpinfoClient({ request });
+
+    // Act
+    const result = await ipinfoClient.getCountryCode();
+
+    // Assert
+    expect(request).toBeCalledTimes(1);
+    expect(result).toBe('FR');
+  });
+
+  it('should return AR if the country code is null', async () => {
+    // Arrange
+    const countryCode = null;
+    const request = jest.fn(async () => ({ data: { country: countryCode } }));
+    const ipinfoClient = createHttpIpinfoClient({ request });
+
+    // Act
+    const result = await ipinfoClient.getCountryCode();
+
+    // Assert
+    expect(request).toBeCalledTimes(1);
+    expect(result).toBe('AR');
+  });
+
+  it('should return AR if the response is unexpected', async () => {
+    // Arrange
+    const request = jest.fn(async () => ({ data: {} }));
+    const ipinfoClient = createHttpIpinfoClient({ request });
+
+    // Act
+    const result = await ipinfoClient.getCountryCode();
+
+    // Assert
+    expect(request).toBeCalledTimes(1);
+    expect(result).toBe('AR');
+  });
+
+  it('should return AR if the response is an exception', async () => {
+    // Arrange
+    const request = jest.fn(async () => {
+      throw new Error();
+    });
+    const ipinfoClient = createHttpIpinfoClient({ request });
+
+    // Act
+    const result = await ipinfoClient.getCountryCode();
+
+    // Assert
+    expect(request).toBeCalledTimes(1);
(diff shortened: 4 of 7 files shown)


andresmoschini:
Welcome to WebApp! 😀

image

cbernat:
@lucianaDilizio you can configure prettier to run automatically on editor save, to forget about running manually each time.

src/components/form-helpers/form-helpers.js
```diff
@@ -320,9 +322,16 @@ const _PhoneFieldItem = ({
       autoPlaceholder: 'aggressive',
       preferredCountries: ['ar', 'mx', 'co', 'es', 'ec', 'cl', 'pe', 'us'],
       initialCountry: 'auto',
-      geoIpLookup: (callback) => {
-        // TODO: determine current country using geolocation
-        callback('ar');
+      geoIpLookup: async (success, failure) => { // TODO: read cache before request
+        try {
+          const response = await axios.get('https://ipinfo.io/json');
+          var countryCode = (response && response.country) ? response.country : "";
```
cbernat: We don't use `var` anymore. 
```suggestion
          const countryCode = (response && response.country) ? response.country : '';
```

src/components/form-helpers/form-helpers.js
```diff
@@ -320,9 +321,17 @@ const _PhoneFieldItem = ({
...
       preferredCountries: ['ar', 'mx', 'co', 'es', 'ec', 'cl', 'pe', 'us'],
       initialCountry: 'auto',
-      geoIpLookup: (callback) => {
-        // TODO: determine current country using geolocation
-        callback('ar');
+      geoIpLookup: async (success, failure) => {
+        // TODO: read cache before request
+        try {
+          const response = await axios.get('https://ipinfo.io/json');
+          var countryCode = response && response.country ? response.country : '';
```
andresmoschini: Oh, it seems that our CI process does not take care of this :disappointed: 

We should have to have special attention to this (and fix our CI).

```suggestion
          const countryCode = response && response.country ? response.country : '';
```

andresmoschini: Also, if I am not wrong, we could take advantage of new ES features (we should verify the generated code):

```suggestion
          const countryCode = response?.country ?? '';
```

andresmoschini: But... why would you want an empty string in `countryCode` 😕?

lucianaDilizio: In this case I think that we should be return AR by default

src/components/form-helpers/form-helpers.js
```diff
@@ -267,22 +267,25 @@ const _formatFieldValueAsInternationalNumber = (iti, fieldName, setFieldValue) =
...
- * @param { import('react-intl').InjectedIntl } props.intl - intl
- * @param { import('formik').FormikProps<Values> } props.formik - formik
- * @param { string } props.className - className
- * @param { string } props.fieldName - fieldName
- * @param { string } props.label - label
- * @param { string } props.placeholder - placeholder
- * @param { React.MutableRefObject<import('intl-tel-input').Plugin> } props.intlTelInputRef - intlTelInputRef
- */
+
+  * @param { Object } props - props
```
andresmoschini, line 271:
```suggestion
  * @param { Object } props - props
```

andresmoschini: or maybe

```suggestion
  *
  * @param { Object } props - props
```

Please, check JSDoc documentation to use the right format

src/components/form-helpers/form-helpers.js
```diff
@@ -267,22 +267,25 @@ const _formatFieldValueAsInternationalNumber = (iti, fieldName, setFieldValue) =
...
+  * @param { import('react-intl').InjectedIntl } props.intl - intl
+  * @param { import('formik').FormikProps<Values> } props.formik - formik
+  * @param { string } props.className - className
+  * @param { string } props.fieldName - fieldName
+  * @param { string } props.label - label
+  * @param { string } props.placeholder - placeholder
+  * @param { React.MutableRefObject<import('intl-tel-input').Plugin> } props.intlTelInputRef - intlTelInputRef
+  * @param { import('../../services/pure-di').AppServices } props.dependencies
```
andresmoschini, line 279: ❤️

src/components/form-helpers/form-helpers.js
```diff
@@ -321,17 +324,10 @@ const _PhoneFieldItem = ({
...
-          var countryCode = response.data && response.data.country ? response.data.country : 'AR';
-          success(countryCode);
-          console.log(response);
-        } catch (error) {
-          success('ES'); // TODO
-          console.log(error.toJSON());
-        }
+      geoIpLookup: async (success) => {
+        const response = await ipinfoClient.getCountryCode();
+        const countryCode = response.data && response.data.country ? response.data.country : 'AR';
```
andresmoschini, line 329: Could you move the request details inside the service? For example:

```suggestion
        const codeOrNull = await ipinfoClient.getCountryCode();
        const countryCode = codeOrNull || 'AR';
```

or better yet:

```suggestion
        const countryCode = await ipinfoClient.getCountryCode({ fallback: 'AR'});
```

src/services/ipinfo-client.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export interface IpinfoClient {
+  getCountryCode(): Promise<IpInfoResult>;
+}
+
+export class HttpIpinfoClient implements IpinfoClient {
+  public async getCountryCode(): Promise<IpInfoResult> {
+    try {
+      const response = await axios.get('http://ipinfo.io/json');
```
andresmoschini, line 14: `axios` should be injected as a dependency here.

src/services/ipinfo-client.ts
```diff
@@ -0,0 +1,21 @@
...
+      const countryCode = response.data && response.data.country ? response.data.country : 'AR';
+      return { data: { country: countryCode } };
+    } catch (error) {
+      return { data: { country: 'AR' } };
```
andresmoschini, line 18: Oh, you are already dealing with the fallback here, maybe you can simplify the result format.

src/services/ipinfo-client.ts
lucianaDilizio: I injected it at the top of the file, should I put it here instead of where I put it?

src/services/ipinfo-client.ts
```diff
@@ -0,0 +1,33 @@
...
+      const response = await this.axios.request({
+        method: 'GET',
+        url: 'http://ipinfo.io/json',
+      });
+      console.log(response);
+      const countryCode = response.data && response.data.country ? response.data.country : 'AR';
+      return { success: true, value: { countryCode: countryCode } };
+    } catch (error) {
+      return { success: false, value: { countryCode: defaultCountryCode } };
+    }
```
andresmoschini, line 31: Why not simply return `AR` if something goes wrong?

src/services/ipinfo-client.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  public async getCountryCode(): Promise<ResultWithoutExpectedErrors<IpinfoResult>> {
+    const defaultCountryCode = 'AR';
+    try {
+      const response = await this.axios.request({
+        method: 'GET',
+        url: 'http://ipinfo.io/json',
+      });
+      console.log(response);
```
andresmoschini, line 26: Remember to remove this.

src/services/ipinfo-client.ts
```diff
@@ -0,0 +1,33 @@
+import { AxiosInstance, AxiosStatic } from 'axios';
+import { ResultWithoutExpectedErrors } from '../doppler-types';
+
+export interface IpinfoResult {
+  countryCode?: string | null;
```
andresmoschini, line 5: If I am not wrong this value will be always defined and cannot be null.

src/services/ipinfo-client.ts
```diff
@@ -0,0 +1,33 @@
...
+}
+
+export interface IpinfoClient {
+  getCountryCode(): Promise<ResultWithoutExpectedErrors<IpinfoResult>>;
```
andresmoschini, line 9: In my opinion, in this case, _ResultWithoutExpectedErrors_ wrapper is not necessary.

```suggestion
  getCountryCode(): Promise<IpinfoResult>;
```

Or even:


```suggestion
  getCountryCode(): Promise<string>;
```

src/services/ipinfo-client.ts
lucianaDilizio: I don't understand... I have to return always an object like that... 🤔

src/components/Signup/Signup.test.js
```diff
@@ -138,4 +144,26 @@ describe('Signup', () => {
...
+          <Router>
+            <Signup location={location} />
+          </Router>
+        </DopplerIntlProvider>
+      </AppServicesProvider>,
+    );
+
+    // Assert
+    await waitFor(() =>
+      expect(container.querySelector('.iti__selected-flag').title).toBe('Puerto Rico: +1'),
```
fcoronelmakingsense, line 166: how do you know show `puerto rico : +1` ? maybe it's better add `ipinfoclient` in the test, what do you think?

src/components/form-helpers/form-helpers.js
```diff
@@ -281,6 +283,7 @@ const _PhoneFieldItem = ({
   placeholder,
   required,
   formik: { values, handleChange, handleBlur, setFieldValue },
+  dependencies: { ipinfoClient },
```
fcoronelmakingsense, line 286: the second `i` dont need to be in uppercase? like `ipInfoClient` ?

src/components/form-helpers/form-helpers.js
```diff
@@ -319,9 +322,9 @@ const _PhoneFieldItem = ({
       autoPlaceholder: 'aggressive',
       preferredCountries: ['ar', 'mx', 'co', 'es', 'ec', 'cl', 'pe', 'us'],
       initialCountry: 'auto',
-      geoIpLookup: (callback) => {
-        // TODO: determine current country using geolocation
-        callback('ar');
+      geoIpLookup: async (success) => {
+        const countryCode = await ipinfoClient.getCountryCode();
```
fcoronelmakingsense, line 326: what happen is fail the request?

lucianaDilizio: The request never fails, if ipinfo api returns something wrong by defautl I will return 'AR'

src/services/ipinfo-client.doubles.ts
```diff
@@ -0,0 +1,11 @@
+import { IpinfoClient } from './ipinfo-client';
+import { timeout } from '../utils';
+
+export const fakeResponse = 'AR';
```
fcoronelmakingsense, line 4: maybe don't need this constant if only use one time

src/services/ipinfo-client.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    const countryCode = 'FR';
+    const request = jest.fn(async () => ({ data: { country: countryCode } }));
+    const ipinfoClient = createHttpIpinfoClient({ request });
+
+    // Act
+    const result = await ipinfoClient.getCountryCode();
+
+    // Assert
+    expect(request).toBeCalledTimes(1);
+    expect(result).toBe(countryCode);
```
fcoronelmakingsense, line 27: maybe is better check with the string like `FR`

src/components/form-helpers/form-helpers.js
lucianaDilizio: I prefered keep it in lowercase because the api's domain es ipinfo.io

src/services/ipinfo-client.doubles.ts
lucianaDilizio: Maybe it's ok to specify that this will be the fake response. To do not return a string that we don't know what is.

src/components/Signup/Signup.test.js
lucianaDilizio: I didn't find any other attr to check that the shown country is the correct. What do you mean with add ipinfoclient?

src/services/ipinfo-client.test.ts
lucianaDilizio: Why? I already have a constant with that value. I think that is no necessary.

src/components/form-helpers/form-helpers.js
andresmoschini: > what happen is fail the request?

It never fails

src/components/Signup/Signup.test.js
cbernat: Perhaps you can have some variable, with a key and a country description to avoid misunderstandings like this:
```
const selectedCountry = {
  key: 'PR',
  description: 'Puerto Rico: +1'
};
```
and then
```
expect(container.querySelector('.iti__selected-flag').title).toBe(selectedCountry.description),
```

cbernat:
👍

src/components/Signup/Signup.test.js
fcoronelmakingsense: maybe it's better put the string in the expect, we use that way in most of the tests

andresmoschini: que tal usar tablas? https://jestjs.io/docs/en/api#2--describeeachtablename-fn-timeout-

```js
it.each`
  countryCode | expectedDescription
  ${'PR'}     | ${'Puerto Rico: +1'}
  ${'ES'}     | ${'España: +34'}
`('should render selected country based in ipinfoClient result', ({countryCode, expectedDescription}) => {
    // Arrange
    const dependencies = defaultDependencies;
    // . . .
```

src/services/ipinfo-client.test.ts
fcoronelmakingsense: if you see the other test we check with the strings. its more explicit. Maybe @andresmoschini know more about that :P

andresmoschini: I like this test with the constant, buet it is very opinable.

We can also use tables here, it will be better.

src/components/Signup/Signup.test.js
lucianaDilizio: I don't realize how can I use the countryCode in the hardcoded dependency ipinfoClient that returns the async response. I think that this won't work in this case.

andresmoschini: oh, and override dependencies does not work :(

lucianaDilizio: Exactly, so I think that by the moment I will use strings in the expects to be consistent with other test files as @fcoronelmakingsense said.

PR #819, opened 2020-04-03
fix(semanitc-release): fix script to build inside container

cbernat:
### Fix release version script
In order to avoid issues with jenkins workspace better try to run this into a docker container. Mapping .git folder seems to do the trick.

This approach then can be done in ui library as well. 😄


andresmoschini:
👍

cbernat:
test this please

PR #828, opened 2020-04-07
fix(LanguageSelector): maintain url parameters

fcoronelmakingsense:
- Refactor component to maintain de urls parameters. Remove `?` and `lang` parameter for url and then url is created again with the lang first and then the exist urls parameters

### DIFF
--- a/src/components/shared/LanguageSelector/LanguageSelector.js
+++ b/src/components/shared/LanguageSelector/LanguageSelector.js
@@ -18,7 +18,7 @@ const LanguageSelector = ({
   const lang = intl.locale;
   const languageDropdown = useRef(null);
   const langToChange = lang === 'en' ? 'es' : 'en';
-  urlParameters = urlParameters.replace(/(\?)?(&)?(lang=(en|es))?(&)?/g, '');
+  urlParameters = urlParameters.replace(/(\?)?((&lang=(en|es)|(lang=(en|es)(&)?)))?/g, '');
   useEffect(() => {
     if (Menubutton) {
       const menubutton = new Menubutton(languageDropdown.current);
--- a/src/components/shared/LanguageSelector/LanguageSelector.js
+++ b/src/components/shared/LanguageSelector/LanguageSelector.js
@@ -9,6 +9,7 @@ import { InjectAppServices } from '../../../services/pure-di';
  * @param { import('react-intl').InjectedIntl } props.intl
  */
 const LanguageSelector = ({
+  urlParameters,
   dependencies: {
     window: { Menubutton },
   },
@@ -16,7 +17,8 @@ const LanguageSelector = ({
   const intl = useIntl();
   const lang = intl.locale;
   const languageDropdown = useRef(null);
-
+  const langToChange = lang === 'en' ? 'es' : 'en';
+  urlParameters = urlParameters.replace(/(\?)?(&)?(lang=(en|es))?(&)?/g, '');
   useEffect(() => {
     if (Menubutton) {
       const menubutton = new Menubutton(languageDropdown.current);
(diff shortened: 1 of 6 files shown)


src/components/shared/LanguageSelector/LanguageSelector.js
```diff
@@ -9,14 +9,16 @@ import { InjectAppServices } from '../../../services/pure-di';
...
   dependencies: {
     window: { Menubutton },
   },
 }) => {
   const intl = useIntl();
   const lang = intl.locale;
   const languageDropdown = useRef(null);
-
+  const langToChange = lang === 'en' ? 'es' : 'en';
+  urlParameters = urlParameters.replace(/(\?)?(&)?(lang=(en|es))?(&)?/g, '');
```
cbernat, line 21: I find this very strange, but I can't think of a better way to do it. 😅

cbernat:
Perhaps you could add tests for language selector, since it's a bit more complex now right?

src/components/shared/LanguageSelector/LanguageSelector.js
fcoronelmakingsense: because we have many case to cover. so.. i delete the parameter first and then put again with all parameters

andresmoschini:
This code seems to be more complex than I like. 

Maybe there is not an alternative.

In that case, what do you think about adding tests?

src/components/shared/LanguageSelector/LanguageSelector.test.js
```diff
@@ -0,0 +1,66 @@
...
+import { Router } from 'react-router-dom';
+import { MemoryRouter } from 'react-router-dom';
+
+const defaultDependencies = {
+  window: {
+    MenuButton: {},
+  },
+};
+
+describe('Language Selector Component', () => {
```
andresmoschini, line 16: Is it possible testing this scenario?

1. Open a page with redirect parameter, for example: `/signup?redirect=test` (and `/signup?another&redirect=test`, and other weird cases)
2. Click in `ES` (or `EN`)
3. Verify if the new URL has the expected redirect parameter.

fcoronelmakingsense: i add more cases now

src/components/shared/LanguageSelector/LanguageSelector.test.js
```diff
@@ -62,5 +81,29 @@ describe('Language Selector Component', () => {
     );
     // Assert
     expect(getByText('ES')).toBeInTheDocument();
+    expect(container.querySelector('.lang--es')).toBeInTheDocument();
+    expect(container.querySelector('.lang--es').search).toEqual(
+      '?lang=es&test&redirect=test&test2',
+    );
+  });
+
+  it('should show ES and maintain all parameters when lang parameter is in the end', () => {
```
cbernat, line 90: How about adding a scenario for lang in the start of search, and lang in the middle. I' ve seen you tested this manually :trollface:

fcoronelmakingsense: haha i added now!

src/components/shared/LanguageSelector/LanguageSelector.test.js
```diff
@@ -0,0 +1,129 @@
...
+    );
+    // Assert
+    expect(getByText('ES')).toBeInTheDocument();
+    expect(container.querySelector('.lang--es')).toBeInTheDocument();
+    expect(container.querySelector('.lang--es').search).toEqual(
+      '?lang=es&test&redirect=test&test2',
+    );
+  });
+
+  it('should show ES and maintain all parameters when lang parameter is in the middle', () => {
```
cbernat, line 110: ❤️

cbernat:
👍

src/components/shared/LanguageSelector/LanguageSelector.test.js
andresmoschini: Si, pero no veo ningún caso donde haya navegación, es decir ese test que te pasé no está, no? o se me escapa algo?

fcoronelmakingsense: Perdon se me paso esto. Voy a investigar

PR #830, opened 2020-04-08
fix(App): act warnings in the tests

fcoronelmakingsense:
- Fix act warnings we had in `App.test`. I used this documentation to fix that:

https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

### DIFF
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -52,11 +52,9 @@ const RouterInspector = withRouter(({ match, location, history, target }) => {
 
 const emptyResponse = { success: false, error: new Error('Dummy error') };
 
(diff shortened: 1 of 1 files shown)


src/App.test.js
```diff
@@ -52,11 +52,9 @@ const RouterInspector = withRouter(({ match, location, history, target }) => {
 
 const emptyResponse = { success: false, error: new Error('Dummy error') };
 
+const promise = Promise.resolve({ success: false, value: '' });
```
andresmoschini, line 55:
```suggestion
const rejectedPromise = Promise.resolve({ success: false, value: '' });
```

fcoronelmakingsense: fixed

PR #831, opened 2020-04-09
fix(validations): email regex bucle problem

fcoronelmakingsense:
- We had problem in login and signup with the validation email. When put long text and validate with the actual regex the screen is frozen. I update the email regex with new one obtain in the Yup library its for validations too and is common use with formik.

https://github.com/jquense/yup/search?q=rEmail&unscoped_q=rEmail


cbernat:
👍

src/validations.ts
```diff
@@ -1,4 +1,5 @@
-const emailRegex = /^([a-zñ\d[\]])(\.?([\wñ&/~\-+\][]+))*@([ñ\w.\-\][]+)\.[\w\-\][.]{2,}(\?)?.*$/i;
+// eslint-disable-next-line
```
andresmoschini, line 1: Why?

fcoronelmakingsense: We saw with @cbernat there is infinite loop with the previews one and found one more new and works with more escenarios and cases

andresmoschini: I was talking about the disabled rule. What is the rule? Why does it cannot be fixed?

fcoronelmakingsense: in the library has too maybe becuse this warning ```
Unexpected control character(s) 
in regular expression: \x09, \x0d, \x0a, \x09,   
\x01, \x08, \x0b, \x0c, \x0e, \x1f, \x01, \x09,   \x0b, \x0c, \x0d, \x09, \x0d, \x0a, \x09
```

andresmoschini: mmm, and what do you think about those characters? is eslint wrong? do we really need them?

fcoronelmakingsense: if we want to support more cases yes,

PR #841, opened 2020-04-13
fix(LoginPages): add retry in forgot password and signup

fcoronelmakingsense:
- Add retry to signup page and forgot password. 
- Refactor to general functionality to add more urls in the future if we need more.

### DIFF
--- a/src/services/doppler-legacy-client.ts
+++ b/src/services/doppler-legacy-client.ts
@@ -312,8 +312,17 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
   private readonly baseUrl: string;
   private window: any;
 
+  private isUrlIncludedInRetryList(currentUrl: any): any {
+    const urls = [
+      'WebAppPublic/Login',
+      'WebAppPublic/CreateUser',
+      'WebAppPublic/SendResetPasswordEmail',
+    ];
+    return urls.includes(currentUrl);
+  }
+
   private isEnabledForRetry(error: AxiosError) {
-    return error.config && error.config.url === 'WebAppPublic/Login';
+    return error.config && this.isUrlIncludedInRetryList(error.config.url);
   }
 
   constructor({
@@ -341,7 +350,6 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
       retryDelay: (retryCount: any, error: any) => {
         addLogEntry({
           origin: window.location.origin,
-          section: 'Login/legacyClient',
           browser: window.navigator.userAgent,
           error: error,
           retryCount: retryCount,
@@ -535,7 +543,7 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
 
   public async sendResetPasswordEmail(model: ForgotPasswordModel): Promise<ForgotPasswordResult> {
     try {
-      const response = await this.axios.post('/WebAppPublic/SendResetPasswordEmail', {
+      const response = await this.axios.post('WebAppPublic/SendResetPasswordEmail', {
         Email: model.email,
         RecaptchaUserCode: model.captchaResponseToken,
       });


src/utils.ts
```diff
@@ -203,3 +203,12 @@ const urlsWebApp: any = [
...
 }
+
+export function findUrlInRetryList(currentUrl: any): any {
+  const urls = [
+    'WebAppPublic/Login',
+    'WebAppPublic/CreateUser',
+    '/WebAppPublic/SendResetPasswordEmail',
+  ];
+  return urls.find((item: string) => item === currentUrl);
+}
```
andresmoschini, line 214: I think that these URLs are very tied to legacy Doppler, what do you think about keep it inside doppler-legacy-client.ts?

src/utils.ts
```diff
@@ -203,3 +203,12 @@ const urlsWebApp: any = [
 export function getCurrentPageForUrl(currentUrl: string): any {
   return urlsWebApp.find((item: any) => item.url === currentUrl);
 }
+
+export function findUrlInRetryList(currentUrl: any): any {
+  const urls = [
+    'WebAppPublic/Login',
+    'WebAppPublic/CreateUser',
+    '/WebAppPublic/SendResetPasswordEmail',
```
andresmoschini, line 211: Why this one has a different format?

src/utils.ts
```diff
@@ -203,3 +203,12 @@ const urlsWebApp: any = [
 export function getCurrentPageForUrl(currentUrl: string): any {
   return urlsWebApp.find((item: any) => item.url === currentUrl);
 }
+
+export function findUrlInRetryList(currentUrl: any): any {
```
lucianaDilizio, line 207: `(currentUrl: any)` why not `string`?

fcoronelmakingsense: can be null

src/utils.ts
fcoronelmakingsense: because in doppler legacy has the `/` maybe need delete both sides

andresmoschini: And it is different in the other cases?

src/utils.ts
andresmoschini: In that case:

```suggestion
export function  isUrlInRetryList(currentUrl: string | null): bool {
```

src/utils.ts
fcoronelmakingsense: yes other cases has the `/` too. maybe after can add refactor to all urls

src/utils.ts
fcoronelmakingsense: but if find the url return the element dont a boolean

andresmoschini: > but if find the url return the element dont a boolean

I know that you can fix that :P

As I understand, you do not want to find a URL, you want to verify if it is in the list, for that reason I purpose to change the name of the function and the result type.

src/services/doppler-legacy-client.ts
```diff
@@ -341,7 +350,6 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
       retryDelay: (retryCount: any, error: any) => {
         addLogEntry({
           origin: window.location.origin,
-          section: 'Login/legacyClient',
```
andresmoschini, line 344: Is it not necessary anymore?

src/services/doppler-legacy-client.ts
```diff
@@ -312,8 +312,17 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
...
   private window: any;
 
+  private findUrlInRetryList(currentUrl: any): any {
+    const urls = [
+      'WebAppPublic/Login',
+      'WebAppPublic/CreateUser',
+      'WebAppPublic/SendResetPasswordEmail',
+    ];
+    return urls.find((item: string) => item === currentUrl);
+  }
```
andresmoschini, line 322:
```suggestion
  private findUrlInRetryList(currentUrl: string | null): bool {
    const urls = [
      'WebAppPublic/Login',
      'WebAppPublic/CreateUser',
      'WebAppPublic/SendResetPasswordEmail',
    ];
    return urls.includes(currentUrl);
  }
```

src/services/doppler-legacy-client.ts
cbernat: We cannot have all logs with the same section, either remove it or replace it. Also this informatiton is a bit redundant, see in the log:

image

Here we have section and then in request information we have also the url.

src/utils.ts
fcoronelmakingsense: fixed

src/utils.ts
fcoronelmakingsense: fixed

src/services/doppler-legacy-client.ts
fcoronelmakingsense: fixed

PR #843, opened 2020-04-15
Fix https issue with ipinfo.io

lucianaDilizio:
It wasn't working because I was using http instead https to call ipinfo.io API.

### DIFF
--- a/src/services/ipinfo-client.test.ts
+++ b/src/services/ipinfo-client.test.ts
@@ -83,7 +83,7 @@ describe('HttpIpinfoClient', () => {
     expect(result).toBe(countryCode);
     expect(request).toHaveBeenCalledWith({
       method: 'GET',
-      url: 'http://ipinfo.io/json',
+      url: 'https://ipinfo.io/json',
     });
   });
 });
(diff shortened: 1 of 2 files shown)


src/services/ipinfo-client.test.ts
```diff
@@ -83,7 +83,7 @@ describe('HttpIpinfoClient', () => {
     expect(result).toBe(countryCode);
     expect(request).toHaveBeenCalledWith({
       method: 'GET',
-      url: 'http://ipinfo.io/json',
+      url: 'https://ipinfo.io/json',
```
cbernat, line 86: ❤️

cbernat:
You can test this in prod like this https://cdn.fromdoppler.com/doppler-webapp/build2297/#/signup
int -->https://cdn.fromdoppler.com/doppler-webapp/int-build2297/#/signup
qa --> https://cdn.fromdoppler.com/doppler-webapp/qa-build2297/#/signup
demo(as local but on cdn) -->https://cdn.fromdoppler.com/doppler-webapp/demo-build2297/#/signup

cbernat:
You can check that is working in prod 

image

andresmoschini:
🚀

PR #846, opened 2020-04-17
fix(Login): remove white spaces before send to BE

fcoronelmakingsense:
- add before submit in login a trim function to remove white spaces in user name

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -113,7 +113,7 @@ const Login = ({ location, dependencies: { dopplerLegacyClient, sessionManager, 
   const onSubmit = async (values, { setSubmitting, setErrors }) => {
     try {
       const result = await dopplerLegacyClient.login({
-        username: values[fieldNames.user],
+        username: values[fieldNames.user].trim(),
         password: values[fieldNames.password],
         captchaResponseToken: values['captchaResponseToken'],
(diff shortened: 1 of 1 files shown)


src/components/Login/Login.js
```diff
@@ -113,7 +113,7 @@ const Login = ({ location, dependencies: { dopplerLegacyClient, sessionManager,
   const onSubmit = async (values, { setSubmitting, setErrors }) => {
     try {
       const result = await dopplerLegacyClient.login({
-        username: values[fieldNames.user],
+        username: values[fieldNames.user].trim(),
```
cbernat, line 116: @andresmoschini I summon you, any better idea than doing this? back end is taking litteraly what we send, if we send spaces it takes them as part of the email. In old login the trimming was done in the view.

cbernat:
If we don't have any opinions yet, I think it's better to merge. 🚀

fcoronelmakingsense:
testing here http://cdn.fromdoppler.com/doppler-webapp/build2304/#/login

src/components/Login/Login.js
andresmoschini: Sorry, I did not saw this one. 

I think that it is fine in this way, it cannot be _null_, right?

PR #857, opened 2020-04-27
feat(ReportsNew): add new reports page

fcoronelmakingsense:
- Add new reports page empty.
- Add test to check load ok.
- Add page in new features.

image

image

### DIFF
--- a/src/components/Reports/ReportsNew.js
+++ b/src/components/Reports/ReportsNew.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import { Helmet } from 'react-helmet';
+
+const ReportsNew = () => {
+  return (
+    <>
+      <FormattedMessage id="reports_title">
+        {(reports_title) => (
+          <Helmet>
+            <title>{reports_title}</title>
+          </Helmet>
+        )}
+      </FormattedMessage>
+      <section className="dp-container m-t-24">
+        <FormattedMessage id="reports_title" />
+      </section>
+    </>
+  );
+};
+
+export default ReportsNew;
--- a/src/components/Reports/ReportsNew.test.js
+++ b/src/components/Reports/ReportsNew.test.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import DopplerIntlProvider from '../../i18n/DopplerIntlProvider.double-with-ids-as-values';
+import Reports from './ReportsNew';
+
+describe('Reports New page', () => {
+  afterEach(cleanup);
+
+  it('render page', () => {
+    // Arrange
+    // Act
+    const { getByText } = render(
+      <DopplerIntlProvider>
+        <Reports />
+      </DopplerIntlProvider>,
+    );
+
+    // Assert
+    expect(getByText('reports_title'));
+  });
+});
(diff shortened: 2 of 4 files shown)


src/components/Reports/ReportsNew.test.js
```diff
@@ -0,0 +1,22 @@
...
+    // Arrange
+    // Act
+    const { getByText } = render(
+      <DopplerIntlProvider>
+        <Reports />
+      </DopplerIntlProvider>,
+    );
+
+    // Assert
+    expect(getByText('Reports new page'));
```
andresmoschini, line 20: What do you think about expect for `reports_title`? 

Because, in other case, `Reports new page` will be modified soon and you will need to update the PR again.

src/components/Reports/ReportsNew.js
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <>
+      <FormattedMessage id="reports_title">
+        {(reports_title) => (
+          <Helmet>
+            <title>{reports_title}</title>
+          </Helmet>
+        )}
+      </FormattedMessage>
+      <section className="dp-container" style={{ marginTop: '20px' }}>
```
GusBaamonde, line 15: Remember that for the margins and paddings we have the atomic classes and they are 6 in 6, in this case it is (m-t-24) and in the case of using a gray header it is not necessary, since the header has its margin

andresmoschini, line 15: is it fine hardcoding the styles here?

CC: @GusBaamonde

GusBaamonde: I mention it in my previous comment

fcoronelmakingsense: yes! sorry i forgot.

fcoronelmakingsense: fixed

src/components/Reports/ReportsNew.test.js
```diff
@@ -0,0 +1,22 @@
...
+    // Arrange
+    // Act
+    const { getByText } = render(
+      <DopplerIntlProvider>
+        <Reports />
+      </DopplerIntlProvider>,
+    );
+
+    // Assert
+    expect(getByText('reports_title'));
```
lucianaDilizio, line 20: This means that it should find exactly the text 'reports_title' in the document, doesn't it? In this case the id="reports_title"

andresmoschini: No, it says that the text `reports_title` will be visible in the rendered output for the user, not in the source code.

### Explanation

I do not totally agree, but the philosophy behind React Testing Library [says "The queries help you to find elements in the same way that users will find them"](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change).

So, in general we find elements by text.

We have a problem with the localization and language, for that reason, we have a special _Intl Provider_:

```
 import DopplerIntlProvider from '../../i18n/DopplerIntlProvider.double-with-ids-as-values';
```

It returns the keys in place of the values. 

So, if you test this page (even when it will be modified) you probably have this text:

image

### Regrets

Oh,  but now I am understanding that `reports_title` text will be in the header, so our test is not so robust :(

lucianaDilizio: Ok, maybe "document" wasn't the best word, but the idea is to find the text in the rendered output. 
I didn't uderstand what you mean about "reports_title will be in the header" (I didn't find it in header component 😕)

PR #866, opened 2020-04-29
refactor(Reports): change the implementation to get total visits

fcoronelmakingsense:
- Refactor the `Reports` component to get the data for the total visits here.
- Refactor the `ReportsBox` component to only show data.

### DIFF
--- a/src/components/Reports/Reports.js
+++ b/src/components/Reports/Reports.js
@@ -34,6 +34,8 @@ const Reports = ({ dependencies: { datahubClient } }) => {
     dailyView: !periodSelectedDaysDefault,
   });
 
+  const [totalVisits, setTotalVisits] = useState({});
+
   const changeDomain = async (name) => {
     const domainFound = state.domains.find((item) => item.name === name);
     setState((prevState) => ({ ...prevState, domainSelected: domainFound }));
@@ -49,6 +51,28 @@ const Reports = ({ dependencies: { datahubClient } }) => {
       dailyView: !daysToFilter,
     }));
   };
+
+  useEffect(() => {
+    const fetchVisitsByPeriod = async () => {
+      const visitsWithoutEmail = await datahubClient.getTotalVisitsOfPeriod({
+        domainName: state.domainSelected.name,
+        dateFrom: state.dateFrom,
+        dateTo: state.dateTo,
+        emailFilter: 'without_email',
+      });
+      const visitsWithEmail = await datahubClient.getTotalVisitsOfPeriod({
+        domainName: state.domainSelected.name,
+        dateFrom: state.dateFrom,
+        dateTo: state.dateTo,
+        emailFilter: 'with_email',
+      });
+
+      setTotalVisits({ withEmail: visitsWithEmail, withoutEmail: visitsWithoutEmail });
+    };
+    if (state.domainSelected) {
+      fetchVisitsByPeriod();
+    }
+  }, [datahubClient, state.domainSelected, state.dateFrom, state.dateTo]);
 
   useEffect(() => {
     const fetchData = async () => {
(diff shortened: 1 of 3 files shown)


src/components/Reports/Reports.js
```diff
@@ -50,6 +52,28 @@ const Reports = ({ dependencies: { datahubClient } }) => {
...
+        domainName: state.domainSelected.name,
+        dateFrom: state.dateFrom,
+        dateTo: state.dateTo,
+        emailFilter: 'without_email',
+      });
+      const visitsWithEmail = await datahubClient.getTotalVisitsOfPeriod({
+        domainName: state.domainSelected.name,
+        dateFrom: state.dateFrom,
+        dateTo: state.dateTo,
+        emailFilter: 'without_email',
```
lucianaDilizio, line 67: Shouldn't be `with_email` or something like that in this case?

fcoronelmakingsense: yes sorry you are right

lucianaDilizio: Shouldn't this data be in a constant?

fcoronelmakingsense: maybe we can, but i didn't put there because this component its temporal until the new one its done, but maybe we can add

src/components/Reports/Reports.js
```diff
@@ -50,6 +52,28 @@ const Reports = ({ dependencies: { datahubClient } }) => {
...
+        emailFilter: 'with_email',
+      });
+
+      setTotalVisits({ withEmail: visitsWithEmail, withoutEmail: visitsWithoutEmail });
+    };
+    if (state.domainSelected) {
+      fetchVisitsByPeriod();
+    }
+  }, [datahubClient, state.domainSelected, state.dateFrom, state.dateTo]);
+
```
andresmoschini, line 76: Oh, you are keeping the refactorized old reports page. 

I imagined starting to splitting both pages and controls here. 

If you are comfortable with this approach, go ahead, if you feel that it is difficult, let me know and I can share with you what I was imagining.

src/components/Reports/ReportsBox/ReportsBox.test.js
```diff
@@ -0,0 +1,111 @@
...
+import { render, cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import DopplerIntlProvider from '../../../i18n/DopplerIntlProvider.double-with-ids-as-values';
+import ReportsBox from './ReportsBox';
+import { addDays } from '../../../utils';
+
+describe('ReportsBox component', () => {
+  afterEach(cleanup);
+
+  it('should show loading when visits are null or undefined', () => {
```
cbernat, line 11: Is this ok? when there are no visits the spinner would indicate that this is loading forever ...

andresmoschini: Good catch @cbernat!

fcoronelmakingsense: its only when is null or undefined. when have 0 visits shows the box

andresmoschini: What do you think about making the loading state explicit with a flag? (in general, it is better being explicit)

For that cases, using used to be better, you can limit the possible outputs.

fcoronelmakingsense: yes i already did now

cbernat: I've talk this with @fcoronelmakingsense and it seems when the promise is being resolved we have a temporal value of null or undefined, while waiting for the request, perhaps it would be great to have a flag for loading, in order to differentiate, null/undefined in loading and null/undefined on error.

src/components/Reports/ReportsBox/ReportsBox.test.js
```diff
@@ -0,0 +1,111 @@
...
+    // Act
+    const { container } = render(
+      <DopplerIntlProvider>
+        <ReportsBox
+          dateFrom={dateFrom}
+          dateTo={dateTo}
+          emailFilter={'with_email'}
+          today={today}
+          visits={visits}
+        />
```
andresmoschini, line 27: You have updated the test name, but I do not see the flag :P

fcoronelmakingsense: i put now is more clear

PR #872, opened 2020-05-04
Feat(ReportsTrafficSources): add new service and new component

fcoronelmakingsense:
- Rename old service method.
- Add new service method
- Add new tests
- Add new service doubles method
- Add new component

### DIFF
--- a/src/components/Reports/Reports.js
+++ b/src/components/Reports/Reports.js
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from 'react';
 import ReportsFilters from './ReportsFilters/ReportsFilters';
 import ReportsBox from './ReportsBox/ReportsBox';
 import ReportsPageRanking from './ReportsPageRanking/ReportsPageRanking';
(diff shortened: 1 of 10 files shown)


src/components/Reports/Reports.js
```diff
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from 'react';
 import ReportsFilters from './ReportsFilters/ReportsFilters';
 import ReportsBox from './ReportsBox/ReportsBox';
 import ReportsPageRanking from './ReportsPageRanking/ReportsPageRanking';
-import ReportsTrafficSources from './ReportsTrafficSources/ReportsTrafficSources';
+import ReportsTrafficSourcesOld from './ReportsTrafficSources/ReportsTrafficSourcesOld';
```
cbernat, line 5: You couldn't add a flag into `ReportsTrafficSources` right? to avoid the duplication.

fcoronelmakingsense: if add flag, need to call another method and show another data too, if you saw the service the interface are differents @cbernat

PR #874, opened 2020-05-07
feat(Reports): add new component and service to show daily visits

fcoronelmakingsense:
- Add new service method and new double service method.
- Add tests for this methods.
- Add new component and rename the old one
- Change some styles to be generic.

---

TODO add more tests

### DIFF
--- a/src/components/Reports/ReportsNew.test.js
+++ b/src/components/Reports/ReportsNew.test.js
@@ -3,6 +3,7 @@ import { render, cleanup, waitFor } from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 import DopplerIntlProvider from '../../i18n/DopplerIntlProvider.double-with-ids-as-values';
 import Reports from './ReportsNew';
+import { AppServicesProvider } from '../../services/pure-di';
 
 describe('Reports New page', () => {
   afterEach(cleanup);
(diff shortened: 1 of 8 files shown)


src/components/Reports/ReportsNew.test.js
```diff
@@ -50,4 +51,69 @@ describe('Reports New page', () => {
     // Assert
     await waitFor(() => expect(getByText('common.unexpected_error')));
   });
+
+  if (
```
cbernat, line 55: Why is this formatted in such a weird way? prettier?

cbernat:
```suggestion
  it (
```

src/components/shared/C3Chart/C3Chart.styles.js
```diff
@@ -47,7 +47,7 @@ export const C3ChartStyled = styled.div`
     opacity: 1;
   }
 
-  .c3-shapes-quantity {
```
cbernat, line 50: Why this is changed now? Won't it break?

fcoronelmakingsense: because quantity is the name of the BE attribute, change for the generic class

src/components/Reports/ReportsNew.test.js
fcoronelmakingsense: yes i changed!

PR #884, opened 2020-05-12
feat(ReportsHours): add new component with new endpoint

fcoronelmakingsense:
- Add new service method to get data with the new datahub endpoint.
- Add new service method in double to work local.
- Add tests in service.
- Rename old component to mantain working 2 components.
- Add the new component with the new endpoint.
- Add tests component


src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.test.js
```diff
@@ -0,0 +1,162 @@
...
+      >
+        <DopplerIntlProvider locale="en">
+          <ReportsHoursVisits domainName={domainName} dateFrom={dateFrom} dateTo={dateTo} />
+        </DopplerIntlProvider>
+      </AppServicesProvider>,
+    );
+
+    // Assert
+    expect(container.querySelector('.loading-box')).toBeInTheDocument();
+    await waitFor(() => expect(getByText('593')));
```
cbernat: Perhaps this `593` could be in a variable.

src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.test.js
```diff
@@ -0,0 +1,162 @@
...
+      >
+        <DopplerIntlProvider locale="en">
+          <ReportsHoursVisits domainName={domainName} dateFrom={dateFrom} dateTo={dateTo} />
+        </DopplerIntlProvider>
+      </AppServicesProvider>,
+    );
+
+    // Assert
+    expect(container.querySelector('.loading-box')).toBeInTheDocument();
+    await waitFor(() => expect(getByText('50')));
```
cbernat, line 160: The same with this `50`

src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.test.js
fcoronelmakingsense: we have before this discussion and says the both options are ok

PR #911, opened 2020-05-28
fix(index): add spinner when load webapp

fcoronelmakingsense:
- Add spinner behind app


andresmoschini:
It looks interesting, keep me updated if it works, you can also share a video or ask me to test it.

PR #918, opened 2020-06-02
fix(Offline): add no index in offline page

fcoronelmakingsense:
- Add no index in offline page and in new features to search robots no index this page.

### DIFF
--- a/src/components/Offline/Offline.js
+++ b/src/components/Offline/Offline.js
@@ -1,7 +1,11 @@
 import React from 'react';
+import { Helmet } from 'react-helmet';
 
 const Offline = () => (
   <div className="dp-app-container">
+    <Helmet>
+      <meta name="robots" content="noindex,nofollow" />
+    </Helmet>
     <header style={{ height: '84px' }}></header>
     <main style={{ display: 'block', widht: '100%' }}>
       <section className="section__block image__main">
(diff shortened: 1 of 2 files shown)


fcoronelmakingsense:
test this please

src/components/Offline/Offline.js
```diff
@@ -1,7 +1,11 @@
 import React from 'react';
+import { Helmet } from 'react-helmet';
 
 const Offline = () => (
   <div className="dp-app-container">
+    <Helmet>
+      <meta name="robots" content="noindex,nofollow" />
+    </Helmet>
```
andresmoschini, line 8: It should be also added in `NewFeatures.js`, right?

fcoronelmakingsense: you right, good catch!

fcoronelmakingsense:
test this please

src/components/Offline/Offline.js
fcoronelmakingsense: changed

src/components/Offline/Offline.js
```diff
@@ -1,7 +1,11 @@
 import React from 'react';
+import { Helmet } from 'react-helmet';
 
 const Offline = () => (
   <div className="dp-app-container">
+    <Helmet>
+      <meta name="robots" content="noindex,nofollow" />
```
cbernat, line 7: ❤️

PR #921, opened 2020-06-05
[DW-304] Subscriber history tabs

lucianaDilizio:
Add tabs in Subscriber History and GDPR screens.

image

### DIFF
--- a/src/components/Reports/GDPR/SubscriberGdpr.js
+++ b/src/components/Reports/GDPR/SubscriberGdpr.js
@@ -12,6 +12,7 @@ import { Loading } from '../../Loading/Loading';
 import { StarsScore } from '../../shared/StarsScore/StarsScore';
 import SafeRedirect from '../../SafeRedirect';
 import * as S from './SubscriberGdpr.styles';
+import { Tabs } from '../../shared/Tabs/Tabs';
 
 const SubscriberGdpr = ({ location, dependencies: { dopplerApiClient } }) => {
   const intl = useIntl();
@@ -60,11 +61,25 @@ const SubscriberGdpr = ({ location, dependencies: { dopplerApiClient } }) => {
           return field;
         });
 
+        const tabsProperties = [
+          {
+            url: `/reports/subscriber-history?email=${subscriber.email}`,
+            descriptionId: 'subscriber_history.title',
+            active: false,
+          },
+          {
+            url: `/reports/subscriber-gdpr?email=${subscriber.email}`,
+            descriptionId: 'subscriber_gdpr.page_title',
+            active: true,
+          },
+        ];
+
         setState({
           loading: false,
           subscriber: subscriber,
           fields: fields,
           email: email,
+          tabsProperties: tabsProperties,
         });
       } else {
         setState({ loading: false });
@@ -105,7 +120,8 @@ const SubscriberGdpr = ({ location, dependencies: { dopplerApiClient } }) => {
           </header>
           <section className="dp-container">
             <div className="dp-rowflex">
-              <div className="col-sm-12 m-b-36">
+              <Tabs tabsProperties={state.tabsProperties}></Tabs>
+              <div className="tab--container col-sm-12 m-b-36">
                 <div className="dp-block-wlp dp-box-shadow">
                   <header className="dp-header-campaing dp-rowflex p-l-18">
                     <div className="col-lg-6 col-md-12 m-b-24">
--- a/src/components/Reports/SubscriberHistory/SubscriberHistory.js
+++ b/src/components/Reports/SubscriberHistory/SubscriberHistory.js
@@ -10,6 +10,7 @@ import {
 } from '../../../utils';
 import { StarsScore } from '../../shared/StarsScore/StarsScore';
 import { Pagination } from '../../shared/Pagination/Pagination';
+import { Tabs } from '../../shared/Tabs/Tabs';
 import SafeRedirect from '../../SafeRedirect';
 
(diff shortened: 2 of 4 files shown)


src/components/Reports/SubscriberHistory/SubscriberHistory.js
```diff
@@ -76,6 +77,20 @@ const SubscriberHistory = ({
...
+            url: `/reports/subscriber-history?email=${subscriber.email}`,
+            descriptionId: 'subscriber_history.title',
+            active: true,
+          },
+          {
+            url: `/reports/subscriber-gdpr?email=${subscriber.email}`,
+            descriptionId: 'subscriber_gdpr.page_title',
+            active: false,
+          },
+        ];
```
andresmoschini, line 92: Is it possible to avoid code duplication here?

src/components/Reports/SubscriberHistory/SubscriberHistory.js
```diff
@@ -117,7 +133,8 @@ const SubscriberHistory = ({
       </header>
       <section className="dp-container">
         <div className="dp-rowflex">
-          <div className="col-sm-12 m-b-36">
+          <Tabs tabsProperties={state.tabsProperties}></Tabs>
```
andresmoschini, line 136: Maybe you can create a component with these specific tabs and reuse it in these pages

andresmoschini: I mean, a new layer over these ones.

cbernat: I agree here with Andres. A nice refactor to avoid repeating. Here remeber about orthogonality, what happens if some day I want to add a third tab? I will need this added in both components...

src/components/Reports/SubscriberHistory/SubscriberHistory.js
lucianaDilizio: But it's not duplicated. Each object has different values. I think that I don't understand the idea.

src/components/Reports/SubscriberHistory/SubscriberHistory.js
lucianaDilizio: I tried to make the component reusable. At least, I have to find a way to send which tab is selected.

src/components/Reports/GDPR/SubscriberGdpr.js
```diff
@@ -60,11 +61,25 @@ const SubscriberGdpr = ({ location, dependencies: { dopplerApiClient } }) => {
...
+            active: true,
+          },
+        ];
+
         setState({
           loading: false,
           subscriber: subscriber,
           fields: fields,
           email: email,
+          tabsProperties: tabsProperties,
```
fcoronelmakingsense, line 82: its neccesary put in the state? maybe never change

src/components/shared/Tabs/Tabs.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+
+export const Tabs = ({ tabsProperties }) => {
+  return (
+    <nav className="tabs-wrapper">
+      <ul className="tabs-nav" data-tab-active="1">
+        {tabsProperties.map((tab) => (
+          <li className="tab--item">
+            <a href={tab.url} className={tab.active ? 'tab--link active' : 'tab--link'}>
```
fcoronelmakingsense, line 10: maybe here its better `Link `component

src/components/Reports/SubscriberHistory/SubscriberHistory.js
```diff
@@ -76,6 +77,20 @@ const SubscriberHistory = ({
...
+        const tabsProperties = [
+          {
+            url: `/reports/subscriber-history?email=${subscriber.email}`,
+            descriptionId: 'subscriber_history.title',
+            active: true,
+          },
+          {
+            url: `/reports/subscriber-gdpr?email=${subscriber.email}`,
+            descriptionId: 'subscriber_gdpr.page_title',
+            active: false,
```
andresmoschini, line 90: Another interesting thing, not related to the previous discussion, is to use a more idiomatic React code. For example, in place of using this kind of array, React developers do something like this:

```html
<TabGroup>
  <TabItem active url={`/reports/subscriber-history?email=${subscriber.email}`}>
    <FormattedMessage id="subscriber_history.title" />
  </TabItem>
  <TabItem url={`/reports/subscriber-gdpr?email=${subscriber.email}`}>
    <FormattedMessage id="subscriber_gdpr.page_title" />
  </TabItem>
</TabGroup>
```

Or maybe:

```html
<TabGroup currentUrl={currentUrl}>
  <TabItem url={`/reports/subscriber-history?email=${subscriber.email}`}>
    <FormattedMessage id="subscriber_history.title" />
  </TabItem>
  <TabItem url={`/reports/subscriber-gdpr?email=${subscriber.email}`}>
    <FormattedMessage id="subscriber_gdpr.page_title" />
  </TabItem>
</TabGroup>
```

lucianaDilizio: Interesting 😄 👍

src/components/Reports/GDPR/SubscriberGdpr.js
andresmoschini: Good catch @fcoronelmakingsense!

andresmoschini:
This PR is deprecated, right?

lucianaDilizio:
@andresmoschini yes! I have to close it! Thanks!

PR #926, opened 2020-06-08
fix(Reports): update default weeks to 3

fcoronelmakingsense:
- change default reports period selected from today to 3 weeks


cbernat:
✔️

PR #936, opened 2020-06-11
fix(App): add ui system library version in env file

fcoronelmakingsense:
- Add new variable in `env` file to set doppler ui library version


andresmoschini:
Is it possible apply the same environment variable into index.html?

We have inconsistencies right now:

https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L17

https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L80

fcoronelmakingsense:
> Is it possible apply the same environment variable into index.html?
> 
> We have inconsistencies right now:
> 
> https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L17
> 
> https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L80

How? because html it's outside react

andresmoschini:
> How? because html it's outside react

I do not know. Another option is adding a  kind of automated test to ensure consistency.

cbernat:
@fcoronelmakingsense you could use this env variable in the html too, like this https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L19

andresmoschini:
> @fcoronelmakingsense you could use this env variable in the html too, like this https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L19

👏 👏 👏

fcoronelmakingsense:
> @fcoronelmakingsense you could use this env variable in the html too, like this https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L19

Thanks y already apply!

.env
```diff
@@ -13,3 +13,4 @@ REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 NODE_ICU_DATA=node_modules/full-icu
 REACT_APP_ROUTER=browser
+REACT_APP_DOPPLER_UI_LIBRARY_VERSION = 3.51.5
```
andresmoschini, line 16: What do you think about including the `v` in the version value?

```suggestion
REACT_APP_DOPPLER_UI_LIBRARY_VERSION = v3.51.5
```

andresmoschini: Another option could be this:

```suggestion
REACT_APP_DOPPLER_UI_BASE_URL = https://cdn.fromdoppler.com/doppler-ui-library/v3.51.5
```

andresmoschini:
🚀

PR #937, opened 2020-06-12
fix: get all subscriber info

lucianaDilizio:
Unsubscribed Date was showed wrong. (It was showing today)

image


src/components/Reports/GDPR/SubscriberGdpr.js
```diff
@@ -39,11 +39,9 @@ const SubscriberGdpr = ({ location, dependencies: { dopplerApiClient } }) => {
       const allFields = await dopplerApiClient.getUserFields();
       if (responseSubscriber.success && allFields.success) {
         const subscriber = {
+          ...responseSubscriber.value,
```
cbernat, line 42: ❤️

PR #947, opened 2020-06-17
feat: add publish webapp image into fromdoppler organization

cbernat:
### Add publish image into fromdoppler organization

Add missing environments into jenkins file to replace the use of `darosw` user in docker hub. 
Then I will update the swarm services.

We had only production and demo environments published in `fromdoppler`

Issue: DW-383

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,7 +45,6 @@ pipeline {
                 }
             }
             steps {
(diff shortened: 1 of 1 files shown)


cbernat:
@andresmoschini I'm not sure if there is a more intelligent approach...

andresmoschini:
👍

PR #958, opened 2020-06-24
DW-395: Push demo

cbernat:
### Small push demo in new features

- In new features added new section push
- This page can be accessed through `/push` url
- Also added configuration for all environments of datahub script

image

When we enter it asks for permission
image

Instructions
image

If we accept we will see a message in console with the token id, that has been saved in datahub
image

Test link --> https://cdn.fromdoppler.com/doppler-webapp/build2474/#/push


cbernat:
test this please

PR #979, opened 2020-07-13
fix(infrastructure): remove akamai from docker configuration

fcoronelmakingsense:


cbernat:
Seems ok :) and it worked wonderfully. Thanks @fcoronelmakingsense ❤️

PR #987, opened 2020-07-15
fix(Signup): flag tests

fcoronelmakingsense:
- Fix tests in signup fail random, because the service of flag have a delay to obtain the data

### DIFF
--- a/src/components/Signup/Signup.test.js
+++ b/src/components/Signup/Signup.test.js
@@ -75,11 +75,11 @@ describe('Signup', () => {
 
     // Assert
     await waitFor(() => {
-      setTimeout(() => {
-        expect(container.querySelector('#iti-0__item-dz').nextElementSibling.id).toBe(
-          'iti-0__item-ar',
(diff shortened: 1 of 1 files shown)


cbernat:
Is it too difficult just to mock de call? Instead of incrementing the timeout

fcoronelmakingsense:
> Is it too difficult just to mock de call? Instead of incrementing the timeout

fixed

src/components/Signup/Signup.test.js
```diff
@@ -153,11 +153,11 @@ describe('Signup', () => {
...
     // Assert
     await waitFor(() => {
-      setTimeout(() => {
-        expect(container.querySelector('#iti-1__item-io').nextElementSibling.id).toBe(
-          'iti-1__item-bn',
-        );
-      }, 1000);
+      expect(
+        container
+          .querySelector('[data-country-code="io"]')
```
cbernat, line 158: 😮

PR #988, opened 2020-07-16
[feat] improve blocked account error msg

lucianaDilizio:
Improve blocked account error message to be more specific about the error. 
CDN: https://cdn.fromdoppler.com/doppler-webapp/int-build2530/#/login
user: lgarcia+25@makingsense.com
pass: 1qaz2wsx

Please @mrunge86 and @santiagopaolucci check this.  

To do: 

- [x] Define a different message for Mon to Fri from 8:00 to 20:00

English:
image

Spanish:
image

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -1,6 +1,6 @@
 import React, { useState, useMemo, useEffect } from 'react';
 import { Link, Redirect } from 'react-router-dom';
-import { useIntl } from 'react-intl';
+import { useIntl, FormattedMessage } from 'react-intl';
 import { Helmet } from 'react-helmet';
 import {
   EmailFieldItem,
@@ -19,7 +19,12 @@ import { FormattedMessageMarkdown } from '../../i18n/FormattedMessageMarkdown';
 import { connect } from 'formik';
 import Promotions from '../shared/Promotions/Promotions';
 import queryString from 'query-string';
-import { addLogEntry, extractParameter } from '../../utils';
+import {
+  addLogEntry,
+  extractParameter,
+  isZohoChatOnline,
+  openZohoChatWithMessage,
+} from '../../utils';
 
 const fieldNames = {
   user: 'user',
@@ -68,11 +73,28 @@ const isActivactionInProgress = (location) => {
   return parsedQuery && parsedQuery === 'true';
 };
 
-const LoginErrorBlockedAccountNotPayed = () => (
-  <p>
-    <FormattedMessageMarkdown id="login.error_payment_MD" />
-  </p>
-);
+const LoginErrorBlockedAccountNotPayed = () => {
+  const intl = useIntl();
+  const _ = (id, values) => intl.formatMessage({ id: id }, values);
+  return (
+    <p>
+      {isZohoChatOnline() ? (
+        <>
+          <FormattedMessage id={'login.error_payment_online_part1'} />
+          <button
+            type="button"
+            onClick={() => openZohoChatWithMessage(_('login.error_payment_online_zoho_msg'))}
+          >
+            <FormattedMessage id="login.error_payment_online_link" />
+          </button>
+          <FormattedMessage id={'login.error_payment_online_part2'} />
+        </>
+      ) : (
+        <FormattedMessageMarkdown id={'login.error_payment_MD_offline'} />
+      )}
+    </p>
+  );
+};
 
 /**
  * Login Page
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -80,14 +80,19 @@ const LoginErrorBlockedAccountNotPayed = () => {
     <p>
       {isZohoChatOnline() ? (
         <>
-          <FormattedMessage id={'login.error_payment_online_part1'} />
-          <button
-            type="button"
-            onClick={() => openZohoChatWithMessage(_('login.error_payment_online_zoho_msg'))}
-          >
-            <FormattedMessage id="login.error_payment_online_link" />
-          </button>
-          <FormattedMessage id={'login.error_payment_online_part2'} />
+          <FormattedMessage
+            id={'login.error_payment_online'}
+            values={{
+              button: (chunk) => (
+                <button
+                  type="button"
+                  onClick={() => openZohoChatWithMessage(_('login.error_payment_online_zoho_msg'))}
+                >
+                  {chunk}
+                </button>
+              ),
+            }}
+          />
         </>
       ) : (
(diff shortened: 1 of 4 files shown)


cbernat:
This message will show always, is it ok like this?

src/i18n/es.js
```diff
@@ -93,7 +94,7 @@ Encontrarás un Email con los pasos a seguir.`,
     copyright_MD: `© ${year} Doppler LLC. Todos los derechos reservados. [Política de Privacidad y Legales](${urlPrivacyFromLogin}).`,
     enter_doppler: `Ingresa a tu cuenta`,
     enter_doppler_sub: `¡Hoy es un buen día para potenciar tu negocio con el poder del Email, Automation & Data Marketing!`,
-    error_payment_MD: `La cuenta está bloqueda, por favor [contáctate con Soporte](${mailtoSupport}).`,
+    error_payment_MD: `¡Ouch! Esta cuenta ha sido suspendida por falta de pago.\n\nPor favor [contacta al equipo de Atención al Cliente](${mailtoSupportBlockedAccount}) para solucionarlo.`,
```
andresmoschini, line 97: It is fine, but what do you think about storing the messages with interlines in a more friendly way?

```suggestion
    error_payment_MD: `
¡Ouch! Esta cuenta ha sido suspendida por falta de pago.

Por favor [contacta al equipo de Atención al Cliente](${mailtoSupportBlockedAccount}) para solucionarlo.`,
```

lucianaDilizio: Maybe I am doing something wrong but it seems to be not too friendly with that format:
image

src/utils.ts
```diff
@@ -204,3 +205,14 @@ const urlsWebApp: any = [
...
   return urlsWebApp.find((item: any) => item.url === currentUrl);
 }
+
+export function isZohoChatOnline() {
+  const currentDate = new Date();
+  // (UTC-03:00) City of Buenos Aires 8:00 to 20:00
+  return currentDate.getUTCHours() >= 11 && currentDate.getUTCHours() < 23;
+}
+
+export function openZohoChat(message: string){
```
cbernat, line 215:
```suggestion
export function openZohoChatWithMessage(message: string){
```

src/utils.ts
```diff
@@ -204,3 +205,14 @@ const urlsWebApp: any = [
 export function getCurrentPageForUrl(currentUrl: string): any {
   return urlsWebApp.find((item: any) => item.url === currentUrl);
 }
+
+export function isZohoChatOnline() {
+  const currentDate = new Date();
+  // (UTC-03:00) City of Buenos Aires 8:00 to 20:00
+  return currentDate.getUTCHours() >= 11 && currentDate.getUTCHours() < 23;
```
andresmoschini, line 212: It should be configurable using a settings file.

Maybe this is not the best moment to do that, in that case, what do you think about adding a TODO note?

```suggestion
  // TODO: allow to configure the schedule in a settings file.
  // (UTC-03:00) City of Buenos Aires 8:00 to 20:00
  return currentDate.getUTCHours() >= 11 && currentDate.getUTCHours() < 23;
```

src/i18n/en.js
```diff
@@ -94,7 +95,13 @@ You'll find an Email with steps to follow.`,
...
     enter_doppler: `Log In`,
     enter_doppler_sub: `Today is a good day to boost your business with Email, Automation & Data Marketing!`,
-    error_payment_MD: `Account blocked, please [contact Support](${mailtoSupport}).`,
+    error_payment_MD_offline: `
+Ouch! This account has been suspended for non-payment.
+
+Please [contact the Customer Support team](${mailtoSupportBlockedAccount}) for help.`,
+    error_payment_online_link: `Chat with the Customer Support team`,
+    error_payment_online_part1: `Ouch! This account has been suspended for non-payment. `,
+    error_payment_online_part2: ` for help.`,
```
andresmoschini, line 104: It does not look fine, there should be a way to inject the button, like this:

```suggestion
    error_payment_online_part1: `Ouch! This account has been suspended for non-payment. {clickeableMessage} for help.`,
```

But, if you have no time to research about it now, please add a TODO note to make it clear that we do not want to deal with this kind of things in this way in the future:

```suggestion
    // Ugly patch because I have no time to research how to inject an element.
    // TODO: merge these two sections in a single entry.
    error_payment_online_part1: `Ouch! This account has been suspended for non-payment. `,
    error_payment_online_part2: ` for help.`,
```

src/components/Login/Login.js
```diff
@@ -68,9 +68,20 @@ const isActivactionInProgress = (location) => {
   return parsedQuery && parsedQuery === 'true';
 };
 
+const message = '¡Hola! ¿Me ayudas con mi cuenta suspendida por falta de pago?'
```
andresmoschini, line 71: I know that it is hardcoded, but what do you thing about moving to the lenguage file anywary?

Or... add a TODO haha

andresmoschini:
@lucianaDilizio I did not understand the issue that you have explained in the standup, do you have an screenshot or a video to undertastan it better? (is this the related PR?)

lucianaDilizio:
@andresmoschini It was happening something like this (but now I can't reproduce it completely, it only happens when you press enter instead clicking on login button) 
 
loginFail

src/i18n/en.js
andresmoschini: Creo que la solución anda por acá: https://formatjs.io/docs/react-intl/components#rich-text-formatting

Si no me equivoco @fcoronelmakingsense estuvo jugando con esas opciones.

andresmoschini:
> @andresmoschini It was happening something like this (but now I can't reproduce it completely, it only happens when you press enter instead clicking on login button)

Bueno, es un caso raro, creo que ya estaba, no lo introduciste en este PR, no?

Si es así, yo lo dejaría afuera y en todo caso intentaría reolverlo después.

src/i18n/en.js
```diff
@@ -94,7 +95,15 @@ You'll find an Email with steps to follow.`,
...
     enter_doppler_sub: `Today is a good day to boost your business with Email, Automation & Data Marketing!`,
-    error_payment_MD: `Account blocked, please [contact Support](${mailtoSupport}).`,
+    error_payment_MD_offline: `
+Ouch! This account has been suspended for non-payment.
+
+Please [contact the Customer Support team](${mailtoSupportBlockedAccount}) for help.`,
+    error_payment_online: `
+Ouch! This account has been suspended for non-payment.
+
+<button>Chat with the Customer Support team</button> for help.`,
```
andresmoschini, line 105: ❤️

PR #998, opened 2020-07-21
feat: add TODO to manage new error

lucianaDilizio:
### Manage new login error

It adds a TODO to manage not payed cancelled accounts.
This change is to avoid show a general error in this cases. 

https://cdn.fromdoppler.com/doppler-webapp/int-build2531/#/login

### DIFF
--- a/src/services/doppler-legacy-client.ts
+++ b/src/services/doppler-legacy-client.ts
@@ -390,6 +390,10 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
           case 'CancelatedAccount': {
             return { expectedError: { cancelatedAccount: true } };
           }
+          case 'CancelatedAccountNotPayed': {
+            // TODO: manage CancelatedAccountNotPayed error correctly
+            return { expectedError: { cancelatedAccount: true } };
+          }
           case 'BlockedAccountInvalidPassword': {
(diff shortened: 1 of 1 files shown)


src/services/doppler-legacy-client.ts
```diff
@@ -390,6 +390,10 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
           case 'CancelatedAccount': {
             return { expectedError: { cancelatedAccount: true } };
           }
+          case 'CancelatedAccountNotPayed': {
```
cbernat, line 393: How about adding a new test for this scenario as well?

cbernat:
You could testing hardcoded values in the service double here https://github.com/FromDoppler/doppler-webapp/blob/master/src/services/doppler-legacy-client.doubles.ts#L27

src/services/doppler-legacy-client.ts
cbernat: I've talked with @lucianaDilizio and we have no tests of this nature yet, so she will test this manually.

cbernat:
> You could testing hardcoded values in the service double here https://github.com/FromDoppler/doppler-webapp/blob/master/src/services/doppler-legacy-client.doubles.ts#L27

It has no meaning doing this so ignore this comment. 😄

cbernat:
🚀

PR #999, opened 2020-07-24
DW-421: set cache headers in nginx

cbernat:
### Add cache control headers in nginx
We had an issue in production with coudflare strong caching of non static resources like html, but the rule `Origin Cache Control: On` allows to do the caching according to the headers sent by our servers. We didn´t have any headers set, now we will.

For all resources outside static:

image

For static requests:

image

#### Changes
- headers on local conf
- headers on swarm production conf
- some testing links that may be used later also

### DIFF
--- a/conf/site-local.conf
+++ b/conf/site-local.conf
@@ -19,18 +19,23 @@ server {
 
         location /static/ {
                 root /usr/share/nginx/html;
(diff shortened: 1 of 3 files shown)


andresmoschini:
❤️

PR #1001, opened 2020-07-27
fix: change regex for subscriber sections

lucianaDilizio:
Fix issue with subscriber pages redirection.
It was redirecting all URL's like: /reports/XXXX-XXXXX

Now it will redirec only: /reports/subscriber-XXXXX

http://cdn.fromdoppler.com/doppler-webapp/build2536/#/reports/partials-campaigns


cbernat:
👍

PR #1008, opened 2020-07-29
feat(PrivatePages): add header section template with common styles

fcoronelmakingsense:
- Create shared component with the internal header section styles and structure for reuse in internal pages.

image

### DIFF
--- a/src/components/NewFeatures/NewFeatures.js
+++ b/src/components/NewFeatures/NewFeatures.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
 import { Helmet } from 'react-helmet';
+import HeaderSection from '../shared/HeaderSection/HeaderSection';
 
 const NewFeatures = () => {
   return (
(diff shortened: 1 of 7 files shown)


src/components/NewFeatures/NewFeatures.js
```diff
@@ -1,27 +1,23 @@
...
-              <h2>Funcionalidades Nuevas</h2>
-              <p>
-                Features sin acceso para los usuarios todavía y con trabajo pendiente de contenidos
-                y diseño.
-              </p>
-            </div>
-          </div>
-          <span className="arrow"></span>
+      <HeaderSection>
+        <div className="col-sm-12 col-md-12 col-lg-12">
```
andresmoschini, line 13: This div cannot be part of the HeaderSection?

fcoronelmakingsense: Because in some pages use two times, like in reports

andresmoschini: 👍 
In the future, we could parametrize it, something like this:

```html
       <HeaderSection col-sm=12 col-md=12 col-lg=12>
```

PR #1017, opened 2020-08-03
Dw 435 plans sliders

lucianaDilizio:
### Basic Plan Calculator Screen

Link test:  https://cdn.fromdoppler.com/doppler-webapp/demo-build2569/#/plan-calculator

✔️Included in this PR:
- Slider element
- basic behavior to change price based on slider value
------------------------------------------------------------------------------
⛔ Not included in this PR:
- styles
- discount element
- content
- double service to obtain data
- params

image

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -22,6 +22,7 @@ import Offline from './components/Offline/Offline';
 import PushNotifications from './components/PushNotifications/PushNotifications';
 import SubscribersLegacyUrlRedirect from './components/Reports/Subscribers/SubscribersLegacyUrlRedirect';
 import { ChangePlan } from './components/ChangePlan/ChangePlan';
+import PlanCalculator from './components/DopplerPlus/PlanCalculator';
 
 /**
  * @param { Object } props - props
@@ -124,6 +125,7 @@ const App = ({ locale, location, dependencies: { appSessionRef, sessionManager }
             exact
             component={ReportsPartialsCampaigns}
           />
+          <PrivateRoute path="/plan-calculator" exact component={PlanCalculator} />
           <PublicRouteWithLegacyFallback exact path="/login" />
           <PublicRouteWithLegacyFallback exact path="/signup" />
           <PublicRouteWithLegacyFallback exact path="/login/reset-password" />
--- a/src/components/DopplerPlus/PlanCalculator.js
+++ b/src/components/DopplerPlus/PlanCalculator.js
@@ -1,20 +1,5 @@
 import React, { useState } from 'react';
-
-const Slider = ({ min, max, step, defaultValue, handleChange }) => {
-  // TODO: add styles
-  return (
-    <input
-      id="slider"
-      style={{ marginTop: '100px', padding: '0' }}
-      type="range"
-      min={min}
-      max={max}
-      step={step}
-      defaultValue={defaultValue}
-      onChange={(e) => handleChange(e.target.value)}
-    />
-  );
-};
+import { Slider } from '../shared/Slider/Slider';
 
 const PlanCalculator = () => {
   // TODO: get data from a double service
--- a/src/components/DopplerPlus/PlanCalculator.js
+++ b/src/components/DopplerPlus/PlanCalculator.js
@@ -1,0 +1,52 @@
+import React, { useState } from 'react';
+
+const Slider = ({ min, max, step, defaultValue, handleChange }) => {
+  // TODO: add styles
+  return (
+    <input
+      id="slider"
+      style={{ marginTop: '100px', padding: '0' }}
+      type="range"
+      min={min}
+      max={max}
+      step={step}
+      defaultValue={defaultValue}
+      onChange={(e) => handleChange(e.target.value)}
+    />
+  );
+};
+
+const PlanCalculator = () => {
+  // TODO: get data from a double service
+  const [state, setState] = useState({ price: 15 });
+  const valuesPerRange = [15, 29, 48, 77, 106, 145, 240, 340, 460];
+
+  const updatePrice = (index) => {
+    setState({
+      price: valuesPerRange[index],
+    });
+  };
+
+  return (
+    <div className="p-t-54 p-b-54" style={{ backgroundColor: '#f6f6f6', flex: '1' }}>
+      <section className="dp-container">
+        <div className="dp-rowflex">
+          <div className="col-sm-12" style={{ textAlign: 'center' }}>
+            {/* TODO: change this to intl elemnt */}
+            <h1>Plan STANDARD</h1>
+            <p style={{ paddingBottom: '50px' }}>
+              {/* TODO: change this to intl elemnt */}
+              ¿Cuántos contactos tienes? Utiliza el slider para calcular el costo final de tu Plan
+            </p>
+            <span>US$</span>
+            <span style={{ fontSize: '40px' }}>{state.price}</span>
+            <p>por mes</p>
+            <Slider min={0} max={8} step={1} defaultValue={0} handleChange={updatePrice} />
+          </div>
+        </div>
+      </section>
+    </div>
+  );
+};
+
+export default PlanCalculator;
--- a/src/components/DopplerPlus/PlanCalculator.js
+++ b/src/components/DopplerPlus/PlanCalculator.js
@@ -15,6 +15,10 @@ const PlanCalculator = () => {
     { idPlan: 9, price: 460, amount: 100000 },
   ];
 
+  const plansTooltipDescriptions = plansList.map((plan) => {
+    return plan.amount + ' Suscriptores';
+  });
+
   const initialPlan = plansList[0];
 
   const [currentPlan, updateSelectedPlan] = useReducer((currentPlan, index) => {
@@ -36,9 +40,7 @@ const PlanCalculator = () => {
             <span style={{ fontSize: '40px' }}>{currentPlan.price}</span>
             <p>por mes</p>
             <Slider
-              min={0}
-              max={plansList.length - 1}
-              step={1}
+              tooltipDescriptions={plansTooltipDescriptions}
               defaultValue={0}
               handleChange={updateSelectedPlan}
             />
--- a/src/components/DopplerPlus/PlanCalculator.js
+++ b/src/components/DopplerPlus/PlanCalculator.js
@@ -1,16 +1,25 @@
-import React, { useState } from 'react';
+import React, { useReducer } from 'react';
 import { Slider } from '../shared/Slider/Slider';
 
 const PlanCalculator = () => {
   // TODO: get data from a double service
-  const [state, setState] = useState({ price: 15 });
-  const valuesPerRange = [15, 29, 48, 77, 106, 145, 240, 340, 460];
+  const plansList = [
+    { idPlan: 1, price: 15, amount: 1500 },
+    { idPlan: 2, price: 29, amount: 2500 },
+    { idPlan: 3, price: 48, amount: 5000 },
+    { idPlan: 4, price: 77, amount: 10000 },
+    { idPlan: 5, price: 106, amount: 15000 },
+    { idPlan: 6, price: 145, amount: 25000 },
+    { idPlan: 7, price: 240, amount: 50000 },
+    { idPlan: 9, price: 340, amount: 75000 },
+    { idPlan: 9, price: 460, amount: 100000 },
+  ];
 
-  const updatePrice = (index) => {
-    setState({
-      price: valuesPerRange[index],
-    });
-  };
+  const initialPlan = plansList[0];
+
+  const [currentPlan, updateSelectedPlan] = useReducer((currentPlan, index) => {
+    return plansList[index] || currentPlan;
+  }, initialPlan);
 
   return (
     <div className="p-t-54 p-b-54" style={{ backgroundColor: '#f6f6f6', flex: '1' }}>
@@ -24,9 +33,15 @@ const PlanCalculator = () => {
               ¿Cuántos contactos tienes? Utiliza el slider para calcular el costo final de tu Plan
             </p>
             <span>US$</span>
-            <span style={{ fontSize: '40px' }}>{state.price}</span>
+            <span style={{ fontSize: '40px' }}>{currentPlan.price}</span>
             <p>por mes</p>
-            <Slider min={0} max={8} step={1} defaultValue={0} handleChange={updatePrice} />
+            <Slider
+              min={0}
+              max={plansList.length - 1}
+              step={1}
+              defaultValue={0}
+              handleChange={updateSelectedPlan}
+            />
           </div>
         </div>
       </section>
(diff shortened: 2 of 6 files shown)


src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from 'react';
+
+const Slider = ({ min, max, step, defaultValue, handleChange }) => {
```
cbernat, line 3: I think this slider perhaps should be in a different file or allowed to be exported from this file. Don't you think?

GusBaamonde:
I like that when you focus on the slider, you can control it with the cursors! very friendly!!

lucianaDilizio:
> I like that when you focus on the slider, you can control it with the cursors! very friendly!!

I didn't note that 😅.  Nice!

src/components/DopplerPlus/PlanCalculator.js
lucianaDilizio: Yes, I thought about it, but since we use it only in this element, maybe it's not necessary. We do something similar in other elements. I don't know what is the rule that makes to keep an element in the same file or make it exportable.

src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -0,0 +1,52 @@
...
+            <p>por mes</p>
+            <Slider min={0} max={8} step={1} defaultValue={0} handleChange={updatePrice} />
+          </div>
+        </div>
+      </section>
+    </div>
+  );
+};
+
+export default PlanCalculator;
```
cbernat, line 52:
```suggestion
export [ PlanCalculator, Slider ];
```

src/components/shared/Slider/Slider.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+export const Slider = ({ min, max, step, defaultValue, handleChange }) => {
```
cbernat, line 3: 💯

src/components/shared/Slider/Slider.js
```diff
@@ -0,0 +1,17 @@
...
+  // TODO: add styles
+  return (
+    <input
+      id="slider"
```
cbernat, line 7: What will happen if you had more than one slider in the same page? this id would be an issue... 🤔

src/components/shared/Slider/Slider.test.js
```diff
@@ -0,0 +1,19 @@
...
+
+  it('renders stars with score 0', () => {
+    // Act
+    const mockedFunction = () => null;
+    const { container } = render(
+      <Slider min={0} max={8} step={1} defaultValue={0} handleChange={mockedFunction} />,
+    );
+
+    // Assert
+    expect(container.querySelector('input[type="range"]')).toBeInTheDocument();
```
cbernat, line 17: ❤️

src/App.js
```diff
@@ -112,6 +113,7 @@ const App = ({ locale, location, dependencies: { appSessionRef, sessionManager }
             exact
             component={ReportsPartialsCampaigns}
           />
+          <PrivateRoute path="/plan-calculator" exact component={PlanCalculator} />
```
andresmoschini, line 116: Is _private_ ok? I mean, only authenticated users can open it?

src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -0,0 +1,37 @@
+import React, { useState } from 'react';
+import { Slider } from '../shared/Slider/Slider';
+
+const PlanCalculator = () => {
+  // TODO: get data from a double service
+  const [state, setState] = useState({ price: 15 });
```
andresmoschini, line 6: What do you think about simply:

```suggestion
  const [price, setPrice] = useState(15);
```

src/components/shared/Slider/Slider.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export const Slider = ({ min, max, step, defaultValue, handleChange }) => {
```
andresmoschini, line 3: Since we will use it with a predefined set of values, what do you think about receiving the set of values in place of min, max, and step?

```suggestion
export const Slider = ({ values, defaultValue, handleChange }) => {
```

src/components/DopplerPlus/PlanCalculator.js
andresmoschini: Why do you want to export the Slider?

Regarding default, [they recommend to avoid it](https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/).

Maybe 

```suggestion
export { PlanCalculator };
```

src/components/shared/Slider/Slider.js
```diff
@@ -0,0 +1,16 @@
...
+  // TODO: add styles
+  return (
+    <input
+      style={{ marginTop: '100px', padding: '0' }}
+      type="range"
```
andresmoschini, line 8: Does it work with all browsers that we want to support?

src/components/shared/Slider/Slider.test.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import { Slider } from './Slider';
+
+describe('Stars score component', () => {
+  afterEach(cleanup);
+
+  it('renders stars with score 0', () => {
```
andresmoschini, line 9:
```suggestion
  it('renders starts with score 0', () => {
```

andresmoschini, line 9: If I am not wrong, there is not any assert to verify that the score is `0`

src/components/shared/Slider/Slider.test.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { render, cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import { Slider } from './Slider';
+
+describe('Stars score component', () => {
```
andresmoschini, line 6: I do not understand this name.

src/components/shared/Slider/Slider.test.js
```diff
@@ -0,0 +1,19 @@
...
+  afterEach(cleanup);
+
+  it('renders stars with score 0', () => {
+    // Act
+    const mockedFunction = () => null;
```
andresmoschini, line 11: In this test or another, it would be nice to test how this function is called.

src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -0,0 +1,37 @@
...
+  const valuesPerRange = [15, 29, 48, 77, 106, 145, 240, 340, 460];
```
andresmoschini, line 7: How are you planning to deal with proportionality? or it is not important?

src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -0,0 +1,37 @@
...
+            {/* TODO: change this to intl elemnt */}
+            <h1>Plan STANDARD</h1>
+            <p style={{ paddingBottom: '50px' }}>
+              {/* TODO: change this to intl elemnt */}
+              ¿Cuántos contactos tienes? Utiliza el slider para calcular el costo final de tu Plan
+            </p>
+            <span>US$</span>
+            <span style={{ fontSize: '40px' }}>{state.price}</span>
+            <p>por mes</p>
+            <Slider min={0} max={8} step={1} defaultValue={0} handleChange={updatePrice} />
```
andresmoschini, line 29:
```suggestion
            <Slider min={0} max={valuesPerRange.length} step={1} defaultValue={0} handleChange={updatePrice} />
```

src/components/shared/Slider/Slider.js
andresmoschini, line 3: What do you think about using default values?

```suggestion
export const Slider = ({ min = 0, max = 100, step = 1, defaultValue = 0, handleChange }) => {
```

andresmoschini: In this case, I am not sure how default values would work.

Does it work?

```suggestion
export const Slider = ({ values, defaultValue = values?.lenght && values[0], handleChange }) => {
```

src/components/DopplerPlus/PlanCalculator.js
cbernat: I wanted it to be able to use it outside. That's why.

src/components/shared/Slider/Slider.test.js
cbernat: Me neither O_o

src/components/shared/Slider/Slider.test.js
cbernat: I'm not so sure it should start with 0, it could start at any value. For plans, for example they do not start at 0.

src/components/DopplerPlus/PlanCalculator.js
andresmoschini: But Slider component is already being exported as another module: `components/shared/Slider/Slider`, am I missing something?

src/components/shared/Slider/Slider.js
andresmoschini: Good catch! I think that the `id` should be removed, if I am not wrong, it is not necessary, right?

src/components/DopplerPlus/PlanCalculator.js
cbernat: It is now, before it was all in the same file, this is an old comment.

src/components/shared/Slider/Slider.js
andresmoschini: Or, if you want to make honor to the original API, what do you think about simply pass the parameters?

```suggestion
export const Slider = (props) => (<input 
      style={{ marginTop: '100px', padding: '0' }}
      type="range"
      {...props}
      // . . .
```

src/components/DopplerPlus/PlanCalculator.js
cbernat: I don't think min should be 0 here.

andresmoschini: `0` means the first element of the array, so I think that (at least in this case) it is fine.

src/components/shared/Slider/Slider.test.js
lucianaDilizio: Yes, it could be nice, I have to find a way to shoot an onChange event in this element. 👍

src/components/shared/Slider/Slider.test.js
lucianaDilizio: I have copied it from another test and I forgot to change the assert text, sorry.
Respect to the start value, the Slider could start with any value you want.

src/components/shared/Slider/Slider.js
lucianaDilizio: I think that yes, At least I have tested it in Firefox, Chrome and Edge and it works fine. In Sites they are using the same element.

src/components/shared/Slider/Slider.js
lucianaDilizio: If I use default values it's doesn't have sense use the original API. I'm not sure if it's necessary to use default values.

src/components/DopplerPlus/PlanCalculator.js
lucianaDilizio: It's not important. The important here is values amount that we should mange in the slider.

andresmoschini: But it will, or I am missunderstanding something?

src/components/DopplerPlus/PlanCalculator.js
lucianaDilizio: Yes. maybe I will have to mange other values in the state in the future, but I'm not sure about it, so I think that for the moment it's a nice change.

src/App.js
lucianaDilizio: Yes, you have to be logged to make an upgrade or buy credits.

src/components/DopplerPlus/PlanCalculator.js
andresmoschini: Right, I am not sure if having more than an independent value in the same "state" is a good practice. And, when the values are not independent, I am pretty sure that using `useReducer` is the best approach.

src/components/DopplerPlus/PlanCalculator.js
lucianaDilizio: All the values in the Slider will have the same space between them.

src/components/DopplerPlus/PlanCalculator.js
lucianaDilizio: And `max `should be `valuesPerRange.length - 1` 🤔

src/components/DopplerPlus/PlanCalculator.js
lucianaDilizio: I mean, maybe I could manage a couple of values like: { price: 15, amount: 1500 }. I should apply this logic now maybe.

andresmoschini: in that case, probably using a reducer is a better option than a complex value with `usesState`

src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -0,0 +1,52 @@
...
+    { idPlan: 3, price: 48, amount: 5000 },
+    { idPlan: 4, price: 77, amount: 10000 },
+    { idPlan: 5, price: 106, amount: 15000 },
+    { idPlan: 6, price: 145, amount: 25000 },
+    { idPlan: 7, price: 240, amount: 50000 },
+    { idPlan: 9, price: 340, amount: 75000 },
+    { idPlan: 9, price: 460, amount: 100000 },
+  ];
+
+  const initialPlan = { idPlan: 1, price: 15, amount: 1500 };
```
andresmoschini, line 18:
```suggestion
  const initialPlan = planList[0];
```

src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -0,0 +1,52 @@
...
+
+  const [currentPlan, updateSelectedPlan] = useReducer((currentPlan, index) => {
+    return plansList[index] || currentPlan;
+  }, initialPlan);
```
andresmoschini, line 22:
```suggestion
  const [currentPlan, updateSelectedPlan] = useReducer((currentPlan, index) => {
    return plansList[index] || currentPlan;
  }, plansList[0]);
```

src/components/shared/Slider/Slider.js
lucianaDilizio: By the moment I won't apply any change here.

src/components/shared/Slider/Slider.js
```diff
@@ -0,0 +1,28 @@
+import React, { useState } from 'react';
+import * as S from './Slider.styles';
+
+export const Slider = ({ values, defaultValue, handleChange }) => {
+  const [currentDescription, setCurrentDescription] = useState(values[defaultValue]);
```
cbernat, line 5: This is looking better. 👍

PR #1021, opened 2020-08-04
fix: remove unneccesary image from release script

cbernat:
### Remove using personal repo in webapp
Remove old image hosted in @swidzinskiMS 's account. 
Right now we are using `fromdoppler` repository in swarm, this is not needed.
Also this change will make deploy faster!


andresmoschini:
👏

PR #1024, opened 2020-08-07
feat: add discounts for plans

lucianaDilizio:
### Discounts for plans

- It add a basic behavior to apply discounts.

CDN: https://cdn.fromdoppler.com/doppler-webapp/demo-build2575/#/plan-calculator

image

### DIFF
--- a/src/components/DopplerPlus/PlanCalculator.js
+++ b/src/components/DopplerPlus/PlanCalculator.js
@@ -15,6 +15,13 @@ const PlanCalculator = () => {
     { idPlan: 9, price: 460, amount: 100000 },
   ];
 
+  const discountsList = [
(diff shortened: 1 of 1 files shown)


src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -36,9 +47,32 @@ const PlanCalculator = () => {
...
+                >
+                  {discount.description}
+                </button>
+              ))}
+            </div>
             <span>US$</span>
-            <span style={{ fontSize: '40px' }}>{currentPlan.price}</span>
-            <p>por mes</p>
+            <span style={{ fontSize: '40px' }}>
+              {Math.round(
```
cbernat, line 69: How about adding the value saved with strikethrough? Like here
image
With less styling of course

PR #1031, opened 2020-08-11
feat: on error put app in offline mode

cbernat:
### Add maintenance screen

https://cdn.fromdoppler.com/doppler-webapp/development-build2587/#/offline
Using GetUserData call, that we always do to get user state, if it were non reachable by a 404 or 503 http status code, that is what the load balances gives for a non available resource, put the app offline. 

ezgif com-video-to-gif (1)

image

image

@swidzinskiMS @fgchaio fyi

Pending:

- [x] Add new images
- [x] Add new styles
- [x] Add new text
- [x] Remove chat

image


src/services/app-session.ts
```diff
@@ -27,6 +27,7 @@ interface AuthenticatedAppSessionWithDatahub
 export type AppSession =
   | { status: 'unknown' }
   | { status: 'non-authenticated' }
+  | { status: 'maintenance' }
```
andresmoschini, line 30: Good idea!

src/services/session-manager.ts
```diff
@@ -71,7 +71,11 @@ export class OnlineSessionManager implements SessionManager {
         } as AppSession, // Cast required because TS cannot resolve datahubCustomerId complexity
       );
     } catch (error) {
-      this.updateSession({ status: 'non-authenticated' });
+      if (!!error.response && (error.response.status === 503 || error.response.status === 404)) {
```
lucianaDilizio, line 74: Maybe we can use something like: 
`if ( error?.response.status === 503 || error?.response.status === 404)`
I'm not sure if this works fine.

cbernat: I've tried `error.response?.status` without success 😕

lucianaDilizio: Yes, it seems to be that is in that way.

cbernat: image
🤔

lucianaDilizio: 🤷‍♀️

PR #1032, opened 2020-08-12
Dw 445 double service for plan calculator

lucianaDilizio:
### Plan calculator screen progressing 

**In this pr:**
- Create a getPlansList method in a double service (moving hardcoded info from the element)
- Add parameters in the url: [type plan id], [discount id] and [promocode]
- Add a button with the url to buy, using values chosen by the user in the screen and promocode if exists

**Todo:**
- Obtain real data from the BE or hardcode it. 

URL DEMO: http://cdn.fromdoppler.com/doppler-webapp/demo-build2595/#/plan-calculator/1/buy?discountId=3&promoId=CUMPLE

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -129,7 +129,17 @@ const App = ({ locale, location, dependencies: { appSessionRef, sessionManager }
                 exact
                 component={ReportsPartialsCampaigns}
               />
-              <PrivateRoute path="/plan-calculator" exact component={PlanCalculator} />
+          <PrivateRoute
+            path={[
+              '/plan-calculator/:typePlanId/buy',
+              '/plan-calculator/:typePlanId/promotions/:promoId/buy',
+              '/plan-calculator/:typePlanId/discounts/:discountId/buy',
+              '/plan-calculator/:typePlanId/promotions/:promoId/discounts/:discountId/buy',
+              '/plan-calculator/:typePlanId/discounts/:discountId/promotions/:promoId/buy',
+            ]}
+            exact
+            component={PlanCalculator}
+          />
               <PublicRouteWithLegacyFallback exact path="/login" />
               <PublicRouteWithLegacyFallback exact path="/signup" />
               <PublicRouteWithLegacyFallback exact path="/login/reset-password" />
--- a/src/App.js
+++ b/src/App.js
@@ -129,17 +129,11 @@ const App = ({ locale, location, dependencies: { appSessionRef, sessionManager }
                 exact
                 component={ReportsPartialsCampaigns}
               />
-          <PrivateRoute
-            path={[
-              '/plan-calculator/:typePlanId/buy',
-              '/plan-calculator/:typePlanId/promotions/:promoId/buy',
-              '/plan-calculator/:typePlanId/discounts/:discountId/buy',
-              '/plan-calculator/:typePlanId/promotions/:promoId/discounts/:discountId/buy',
-              '/plan-calculator/:typePlanId/discounts/:discountId/promotions/:promoId/buy',
-            ]}
-            exact
-            component={PlanCalculator}
-          />
+              <PrivateRoute
+                path={'/plan-calculator/:typePlanId/buy'}
+                exact
+                component={PlanCalculator}
+              />
               <PublicRouteWithLegacyFallback exact path="/login" />
               <PublicRouteWithLegacyFallback exact path="/signup" />
               <PublicRouteWithLegacyFallback exact path="/login/reset-password" />
--- a/src/components/DopplerPlus/PlanCalculator.js
+++ b/src/components/DopplerPlus/PlanCalculator.js
@@ -4,29 +4,51 @@ import { InjectAppServices } from '../../services/pure-di';
 import { Loading } from '../Loading/Loading';
 import { useRouteMatch } from 'react-router-dom';
 import { useIntl } from 'react-intl';
+import queryString from 'query-string';
+import { extractParameter } from '../../utils';
 
-const PlanCalculator = ({ dependencies: { dopplerLegacyClient } }) => {
+const PlanCalculator = ({ location, dependencies: { dopplerLegacyClient } }) => {
+  const safePromoId = extractParameter(location, queryString.parse, 'promoId') || '';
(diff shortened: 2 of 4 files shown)


src/App.js
```diff
@@ -125,7 +125,17 @@ const App = ({ locale, location, dependencies: { appSessionRef, sessionManager }
...
             component={ReportsPartialsCampaigns}
           />
-          <PrivateRoute path="/plan-calculator" exact component={PlanCalculator} />
+          <PrivateRoute
+            path={[
+              '/plan-calculator/:typePlanId/buy',
+              '/plan-calculator/:typePlanId/promotions/:promoId/buy',
+              '/plan-calculator/:typePlanId/discounts/:discountId/buy',
+              '/plan-calculator/:typePlanId/promotions/:promoId/discounts/:discountId/buy',
+              '/plan-calculator/:typePlanId/discounts/:discountId/promotions/:promoId/buy',
```
andresmoschini, line 134: It is nice but in my opinion, having a only way (or at least a canonical way) to express a URL is better. 

Remember that the good URL is the URL that does not change, when you have a lot of URLs you should make honor to them.

Also, if promotion and discounts are optional and there is not a hierarchical relation between them, maybe it is better representing them as query string parameters in place of URL segments, ie:

```
/plan-calculator/PlaABC/discounts/VetteranDiscount/promotions/ChristmassPromotion/buy

vs 

/plan-calculator/PlaABC/buy?discount=VetteranDiscount&promotion=ChristmassPromotion
```

src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -1,42 +1,67 @@
...
+    },
+    { idPrice: 0, price: 0, amount: 0 },
+  );
+
+  const [currentDiscount, applyDiscount] = useReducer(
+    (discount, index) => {
+      return state.discountsList[index] || discount;
+    },
+    { id: 0, percent: 0, monthsAmmount: 0, description: '' },
+  );
```
andresmoschini, line 30: Probably it should be represented with only one reducer.

src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -1,42 +1,67 @@
...
+      const responsePlansList = await dopplerLegacyClient.getPlansList(typePlanId);
+      if (responsePlansList.success) {
+        setState({
+          loading: false,
+          planList: responsePlansList.planList,
+          discountsList: responsePlansList.discounts,
+          success: true,
+        });
+        setSelectedPlan(0);
+        applyDiscount(safeDiscountId);
```
andresmoschini, line 44: In my opinion these three operations should be done as a single reducer operation.

src/components/DopplerPlus/PlanCalculator.js
cbernat: Yes I agree two reducers seem a bit odd, perhaps there is a better way of achieving what you seek...

lucianaDilizio: Yes, I will try to find a way to simplify this.

src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -4,59 +4,80 @@ import { InjectAppServices } from '../../services/pure-di';
...
+            ? responsePlansList.discounts.find((discount) => {
+                return discount.id === discountId;
+              })
+            : responsePlansList.discounts[0],
+        });
         setState({
           loading: false,
           planList: responsePlansList.planList,
           discountsList: responsePlansList.discounts,
           success: true,
```
andresmoschini: No estoy seguro, pero tendrá sentido centralizar todo el estado en un solo reducer en lugar de tener el _useReducer_ y _useState_?

lucianaDilizio: I'm not sure neither. One is related with current values and things that may be changed, and the other with information from BE and loading state and so on, but are things that once you have loaded it won't change. It will be a too big reducer if we manage all that states there, but I don't know.

cbernat: Capaz tiene sentido lo que dice, pensalo 😄 los datos de plan mas alla de que los pidas una sola vez estan asociados a la seleccion de plan, podrias probar a ver que pasa, y si no funciona descartar la idea.

lucianaDilizio: Lo estuvimos hablando con Andres y ambos creemos que no esta mal tener los dos estados por separado, por un lado los datos fijos y por otro lo que va modificando el usuario.

Ahora voy a aplicar un pequeño cambio para evitar tener tantas dependencias en el useEffect, aunque en realidad me desligo de una sola, pero cuantas menos tengamos mejor.

src/components/DopplerPlus/PlanCalculator.js
```diff
@@ -48,48 +92,75 @@ const PlanCalculator = () => {
...
             )}
             <span>US$</span>
             <span style={{ fontSize: '40px' }}>
               {Math.round(
-                currentPlan.price *
-                  (1 - currentDiscount.percent / 100) *
-                  currentDiscount.monthsAmmount,
+                planData.plan.price *
+                  (1 - planData.discount.percent / 100) *
+                  planData.discount.monthsAmmount,
```
andresmoschini, line 126: Maybe these (and other) calculous could be done in the reducer, to simplify the template.

lucianaDilizio: I'm not sure, but it seems to be that it's not too easy as appears. 🤔

andresmoschini: Ok, by the moment it is fine, we can refactor it later if we consider that it could be useful.

cbernat:
Take in account that we have two pages for the same section in different folders: 
image
Also I don't think that DopplerPlus is the correct name for a component folder... PlanCalculator would be a more consisten name with what we have.

cbernat:
And later on the url can change into something like this for upgrade for example  `plans/types/:type/calculate` --> `plans/types/standard-monthly/calculate` then this could evolve for changing plans into something like this `plans/types/:type/calculate?currentPlan=:planId` This is not for right now but for us to think in background.

PR #1041, opened 2020-08-19
DW-452 add Doppler plan service

cbernat:
### Add Doppler Plan Service
- Add call to get all raw data in doppler legacy client
- Add service to cache the request and admin petitions:
* get featured plan: get a standard plan data and it´s correspondant in plus and pro based on amount of emails or subscribers
* get plans filtered by type for slider: can get all standard high volume, or all plus high volume, or all plus, here we have two parameters plan type: (free | standard | plus | pro | enterprise), user type: usual user types we have in doppler.


src/services/doppler-legacy-client.ts
```diff
@@ -371,6 +454,15 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
     return mapHeaderDataJson(response.data);
   }
 
+  public async getAllPlans(): Promise<PlanModel[]> {
+    const response = await this.axios.get('/WebApp/GetAllPlans');
+    if (!response || !response.data) {
```
lucianaDilizio, line 459: Here maybe you can use the magnificent `!response?.data` 🤔

PR #1062, opened 2020-09-01
fix: link to partials-campaigns when sending now

lucianaDilizio:
Fix issue in subscriber sent campaigns report when the campaign is sending right now and the campaing don't have reports yet.
To fix this, when a user clicks in a campaign that it is sending now, it will be linked to partial-campaigns url.

### DIFF
--- a/src/services/doppler-api-client.ts
+++ b/src/services/doppler-api-client.ts
@@ -50,6 +50,7 @@ interface CampaignDelivery {
   campaignSubject: string;
   clicksCount: number;
   deliveryStatus: string;
+  isSendingNow: boolean;
   urlImgPreview: string;
 }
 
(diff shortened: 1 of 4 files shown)


src/services/doppler-api-client.ts
```diff
@@ -163,6 +164,7 @@ export class HttpDopplerApiClient implements DopplerApiClient {
       campaignSubject: x.campaignSubject,
       clicksCount: x.clicksCount,
       deliveryStatus: x.deliveryStatus,
+      isSendingNow: x.isSendingNow,
```
cbernat, line 167: Beware here, right now api is not deployed, so I guess this value will be undefined. Did you take it in account?

lucianaDilizio: Yes, this value will be undefined and it will use the reports url because of this validation:

image

PR #1074, opened 2020-09-16
DW-464 Change plan first screen with real data

cbernat:
### Change plan first screen
- Add get paths and map into cards
- Add features by path as text
- Add titles and descriptions based on path type
- Add actions based on type
- Add get current plan fee and feature types based on plan type and amount of subscribers/emails asuming feature standard (I´m not so sure about this)

https://cdn.fromdoppler.com/doppler-webapp/build2684/#/plan-selection
https://cdn.fromdoppler.com/doppler-webapp/int-build2684/#/plan-selection

_free or demo_
image

_By subscribers_
image

_Prepaid users_ 
image

_Monthly for higher plan_
image

_By subscribers with possible plus upgrade_
image

*Done*

- [x]  Refactor change plan component to be more readable
- [x]  Fix some intermediate states (Actual with no more actions possible)
- [x]  Get plan id in doppler BE and use it here PR --> https://github.com/MakingSense/Doppler/pull/7299


src/components/ChangePlan/ChangePlan.js
```diff
@@ -1,58 +1,145 @@
...
+            featureSet: 'agency',
+          };
+        default: {
+        }
+      }
+    };
+    const fetchData = async () => {
+      setState({ loading: true });
+      const planList = await planService.getPlanList();
+      const currentPlan = mapCurrentPlan(appSessionRef.current.userData.user.plan, planList);
```
lucianaDilizio, line 119: Why don't get the current plan from the service instead mapping it here?

cbernat: Inside `mapCurrentPlan` function there is a _getting from the service_  by id. All the information regarding actual plan is in the header and user data.

lucianaDilizio: Ok, surely I'm missing something, but, what about to send this `appSessionRef.current.userData.user.plan` to the service and manage the mapping there? Maybe I have a wrong concept and it's not the best way to manage this in a service but I think that would be clearer.

src/components/ChangePlan/ChangePlan.js
```diff
@@ -1,58 +1,145 @@
...
+    loading: true,
+    isFeaturesVisible: false,
+  });
+  useEffect(() => {
+    const mapCurrentPlan = (plan, planList) => {
+      const exclusivePlan = { type: 'exclusive' };
+      switch (plan.planType) {
+        case 'demo':
+        case 'free':
+          return {
```
lucianaDilizio, line 95: What if we always return demo and free as a default value to avoid have the default empty. 🤔 

```
case 'demo':
case 'free':
default: 
return {
     type: 'free',
     subscriberLimit: 500,
     featureSet: 'free',
}
```

or just:

```
default: 
return {
     type: 'free',
     subscriberLimit: 500,
     featureSet: 'free',
}
```

src/components/ChangePlan/ChangePlan.js
cbernat: Maybe you are right, I will give it a try.

src/components/ChangePlan/ChangePlan.js
```diff
@@ -1,58 +1,145 @@
...
+const bigDataBullet = (tooltipText) => {
+  return (
+    <div className="dp-tooltip-container">
+      <span className="dp-icobd">BD</span>
+      <div className="dp-tooltip-block">
+        <span className="tooltiptext">{tooltipText}</span>
+      </div>
+    </div>
+  );
+};
```
andresmoschini, line 77: Esto es raro, por que son funciones y no componentes?

src/components/ChangePlan/ChangePlan.js
```diff
@@ -61,78 +148,87 @@ const ChangePlan = ({ location }) => {
...
+                        {_('change_plan.calculate_price')}
+                      </CardAction>
+                    )}
+                    {state.isFeaturesVisible ? (
+                      <CardFeatures>
+                        <h4>{getFeatureTitleByType(path.type)}</h4>
+                        <BulletOptions type={path.type} />
+                      </CardFeatures>
+                    ) : null}
+                  </Card>
```
andresmoschini, line 220: Creo que estaría bien meter todo esto en un componente que represente el card de un path.

src/components/ChangePlan/ChangePlan.js
```diff
@@ -61,78 +148,87 @@ const ChangePlan = ({ location }) => {
...
+              )}
             </div>
           </div>
-          <div className="dp-align-center p-t-30 p-b-30">
+          <div className="p-t-30 p-b-30">
             <button
               className="dp-compare-details-plans"
               onClick={() => {
-                setIsFeaturesVisible(!isFeaturesVisible);
+                setState({ ...state, isFeaturesVisible: !state.isFeaturesVisible });
```
andresmoschini, line 231: Creo que un reducer funcionaría mucho mejor que state.

src/i18n/es.js
```diff
@@ -34,20 +34,47 @@ const urlMasterSubscriber = `${urlDopplerLegacy}/Lists/MasterSubscriber/`;
...
+    features_HTML_free: `<option>Atención al Cliente</option>
+    <option>Campañas Convencionales</option>
+    <option>Campañas Automation</option>`,
+    features_HTML_plus: `<star>Atención preferencial</star>
+    <newOption>Detener Campañas</newOption>
+    <option>Landing pages con dominio personalizado</option>
+    <newBigData>Recomendación de Asuntos</newBigData>`,
+    features_HTML_standard: `<option>Acceso a la API de Doppler</option>
+    <option>Campañas Automation de Comportamiento en Sitio</option>
+    <option>Envíos con tu firma personalizada</option>`,
```
andresmoschini, line 63: Muy piolas los elementos custom para que quien edita el contenido pueda representar los diferentes items de diferente manera!

andresmoschini, line 63: En general en los contenidos multilinea los estabamos comenzando junto al margen con una interlinea anterior:

```suggestion
    features_HTML_standard: `
<option>Acceso a la API de Doppler</option>
<option>Campañas Automation de Comportamiento en Sitio</option>
<option>Envíos con tu firma personalizada</option>`,
```

Ejemplo:
image

src/doppler-types.ts
```diff
@@ -68,12 +74,14 @@ export interface PrepaidPack {
 }
 
 export interface FreePlan {
+  id: number;
```
andresmoschini, line 77: Does the freePlan have an id? how is it useful for use?

src/components/ChangePlan/ChangePlan.js
```diff
@@ -1,58 +1,145 @@
...
+  });
+  useEffect(() => {
+    const mapCurrentPlan = (plan, planList) => {
+      const exclusivePlan = { type: 'exclusive' };
+      switch (plan.planType) {
+        case 'demo':
+        case 'free':
+          return {
+            type: 'free',
+            subscriberLimit: 500,
```
andresmoschini, line 97: No está bueno que esté hardcodeado, y si lo está tal vez deberíamos ponerlo en los archivos de constantes, creo que eso permitiría reusarlo en los contenidos, por ejemplo aquí:

image

src/components/ChangePlan/ChangePlan.js
andresmoschini: Incluso, si ese componente se definiera usano typescript nos ayudaría a sacar ventaja de los tipos y confirmar que no se nos escapa nada.

src/doppler-types.ts
```diff
@@ -13,7 +13,13 @@ export type EmptyResultWithoutExpectedErrors = { success: true } | UnexpectedErr
...
 export type PathType = 'free' | 'standard' | 'plus' | 'agencies';
 
-export type PlanType = 'free' | 'prepaid' | 'monthly-deliveries' | 'subscribers';
+export type PlanType =
+  | 'free'
+  | 'prepaid'
+  | 'monthly-deliveries'
+  | 'subscribers'
+  | 'agencies'
+  | 'demo';
```
andresmoschini, line 22: Es `demo` un tipo de plan? como entra en los paths existentes?

src/components/ChangePlan/ChangePlan.js
cbernat: I think of this functions as components, it´s true I'm using them as functions I will fix it.

src/components/ChangePlan/ChangePlan.js
cbernat: Si tambien lo venia pensando pq hay muchos ifs con type, medio que depende bastante del `path.type`

src/components/ChangePlan/ChangePlan.js
andresmoschini: Creo que para usarlas como componentes los nombres tienen que comenzar con mayusculas, y los parámetros tienen que entra como props (o childs)

src/i18n/es.js
cbernat: Pero esos son markup y mis textos son un casi html, estaría bien igual?

andresmoschini: creo que si

src/doppler-types.ts
cbernat: To look for current plan by id. [Here](https://github.com/FromDoppler/doppler-webapp/pull/1074/files#diff-5710472294fa4b50ebeeb68baed8547aR103)

src/doppler-types.ts
cbernat: Demo is a type that overlaps with free, it's a free with some other functionalities enabled. Demo is supposed to behave as free in paths.

src/doppler-types.ts
andresmoschini: But, you already have the plan, why did you search it by id?

```js
const monthlyPlan = planService.getPlanById(plan.idPlan, planList);
//                                          ^^^^ this is the plan
```

src/doppler-types.ts
andresmoschini: No estoy seguro de que sea buena idea meterlo como un tipo, nos hace el modelo más complejo.

Por ejemplo, tenés tipos para los planes free, pero me parece que no tenés para el demo.

src/doppler-types.ts
cbernat: I don´t have all plan data in session information, the name is wrong it should be session instead of plan perhaps. The only data I have for plan is: id, description, subscribers by month or emails, remaining credits, and that´s all, I need the fee for paths to work and to compare.

andresmoschini: I understand, it is a valid solution, but it is still making noise to me.

I would prefer something like `getPlanBySessionPlan()`, if we need that the session plan is the free plan, we should return the free plan, without worry about the ID. But, I am not sure, maybe this approach has worse problems.

cbernat: I will reformulate this, yes its weird, thanks!.

src/doppler-types.ts
cbernat: I will think a way of doing without this demo type, seems about right.

src/components/ChangePlan/ChangePlan.js
cbernat: Adding `mapCurrentPlan` into the service would imply mapping the session type, or taking this as an any. This is not so clean. And this session plan data has nothing to do with the plan structure we are handling, this is why I´m doing the mapping in the first place, here and outside the plan service.

lucianaDilizio: Ok. And what about move this function outside useEffect?

cbernat: If I move it outside I need to add it in the dependency array. It was a compiler suggestion to put it inside of `useEffect` hook

lucianaDilizio: Oh 😞 it's true. I'm not convinced about this but it's seems to be that we have no choice.

src/i18n/en.js
```diff
@@ -34,20 +34,49 @@ const urlMasterSubscriber = `${urlDopplerLegacy}/Lists/MasterSubscriber/`;
 
 export default {
   change_plan: {
-    all_of_plan: `Everything from {plan} Plus:`,
+    agencies_icon: `https://cdn.fromdoppler.com/doppler-ui-library/${dopplerUiLibraryVersion}/img/icono-agencias.svg`,
+    all_of_plan: `Everything from {plan} plus:`,
+    ask_demo: 'Ask for a demo',
```
cbernat, line 39: @mrunge86  fyi. Contenidos temporales de las tarjetas para doppler plus.

PR #1083, opened 2020-09-22
fix: replace version for ui library js script

cbernat:
Added versioning for js file in ui library for webapp

Replaced old v3.38.0 for current version.

image


lucianaDilizio (commented):
👍

PR #1084, opened 2020-09-22
feat: agencies form page

lucianaDilizio:
### Agencies form page

First version for agencies form page.

To do: 
- apply content (added as comment in the code)
- add check to accept privacy policies (added as comment in the code)
- define if it's necessary to autocomplete data beyond the email (that implies changes from BE)
- redefine the url if it's necessary

image

image

image


src/services/doppler-legacy-client.ts
```diff
@@ -689,6 +703,27 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
...
+    const response = await this.axios.post('/WebApp/RequestAgenciesDemo', {
+      Email: model.email,
+      Firstname: model.firstname,
+      Lastname: model.lastname,
+      Phone: model.phone,
+      ContactSchedule: model.range_time,
+      SendingVolume: model.volume,
+    });
+
+    if (!response.data.success) {
```
lucianaDilizio, line 718: I have a doubt here because if something wrong happens in BE (a problem with the api or a problem sending the email) I won't show the success message (and we don't have an error message neither). 
It doesn't really matters if, for example, the subscriber isn't added to the Doppler account list but we send the email anyway. 
It won't be nice to show a success message always... or yes? 🤔 

My doubt is: when I should show the success message?

lucianaDilizio: Another thing... what happens if the email that I want to register is in a black list? It's ok to send the email to unbounce in that case? 🤔

cbernat: I'm not sure if the email is in black list how the process will be handled, perhaps @javierburger will know. Is there an interceptor for this cases or should we look into the blacklist? Either way this is not done here but in old doppler.

cbernat: > It doesn't really matters if, for example, the subscriber isn't added to the Doppler account list but we send the email anyway.

I would say in most cases like this, like in reset password we will always tell the user that all is ok. But it may fail.

lucianaDilizio: Ok, it won't be normal to fail in both actions, so well, I will always show the success message (in the worst case the user won't be never called 😋)

src/components/DopplerPlus/AgenciesForm.js
```diff
@@ -0,0 +1,272 @@
...
+const AgenciesForm = ({ dependencies: { dopplerLegacyClient, appSessionRef } }) => {
+  const intl = useIntl();
+  const _ = (id, values) => intl.formatMessage({ id: id }, values);
+
+  // TODO: apply content
+  // TODO: add check to accept privacy policies
+
+  const volumeOptions = [
+    {
+      id: 'lessThan500k',
```
cbernat, line 27: Why not add all this options inside resource file? Check that `value` and `description` are the same. 🤔

src/components/DopplerPlus/AgenciesForm.js
```diff
@@ -0,0 +1,272 @@
...
+        <div className="col-sm-12 col-md-12 col-lg-12">
+          <nav className="dp-breadcrumb">
+            <ul>
+              <li>
+                <a href={_('agencies.breadcrumb_url')}>{_('agencies.breadcrumb')}</a>
+              </li>
+              <li>{_('agencies.title')}</li>
+            </ul>
+          </nav>
+          <h2>Plan para agencias</h2>
```
cbernat, line 106: Even if you don{t have all correct contents, perhaps adding this into resource file is better for translation later. Or you can create a task for contents. Either way it is nice that each screen is as complete as it can be. (english/spanish)

src/components/DopplerPlus/AgenciesForm.js
lucianaDilizio: When you say resource file you mean utils.js? 

The value not necessarily has to be equal to the description. The description will be a translation.

src/components/DopplerPlus/AgenciesForm.js
lucianaDilizio: I have created a task for that. Due to I don't know what to put in this texts, I preferred to have something usefully in production to other people can see this.

cbernat: It is best that we try that all uploaded screens are as finished as it can be. So in english and spanish with some content, most of all because this is nearly finished. It's just an opinion though. 😄

PR #1086, opened 2020-09-25
DW-458 slider real data

cbernat:
Applied real data into slider.
For now it works only by url, not by selecting the types in tabs.

image

For now type only changes by url as this for example:
https://cdn.fromdoppler.com/doppler-webapp/int-build2701/#/plan-selection/standard/prepaid
https://cdn.fromdoppler.com/doppler-webapp/int-build2701/#/plan-selection/standard/subscribers
https://cdn.fromdoppler.com/doppler-webapp/int-build2701/#/plan-selection/standard/monthly-deliveries

Demo for free user for example:
https://cdn.fromdoppler.com/doppler-webapp/demo-build2701/#/plan-selection/standard/monthly-deliveries


src/components/ChangePlan/ChangePlan.js
```diff
@@ -159,7 +159,7 @@ const AgenciesCard = ({ showFeatures }) => {
   );
 };
 
-const CardWithPrice = ({ path, showFeatures, currentPlanType, promoCode }) => {
+const CardWithPrice = ({ path, showFeatures, avaiablePlanType, promoCode }) => {
```
cbernat, line 162: Why rename this to `avaiablePlanType` if this ought to be the currently selected plan type 🤔

src/components/ChangePlan/ChangePlan.js
```diff
@@ -272,7 +274,10 @@ const ChangePlan = ({ location, dependencies: { planService, appSessionRef } })
                         key={index}
                         path={path}
                         showFeatures={isFeaturesVisible}
-                        currentPlanType={state.currentPlan.type}
+                        avaiablePlanType={planService.getPreSelectedPlanTypePerPathType(
```
cbernat, line 277: Why add a preselected type here? the issue here is that if user plan is prepaid, logically we follow that type to be selected, the only case that there is no type to be preselected would be for `free` only. For `free` we could select the first and that's it.

src/services/plan-service.ts
```diff
@@ -296,4 +296,19 @@ export class PlanService implements PlanHierarchy {
         };
     }
   };
+
+  getPreSelectedPlanTypePerPathType = (planType: PlanType, pathType: PathType) => {
```
cbernat, line 300: I don´t think it's ok to add a mapper for old types to new in this service, perhaps in utils, like the rest of the mappers? 🤔

src/components/ChangePlan/ChangePlan.js
cbernat: @lucianaDilizio

src/components/ChangePlan/ChangePlan.js
cbernat: @lucianaDilizio

src/services/plan-service.ts
cbernat: @lucianaDilizio

src/components/ChangePlan/ChangePlan.js
lucianaDilizio: I did this because the planType for free depends of the path. Since we don't have prepaid plans for PLUS, I made a function to return what is the basic plan for a specific plan and path. 
This works for prepaid plans too. If I have a prepaid plan and I want to hire a plus plan, I don't have prepaid plans there so I have to preselect a monthly plan.

cbernat:
This PR is now being finished in @lucianaDilizio PR https://github.com/FromDoppler/doppler-webapp/pull/1101 I´m closing this...

PR #1087, opened 2020-09-28
DAT-128-Add invoices list

jhoffman-ms:
Changes:

- Included new item in the "new features". The sections is "Facturas emitidas":

  
image

- Created new component to shows the invoices given an idUser:

image

Task: [DAT-128](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-128)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -24,6 +24,7 @@ import SubscribersLegacyUrlRedirect from './components/Reports/Subscribers/Subsc
 import ChangePlan from './components/ChangePlan/ChangePlan';
 import PlanCalculator from './components/DopplerPlus/PlanCalculator';
 import AgenciesForm from './components/DopplerPlus/AgenciesForm';
+import InvoicesList from './components/InvoicesList/InvoicesList';
 
 /**
  * @param { Object } props - props
@@ -126,6 +127,7 @@ const App = ({ locale, location, dependencies: { appSessionRef, sessionManager }
                 exact
                 component={ReportsPartialsCampaigns}
               />
+              <PrivateRoute path={['/billing/invoices']} exact component={InvoicesList} />
               <PublicRouteWithLegacyFallback exact path="/login" />
               <PublicRouteWithLegacyFallback exact path="/signup" />
               <PublicRouteWithLegacyFallback exact path="/login/reset-password" />
--- a/src/components/InvoicesList/InvoicesList.js
+++ b/src/components/InvoicesList/InvoicesList.js
@@ -1,0 +1,118 @@
+import React, { useState, useEffect } from 'react';
+import { Loading } from '../Loading/Loading';
+import { InjectAppServices } from '../../services/pure-di';
+import { Helmet } from 'react-helmet';
+import HeaderSection from '../shared/HeaderSection/HeaderSection';
+import { FormattedDate, FormattedMessage, useIntl } from 'react-intl';
+
+const InvoicesList = ({ dependencies: { dopplerBillingApiClient } }) => {
+  const [stateInvoices, setStateInvoices] = useState({ loading: true });
+  const intl = useIntl();
+  const _ = (id, values) => intl.formatMessage({ id: id }, values);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      setStateInvoices({ loading: true });
+      const invoicesResponse = await dopplerBillingApiClient.getInvoices(0, 0);
+      if (!invoicesResponse.success) {
+        setStateInvoices({ loading: false, items: [], success: false });
+      } else {
+        setStateInvoices({
+          loading: false,
+          success: true,
+          items: invoicesResponse.value.items,
+          totalItems: invoicesResponse.value.totalItems,
+        });
+      }
+    };
+
+    fetchData();
+  }, [dopplerBillingApiClient]);
+
+  return (
+    <>
+      <Helmet>
+        <title>{_('invoices_list.title')}</title>
+        <meta name="invoices" />
+      </Helmet>
+      <HeaderSection>
+        <div className="col-sm-12 col-md-12 col-lg-12">
+          <h2>
+            <FormattedMessage id="invoices_list.title" />
+          </h2>
+          <p>
+            <FormattedMessage id="invoices_list.sub_title" />
+          </p>
+        </div>
+      </HeaderSection>
+      <section className="dp-container">
+        <div className="dp-table-responsive">
+          {stateInvoices.loading ? (
+            <Loading page />
+          ) : !stateInvoices.success ? (
+            <div class="dp-msj-error dpsg-slow-animation">
+              <p>
+                <FormattedMessage id="invoices_list.error_msg" />
+              </p>
+            </div>
+          ) : (
+            <table className="dp-c-table">
+              <thead>
+                <tr>
+                  <th scope="col">
+                    <FormattedMessage id="invoices_list.account_id_column" />
+                  </th>
+                  <th scope="col">
+                    <FormattedMessage id="invoices_list.product_column" />
+                  </th>
+                  <th scope="col">
+                    <FormattedMessage id="invoices_list.date_column" />
+                  </th>
+                  <th scope="col">
+                    <FormattedMessage id="invoices_list.currency_column" />
+                  </th>
+                  <th scope="col">
+                    <FormattedMessage id="invoices_list.amount_column" />
+                  </th>
+                  <th scope="col"></th>
+                </tr>
+              </thead>
+              <tbody>
+                {stateInvoices.totalItems > 0 ? (
+                  <>
+                    {stateInvoices.items.map((invoice, index) => {
+                      return (
+                        <tr key={index}>
+                          <td>{invoice.accountId}</td>
+                          <td>{invoice.product}</td>
+                          <td>
+                            <FormattedDate value={invoice.date} />
+                          </td>
+                          <td>{invoice.currency}</td>
+                          <td>{invoice.amount}</td>
+                          <td>
+                            <FormattedMessage id="invoices_list.download_msg" />
+                          </td>
+                        </tr>
+                      );
+                    })}
+                  </>
+                ) : (
+                  <tr>
+                    <td colSpan={6}>
+                      <span className="bounceIn">
+                        <FormattedMessage id="invoices_list.no_data_msg" />
+                      </span>
+                    </td>
+                  </tr>
+                )}
+              </tbody>
+            </table>
+          )}
+        </div>
+      </section>
+    </>
+  );
+};
+
+export default InjectAppServices(InvoicesList);
--- a/src/components/InvoicesList/InvoicesList.test.js
+++ b/src/components/InvoicesList/InvoicesList.test.js
@@ -1,0 +1,115 @@
+import React from 'react';
+import { render, cleanup, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import IntlProvider from '../../i18n/DopplerIntlProvider.double-with-ids-as-values';
+import { AppServicesProvider } from '../../services/pure-di';
+import { BrowserRouter } from 'react-router-dom';
+import InvoicesList from './InvoicesList';
+
+describe('InvoiceList component', () => {
+  afterEach(cleanup);
+
+  it('should show invocies', async () => {
+    // Arrange
+    const invoicesCollection = {
+      items: [
+        {
+          accountId: 'CD0000000073689',
+          product: 'Prod 1',
+          date: new Date('2020-09-25T00:00:00'),
+          currency: 'ARS',
+          amount: 1500,
+          filename: 'invoice_2020-10-05_10.pdf',
+        },
+        {
+          accountId: 'CD0000000073690',
+          product: 'Prod 2',
+          date: new Date('2020-09-26T00:00:00'),
+          currency: 'ARS',
+          amount: 1500,
+          filename: 'invoice_2020-10-05_10.pdf',
+        },
+      ],
+      totalItems: 2,
+    };
+
+    const dopplerBillingApiClientDouble = {
+      getInvoices: async () => {
+        return { success: true, value: invoicesCollection };
+      },
+    };
+
+    // Act
+    const { getByText } = render(
+      <AppServicesProvider
+        forcedServices={{
+          dopplerBillingApiClient: dopplerBillingApiClientDouble,
+        }}
+      >
+        <IntlProvider>
+          <BrowserRouter>
+            <InvoicesList />
+          </BrowserRouter>
+        </IntlProvider>
+      </AppServicesProvider>,
+    );
+    // Assert
+    await waitFor(() => expect(getByText('CD0000000073689')).toBeInTheDocument());
+  });
+
+  it('should show empty message', async () => {
+    // Arrange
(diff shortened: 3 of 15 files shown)


src/components/InvoicesList/InvoicesList.js
```diff
@@ -0,0 +1,96 @@
...
+      }
+    };
+
+    fetchData();
+  }, [dopplerBillingApiClient]);
+
+  return (
+    <>
+      <Helmet>
+        <meta name="invoices" />
```
fcoronelmakingsense, line 31: maybe need add title too

src/components/InvoicesList/InvoicesList.js
```diff
@@ -0,0 +1,96 @@
...
+      </Helmet>
+      <HeaderSection>
+        <div className="col-sm-12 col-md-12 col-lg-12">
+          <h2>Facturas emitidas</h2>
```
fcoronelmakingsense, line 35: only use 1 language because its for us? or why?

src/components/NewFeatures/NewFeatures.js
```diff
@@ -63,6 +63,12 @@ const NewFeatures = () => {
             </Link>
           </div>
         </div>
+        <div className="dp-block-wlp dp-box-shadow m-t-36">
+          <div style={{ marginLeft: '20px' }}>
```
GusBaamonde, line 67: maybe I could use this className = "m-l-18" o "m-l-24" and thus not use inline style. the atomic classes of the library jump from 6px to 6px

src/components/InvoicesList/InvoicesList.js
jhoffman-ms: We will suport different language in the future PRs. It was initial PR with minimal functionallity. Thanks for your feedback

src/components/InvoicesList/InvoicesList.js
jhoffman-ms: 👍

src/components/NewFeatures/NewFeatures.js
jhoffman-ms: 👍

src/components/InvoicesList/InvoicesList.test.js
```diff
@@ -0,0 +1,82 @@
...
+        }}
+      >
+        <BrowserRouter>
+          <InvoicesList />
+        </BrowserRouter>
+      </AppServicesProvider>,
+    );
+
+    // Assert
+    await waitFor(() => expect(getByText('No hay facturas para mostrar')).toBeInTheDocument());
```
lucianaDilizio, line 80: I have read that you are not going to add translations now. 
However, maybe it should be better to have the translations structure although we don't have the final content yet.
Beyond that, if you are not going to apply this changes here, keep in mind, that this text should be with the translation key to avoid problems if the text changes in a future.

jhoffman-ms: Luciana, I will include the translation in the current PR. Thanks for your feedback

src/App.js
```diff
@@ -139,6 +140,7 @@ const App = ({ locale, location, dependencies: { appSessionRef, sessionManager }
                 to="/forgot-password?lang=es"
               />
               <RedirectWithQuery exact from="/forgot-password" to="/login/reset-password" />
+              <PrivateRoute path={['/invoices-list']} exact component={InvoicesList} />
```
lucianaDilizio, line 143: Minor: Maybe you could move this line to keep all PrivateRoute elements together.

jhoffman-ms: 👍

jhoffman-ms:
PR updated with the feedback

src/App.js
```diff
@@ -126,6 +127,7 @@ const App = ({ locale, location, dependencies: { appSessionRef, sessionManager }
                 exact
                 component={ReportsPartialsCampaigns}
               />
+              <PrivateRoute path={['/invoices-list']} exact component={InvoicesList} />
```
andresmoschini, line 130: Maybe `billing/invoices`?

src/components/InvoicesList/InvoicesList.js
```diff
@@ -0,0 +1,113 @@
...
+      }
+    };
+
+    fetchData();
+  }, [dopplerBillingApiClient]);
+
+  return (
+    <>
+      <Helmet>
+        <title>{'Invoices'}</title>
```
andresmoschini, line 32: Please, translate this term.

src/components/InvoicesList/InvoicesList.js
```diff
@@ -0,0 +1,113 @@
...
+          <h2>
+            <FormattedMessage id="invoices_list.title" />
+          </h2>
+          <p>
+            <FormattedMessage id="invoices_list.sub_title" />
+          </p>
+        </div>
+      </HeaderSection>
+      <section className="dp-container">
+        {/* <div> */}
```
andresmoschini, line 46: What is this?

src/components/InvoicesList/InvoicesList.js
```diff
@@ -0,0 +1,113 @@
...
+                    <td>
+                      <span className="bounceIn">
+                        <FormattedMessage id="invoices_list.no_data_msg" />
+                      </span>
+                    </td>
+                    <td></td>
+                    <td></td>
+                    <td></td>
+                    <td></td>
+                    <td></td>
```
andresmoschini, line 101: What are these empty elements?

andresmoschini: Oh, it is a kind of patch because you do not have more columns. Maybe we do not need the table when there are not elements. @GusBaamonde how are we dealing with it in other cases?

src/components/InvoicesList/InvoicesList.js
```diff
@@ -0,0 +1,113 @@
...
+
+const InvoicesList = ({ dependencies: { dopplerBillingApiClient } }) => {
+  const [stateInvoices, setStateInvoices] = useState({ loading: true });
+
+  useEffect(() => {
+    const fetchData = async () => {
+      setStateInvoices({ loading: true });
+      const invoicesResponse = await dopplerBillingApiClient.getInvoices(0, 0);
+      if (!invoicesResponse.success) {
+        setStateInvoices({ loading: false, items: [] });
```
andresmoschini, line 16: Warning!, when there is an error, it says that "there are no invoices to show", I think that it is not true.

src/services/doppler-billing-api-client.ts
```diff
@@ -0,0 +1,106 @@
...
+      !connectionData ||
+      connectionData.status !== 'authenticated' ||
+      !connectionData.jwtToken ||
+      !connectionData.userData
+    ) {
+      throw new Error('Doppler Billing API connection data is not available');
+    }
+    return {
+      jwtToken: connectionData.jwtToken,
+      idUser: connectionData ? connectionData.userData.user.idUser : '',
```
andresmoschini, line 62: In the first line, you assume that `connectionData` is defined, but not in the second one, which is fine?

```
      jwtToken: connectionData.jwtToken,
      //        ^^^^^^^^^^^^^^^ You are assuming that connectionData is truly
      idUser: connectionData ? connectionData.userData.user.idUser : '',
      //      ^^^^^^^^^^^^^^^^^ You are asking if connectionData is truly or falsy
```

src/services/doppler-billing-api-client.ts
```diff
@@ -0,0 +1,106 @@
...
+    };
+  }
+
+  private mapInvoices(data: any): Invoice[] {
+    return data.map((x: any) => ({
+      accountId: x.accountId,
+      product: x.product,
+      date: x.date.toString('dd/mm/yyyy'),
```
andresmoschini, line 70: Please, always use plain dates (or moment objects), but never formatted values, the output of our clients should be flexible data independent of visualization.

src/services/doppler-billing-api-client.ts
jhoffman-ms: The validation of null in the second line is unnecessary because previously we checked it. I will remove it.

src/components/InvoicesList/InvoicesList.js
jhoffman-ms: When the service throws an error, what is the expected behavior?

src/components/InvoicesList/InvoicesList.js
jhoffman-ms: 👍

src/services/doppler-billing-api-client.ts
jhoffman-ms: Yes, good. I will update it.

src/components/InvoicesList/InvoicesList.js
andresmoschini: Maybe you could show a similar message but saying that there was an error? Probably @GusBaamonde has a special component to show errors, but for the moment I think that simply say "There was an error gathering the invoices" should be fine.

jhoffman-ms: Perfect, thanks @andresmoschini

src/App.js
andresmoschini: Well, this URL can be modified in the future, so, it is not so important, but I am starting to think about the UI, there will probably be a section for "billing" and inside that section a subsection for "invoices", for that reason I like a hierarchical URL.

jhoffman-ms: for me is ok if you want that I change the url.

andresmoschini: @lucianaDilizio what do you think?

lucianaDilizio: I don't have a strong opinion about that. I think that `billing/invoices` is ok. 👍

jhoffman-ms:
Update PR with Andres's feedback and also added new UTs when the service throws an error

package.json
```diff
@@ -44,6 +45,7 @@
     "react-helmet": "^6.1.0",
     "react-intl": "^5.8.0",
     "react-markdown": "^4.3.1",
+    "react-moment": "^1.0.0",
```
andresmoschini, line 48: I think that it is not a good idea including _Moment_ if WebApp team is not already using it.

src/components/InvoicesList/InvoicesList.test.js
```diff
@@ -0,0 +1,115 @@
...
+  afterEach(cleanup);
+
+  it('should show invocies', async () => {
+    // Arrange
+    const invoicesCollection = {
+      items: [
+        {
+          accountId: 'CD0000000073689',
+          product: 'Prod 1',
+          date: '09/25/2020',
```
andresmoschini, line 19: It is not a realistic expected result

src/services/doppler-billing-api-client.ts
```diff
@@ -0,0 +1,106 @@
...
+      if (response.status === 200 && response.data) {
+        return {
+          success: true,
+          value: {
+            items: this.mapInvoices(response.data.items),
+            totalItems: response.data.totalItems,
+          },
+        };
+      } else {
+        return { success: false, error: response.data.title };
```
andresmoschini, line 100: If I am not wrong, in this place data can be null (I do not understand how TS does not throw an error), in that case reading `response.data.title` will throw an error, what do you think about returning all the response as the error? Or maybe adding a new Error as a wrapper (but I am not enough familiar with that approach).

```suggestion
        return { success: false, error: response };
```

src/components/InvoicesList/InvoicesList.js
```diff
@@ -0,0 +1,119 @@
...
+              <tbody>
+                {stateInvoices.totalItems > 0 ? (
+                  <>
+                    {stateInvoices.items.map((invoice, index) => {
+                      return (
+                        <tr key={index}>
+                          <td>{invoice.accountId}</td>
+                          <td>{invoice.product}</td>
+                          <td>
+                            <Moment format="DD/MM/YYYY">{invoice.date}</Moment>
```
andresmoschini, line 90: @fcoronelmakingsense If I am not wrong we are formatting dates without using Moment, could you share us an example?

src/services/doppler-billing-api-client.ts
```diff
@@ -0,0 +1,106 @@
...
+      jwtToken: connectionData.jwtToken,
+      idUser: connectionData.userData.user.idUser,
+    };
+  }
+
+  private mapInvoices(data: any): Invoice[] {
+    return data.map((x: any) => ({
+      accountId: x.accountId,
+      product: x.product,
+      date: x.date,
```
andresmoschini, line 70: If I am not wrong, we are using Dates in place of strings. so, here you should convert the _string date_ to a _Date date_.

 @fcoronelmakingsense could you share some examples about how are we doing that today?

src/components/InvoicesList/InvoicesList.test.js
jhoffman-ms: i forgot the replace the date in the test, thanks @andresmoschini . I will replace the format in the test

src/services/doppler-billing-api-client.ts
andresmoschini: Yes, you should return date and not a string (I do not understand why TS is passing):

```ts
export interface Invoice {
  accountId: string;
  product: string;
  date: Date;
  currency: string;
  amount: number;
  filename: string;
}
```

src/components/InvoicesList/InvoicesList.js
jhoffman-ms: I saw in the code and the team is using **FormattedDate**. Is it correct?

jhoffman-ms:
Rollback moment and I used the FormattedDate and it looks fine:

image

src/services/doppler-billing-api-client.ts
jhoffman-ms: Currently the client returns a Date not string for  Date property how you show in the above screenshot

src/components/InvoicesList/InvoicesList.js
jhoffman-ms: I used the FormattedDate and it looks fine:

image

src/services/doppler-billing-api-client.ts
jhoffman-ms: Done

src/components/InvoicesList/InvoicesList.test.js
jhoffman-ms: Done

package.json
jhoffman-ms: I did rollback of  moment

src/components/InvoicesList/InvoicesList.js
GusBaamonde: We already reviewed the case with Juan, he will apply a simple solution. The empty table is shown saying that "there is no data available" since the design does not contemplate this case. It can be reviewed with design later.

jhoffman-ms:
Updated PR with the feedback

src/services/doppler-billing-api-client.ts
andresmoschini: We talked with @jhoffman-ms about this and he have applied the related changes. Resolving...

src/components/InvoicesList/InvoicesList.test.js
andresmoschini, line 19:
```suggestion
          date: new Data('2020-09-25'),
```

jhoffman-ms: Done

PR #1088, opened 2020-09-28
fix(ChangePlan): add animation in compare features plan.

fcoronelmakingsense:
- Add toggle animation when compare features between plans. like in the UI library. https://cdn.fromdoppler.com/doppler-ui-library/latest/doppler-plus.html

image


andresmoschini:
Nice!

PR #1093, opened 2020-10-02
DAT-140: feat(invoices): add download invoice

jhoffman-ms:
Added the invoice  download in the Invoices list page.

**Changes:**

-  Added message when the invoice hasn't url to download.

image

- Added a link to download the invoice:

image

- Added new unit tests for it functionality

**Task:** [DAT-140](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-140)


src/services/doppler-billing-api-client.ts
```diff
@@ -71,6 +72,7 @@ export class HttpDopplerBillingApiClient implements DopplerBillingApiClient {
       currency: x.currency,
       amount: x.amount,
       filename: x.filename,
+      downloadInvoiceUrl: x.downloadInvoiceUrl,
```
andresmoschini, line 75: I have no preference for your internal name, but in the billing API I would prefer something more concise. What do you think about `downloadUrl`, `fileUrl` or `file`?

Another option is going back to our initial approach and use a dynamic `_links` field with a descriptive `rel` property: https://github.com/FromDoppler/doppler-billing-api/commit/4411b0a62f460b0c2cbfedc9afd446f283ec560f#diff-04c6e90faac2675aa89e2176d2eec7d8L26-R26

jhoffman-ms: I like the idea to use a dynamic links property instead of a new property for the download invoice url. At the moment the property can have:

links: [{
              "rel": "downloadUrl",
               "path": "/accounts/doppler/22235/invoices?from=2020-02-01T00:00:00Z&to=2020-07-01T00:00:00Z"
           }]

Do you agree?

andresmoschini: Yes, but I think that we should use absolute URLs, so:

```json
{
  // . . .
  "_links": [
    {
      "rel": "downloadUrl",
      "path": "https://apis.fromdoppler.com/billing-api/accounts/doppler/22235/invoices/invoice_2020-10-05_10.pdf?0bd058e8fa5788b889dedf74ab900d5e51532d81"
    }
  ]
}
```

andresmoschini: Notice that I have replaced `links` by `_links` (to denote metadata), `path` by `href`, and the `rel` value by `file`; also added a `description` field.

jhoffman-ms: yes, perfect. We will implement it.

jhoffman-ms: Done. I updated the code to implment it also I added the unit tests for the different scenarios.

src/services/doppler-billing-api-client.ts
```diff
@@ -63,6 +64,15 @@ export class HttpDopplerBillingApiClient implements DopplerBillingApiClient {
     };
   }
 
+  private getDownloadUrl(links: any[]): string {
+    if (links) {
+      const link = links.find((l) => l.rel === 'file');
```
andresmoschini, line 69: Cuidado, el rel es un string que puede tener muchos términos, creo que en lugar de `===` deberías usar una expresión regular que acepte la palabra completa `file` incluso si hubiera más palabras, creo que deberías hacer tests para verificar estos casos: 
* `file otras palabras` → _Es un archivo_
* `otras palabras file` → _Es un archivo_
* `otras file palabras` → _Es un archivo_
* `fileotro` → _No es un archivo_
* `otrofile` → _No es un archivo_
* `file`

jhoffman-ms: perfect, I will work on it.

jhoffman-ms: Done.

**Changes:**

- Added the function: searchLinkByRegularExpressionRel in utils.ts. The function receives two parameters: the links and the regular expresion to search.
- Added unit tests for the scenarios above.

andresmoschini:
Good job! thanks!

src/i18n/en.js
```diff
@@ -153,6 +153,7 @@ You'll find an Email with steps to follow.`,
     download_msg: `Download`,
     error_msg: `There was an error gathering the invoices.`,
     no_data_msg: `There are no invoices to show`,
+    no_download_msg: `There is not invoice to download`,
```
cbernat, line 156:
```suggestion
    no_download_msg: `There is no invoice to download`,
```

jhoffman-ms: Thanks Ceci, I will update the message in the next PR.

PR #1109, opened 2020-10-14
fix(security): prevent clickjacking header

cbernat:
### Add X-frame-options header
To prevent that webapp could be rendered inside an iframe, added x-frame-options header in deny option. 

image

image


cbernat:
@andresmoschini fyi

andresmoschini:
👍

PR #1110, opened 2020-10-14
Dw 524 apply ui in plan calculator

lucianaDilizio:
Apply UI in plan calculator page. 

Test Url: https://cdn.fromdoppler.com/doppler-webapp/demo-build2770/#/plan-selection

image


GusBaamonde:
the green message of: do you have more than 100,000 contacts? appears only when you reach the top of the bar, before it is not displayed.

cbernat:
@GusBaamonde how about this? The value falls O_o
image

cbernat:
> the green message of: do you have more than 100,000 contacts? appears only when you reach the top of the bar, before it is not displayed.

I think here is the same for subscribers plan right? if you've reached: the peak, get a custom plan.

GusBaamonde:
> @GusBaamonde ¿qué tal esto? El valor cae O_o
If we already saw it, I adjust it for two words, but it should only be one

src/components/ChangePlan/PlanCalculator.js
```diff
@@ -29,24 +29,42 @@ const NavigatorTabs = ({ tabs, pathType, selectedPlanType }) => {
...
-            setSelectedDiscount(discount);
-          }}
-        >
-          {discount.description}
-        </button>
-      ))}
-    </div>
+    <>
+      <div className="dp-wrap-subscription">
+        <h4>Suscripcion</h4>
```
cbernat, line 34: Please add the language variable, so when we have the real text is easier to change

cbernat: The same for all loose texts...

src/components/ChangePlan/PlanCalculator.js
```diff
@@ -125,15 +143,28 @@ const PlanCalculator = ({
...
+              amount:
+                plan.type === 'prepaid'
+                  ? plan.credits
+                  : plan.type === 'subscribers'
+                  ? plan.subscriberLimit
+                  : plan.type === 'monthly-deliveries'
+                  ? plan.emailsByMonth
+                  : null,
+              description: _(
+                'plan_calculator.' + plan.type.replace('-', '_') + '_amount_description',
```
cbernat, line 164: I'm not so sure about this, think that if someone were to add a new plan and you do not have the resource in the language file, you would see gibberish.

src/components/ChangePlan/PlanCalculator.js
lucianaDilizio: What do you think about add this changes in other PR. There are some texts that aren't so easy to apply because, for example, can change based on what type of plan do you select or depends on variable data.

src/components/ChangePlan/PlanCalculator.js
lucianaDilizio: I don't understand how can we fix this. In that case I suppose that we have to code the FE before, to anticipate us at the new changes.

lucianaDilizio: Or we can get the description with their translation from the BE, but I think that it wouldn't be a good idea.

src/components/ChangePlan/PlanCalculator.js
```diff
@@ -186,74 +215,120 @@ const PlanCalculator = ({
...
+                                  idDiscount: discount.id,
+                                });
+                              }}
+                            />
+                          ) : (
+                            <></>
+                          )}
+                        </article>
+                      </div>
+                      <div className="col-md-6 col-sm-12">
```
cbernat, line 251: Perhaps all price handling could become something like a `price` Component or `priceWithDiscount` component, to give more meaning to this section. You could have something like this:
`
<SelectedPlanPrice>
   <RegularPlanPrice/>
   <PlanPriceWithPromotions/>
</SelectedPlanPrice>
`
Does it make sense?

src/components/ChangePlan/PlanCalculator.js
```diff
@@ -186,74 +215,120 @@ const PlanCalculator = ({
...
+                    to={`/plan-selection${safePromoId ? `?promo-code=${safePromoId}` : ''}`}
+                  >
+                    Volver a Planes
+                  </Link>
+
+                  <a
+                    className="dp-button button-medium primary-green"
+                    href={
+                      _('common.control_panel_section_url') +
+                      `/AccountPreferences/UpgradeAccountStep2?IdUserTypePlan=${planData.plan.id}&fromStep1=True` +
```
cbernat, line 318: Perhaps this link could be generated elsewhere like in the plan service for example. What do you think?

src/i18n/en.js
```diff
@@ -206,6 +206,11 @@ You'll find an Email with steps to follow.`,
     grid_delivery: `Delivery Status`,
     grid_subject: `Subject`,
   },
+  plan_calculator: {
+    monthly_deliveries_amount_description: 'Emails',
+    prepaid_amount_description: 'Credits',
+    subscribers_amount_description: 'Suscribers',
```
cbernat, line 212:
```suggestion
    subscribers_amount_description: 'Subscribers',
```

src/components/ChangePlan/PlanCalculator.js
lucianaDilizio: Yes, I can try to improve it, but I don't know if this is something to resolve in this PR. Maybe create a ticket to isolate price section in a new element, what do you think?

src/components/ChangePlan/PlanCalculator.js
cbernat: A function would do fine.

src/components/ChangePlan/PlanCalculator.js
cbernat: You owe this change since some PRs... There's no hurry here, you can make this texts changes in another commit perhaps.

src/components/ChangePlan/PlanCalculator.js
lucianaDilizio: This will be resolve in [DW-529](https://makingsense.atlassian.net/jira/software/projects/DW/boards/16?selectedIssue=DW-529)

lucianaDilizio:
> the green message of: do you have more than 100,000 contacts? appears only when you reach the top of the bar, before it is not displayed.

That will be resolve in [DW-530](https://makingsense.atlassian.net/browse/DW-530)
For the moment it will be hidden.

src/components/ChangePlan/PlanCalculator.js
cbernat:
```
...
descriptionId: planItemDescription(plan.type)
...
const planItemDescription = (planType) => {
   switch (planType): {
      case 'subscribers': 
         return ('plan_calculator.subscribers_amount_description');
      ...
      default: 
         return ('plan_calculator.unknown_amount_description'); // --> this could be empty or default to something else
   }
}
```

src/components/ChangePlan/PlanCalculator.js
cbernat: Like this 
```
planService.getBuyLink(planData.plan.id, planData.discount?.id, safePromoId);
```

cbernat:
Perhaps you should collapse commits like the one of the warnings, and also, it would be nice to improve commit readability to have  the section associated in the change asserted like this: chore(sliderComponent): add translations.

PR #1114, opened 2020-10-19
feat(invoices): show paid and balance columns into the invoices list

jhoffman-ms:
Added Paid and Balance columns  into the invoices list:

image

**Changes:**

- Added Paid and Balance columns  into the invoices list

- Added new unit tests

**Task:** [DAT-157](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-157)


src/services/doppler-billing-api-client.ts
```diff
@@ -83,6 +96,8 @@ export class HttpDopplerBillingApiClient implements DopplerBillingApiClient {
       dueDate: !!x.dueDate ? new Date(x.dueDate) : null,
       currency: x.currency,
       amount: x.amount,
+      paidToDate: !!x.paidToDate ? x.paidToDate : 0,
+      balance: this.getBalance(x),
```
andresmoschini, line 100: Is it really necessary? I mean, if you do not have the `balance` field you neither have the `paidToDate` field.

So, in my opinion, it should be simply: 

```suggestion
      balance: data.balance == undefined ? x.amount : x.balance
```

or 

```suggestion
      balance: data.balance == undefined ? null : x.balance
```

andresmoschini: I mean, I think that the backend is responsible for the balance value.

jhoffman-ms: Perfect, I like it:

```balance: data.balance == undefined ? x.amount : x.balance```

I will replace for it.

andresmoschini: ok, please test it, I did not.

jhoffman-ms: yes, totally agree

jhoffman-ms: @andresmoschini, I will use: ```balance: !!x.balance ? x.balance :  x.amount,``` to be consistent with the other statements. Do you agree?

jhoffman-ms: Done

cbernat:
image
What about this?

jhoffman-ms:
Thanks @cbernat , I will review it.

jhoffman-ms:
@cbernat, the problem is because the header for it "download" column is empty. Maybe we need set a header for example : "Actions". @GusBaamonde/ @cbernat  do you have any suggestions?.

GusBaamonde:
We should not have empty headers, if the action is download, that should be our **"Downloads"** header. If later we have more shares we can change the name, what do you think?

jhoffman-ms:
Thanks @GusBaamonde. For me is Ok. I will add it in the header.

jhoffman-ms:
Done.

image

jhoffman-ms:
Do you have another feedback or suggestion or is ready to merge?

PR #1115, opened 2020-10-19
fix(ChangePlan): do not show "since" always when current plan

lucianaDilizio:
If the card is showing current plan, it has not to show the legend "since".

image


cbernat:
👍

PR #1124, opened 2020-10-20
fix(PlanCalculator): select existed plan or first one

lucianaDilizio:
Fix issue when the user reach the top of subscribers plans. To avoid problems, the Plan Calculator screen will show the plan type passed by param or the first one in the list of available plan types if the one passed by param does not exist.


cbernat:
👍

PR #1127, opened 2020-10-21
feat(PlanCalculator): show banner when reach max of slider

lucianaDilizio:
Show banner when the user reach the max of a plan.

For Prepaid  --> Goes to Sending tab
For Contacts --> Goes to Sending tab
For Sending --> Goes to Agencies form (it should go to a similar page that we don't have yet)

image


cbernat:
@lucianaDilizio in prepaid plan there should not be an option for exclusive plan, all exclusive plans are monthly as you know.

image

lucianaDilizio:
> @lucianaDilizio in prepaid plan there should not be an option for exclusive plan, all exclusive plans are monthly as you know.
> 
> image

Yes, it's true. Maybe we should show exclusive plans banner only to hight volume plans where you don't have more to grow up. Even it could be only in plus. I have to find out this.

PR #1142, opened 2020-10-27
feat(Pagination): improve pagination element

lucianaDilizio:
### Improve pagination element
Avoid problems when there are too many pages.

- Show only 5 pages to select.
- Show first and last page when you can't see them in the page list.
- Add (...) to slide to left or right between pages.
- Add < and > to go to next and previous page.

image


andresmoschini:
Could you add tests?

src/components/shared/Pagination/Pagination.test.js
```diff
@@ -0,0 +1,142 @@
...
+  it('should show previous page element when current page is not the first one', () => {
+    //Act
+    const { container } = render(
+      <BrowserRouter>
+        <Pagination currentPage={2} pagesCount={100} urlToGo={'https://domain.com/section?'} />
+      </BrowserRouter>,
+    );
+
+    // Asserts
+    expect(container.querySelector('.icon-arrow-prev')).toBeInTheDocument();
```
cbernat, line 19: I think here you do both validations for prev and next right?
```suggestion
    expect(container.querySelector('.icon-arrow-prev')).toBeInTheDocument();
    expect(container.querySelector('.icon-arrow-next')).toBeInTheDocument();
```

lucianaDilizio: Mmmmm, no. I have a validation for next element in another test.

src/components/shared/Pagination/Pagination.test.js
```diff
@@ -0,0 +1,142 @@
...
+      <BrowserRouter>
+        <Pagination currentPage={100} pagesCount={100} urlToGo={'https://domain.com/section?'} />
+      </BrowserRouter>,
+    );
+
+    // Asserts
+    expect(container.querySelector('.icon-arrow-next')).not.toBeInTheDocument();
+  });
+
+  it('should show ... to slide to left when selected page is not the first one', () => {
```
cbernat, line 58: Here shouldn´t it be shown the points when you are farther away from the begining? like at least 5 pages away?

src/components/shared/Pagination/Pagination.test.js
```diff
@@ -0,0 +1,142 @@
...
+      <BrowserRouter>
+        <Pagination currentPage={1} pagesCount={0} urlToGo={'https://domain.com/section?'} />
+      </BrowserRouter>,
+    );
+
+    // Asserts
+    expect(container.querySelector('.dp-pagination')).not.toBeInTheDocument();
+  });
+
+  it('should show pagination if there is at least one page', () => {
```
cbernat, line 131: I'm not so sure about this, why would we want a pagination if you can't paginate over another page? 🤔

src/components/shared/Pagination/Pagination.test.js
lucianaDilizio: It could be an improvement.
If you are in page 3 you will see:
< 1 ... 3 4 5 6 7 ... 100>

if you press the first ... you will see:
< 1 ... 2 3 4 5 6 ... 100 >

if you press it again you will see:
< 1 2 3 4 5 ... 100 >

The idea of the ... is allow you to move between pages without going to any one until you select one.

cbernat: Thanks for the explanation.

src/components/shared/Pagination/Pagination.test.js
cbernat: Yeah I know, what I'm saying is perhaps it can be done all in the same test right?

src/components/shared/Pagination/Pagination.test.js
cbernat: < 1 > 🤔

lucianaDilizio: Yes, is a little weird but another test exists, for invoices list page, that specify that the pagination element should be shown when there are information to show. I didn't want to have problem with that specification.

cbernat: Perhaps the implementation is not so good. This is the great thing about tests, they let us think more about possible cases...

src/components/shared/Pagination/Pagination.test.js
lucianaDilizio: Yes, but I preferred to keep them separated because the conditions to show each ... is different in each case.

src/components/shared/Pagination/Pagination.test.js
lucianaDilizio: Yes, I am agree with you. I can change the rules for this other test.

cbernat: Perhaps we should ask for ux opinion @santiagopaolucci what do you think?

lucianaDilizio: As we talk in slack with @santiagopaolucci and @mdirago we will show pagination element even if there is only one page.

PR #1147, opened 2020-10-30
Dw 536 plan calculator tests

lucianaDilizio:
### TESTS for PlanCalculator 
- Remove unnecessary dependencies. 
- Fix tests errors in Slider tests for unknown intl ids.

**Tests:**
- Check that tabs are showed
- Check that a message is showed when user reach the max plan in the slider
- Check that the message before is not showed in the oposite case
- Check that the discount element is showed if the first plan of the type has discounts to apply
- Check that the discount element is not showed if the first plan of the type doesn't have discounts to apply
- Check that a message is showed when the discount is applied
- Check that a is no showing any message if the price doesn't have applied any discount
- Check that plan price is showed with a discount if applies 
- Check that the slider element is showed
- Check that an error is showed when there is no plans to select

### DIFF
--- a/src/components/ChangePlan/PlanCalculator.test.js
+++ b/src/components/ChangePlan/PlanCalculator.test.js
@@ -1,0 +1,348 @@
+import React from 'react';
+import { render, cleanup, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import PlanCalculator from './PlanCalculator';
+import IntlProvider from '../../i18n/DopplerIntlProvider.double-with-ids-as-values';
+import { AppServicesProvider } from '../../services/pure-di';
+import { MemoryRouter, Route } from 'react-router-dom';
+
+describe('PlanCalculator component', () => {
+  afterEach(cleanup);
+
+  const dependencies = {
+    appSessionRef: {
+      current: {
+        userData: {
+          user: {
+            plan: {},
+          },
+        },
+      },
+    },
+  };
+
+  const planServiceDoubleBase = {
+    getPlanList: () => [],
+    mapCurrentPlanFromTypeOrId: () => {},
+    getPlanTypes: () => ['prepaid', 'subscribers', 'monthly-deliveries'],
+    getBuyUrl: () => '',
+  };
+
+  const PlanCalculatorElement = ({ planServiceDouble }) => (
+    <MemoryRouter initialEntries={['plan-selection/standard']}>
+      <Route path="plan-selection/:pathType/:planType?">
+        <AppServicesProvider forcedServices={{ ...dependencies, planService: planServiceDouble }}>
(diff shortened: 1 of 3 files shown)


src/components/ChangePlan/PlanCalculator.test.js
```diff
@@ -0,0 +1,374 @@
...
+        userData: {
+          user: {
+            plan: {},
+          },
+        },
+      },
+    },
+  };
+
+  it('should show all plan types in tabs', async () => {
```
cbernat, line 24: How about validate for the slider to have some elements? Or the slider to be present here.

cbernat:
Very nice set of tests!

src/components/ChangePlan/PlanCalculator.test.js
```diff
@@ -0,0 +1,374 @@
...
+          price: 15,
+          featureSet: 'standard',
+        },
+      ],
+    };
+
+    // Act
+    const { container, getByText } = render(
+      <AppServicesProvider forcedServices={{ ...dependencies, planService: planServiceDouble }}>
+        <IntlProvider onError={() => null}>
```
cbernat, line 46:
```suggestion
        <DopplerIntlProvider>
```
As we speaked better using this provider as in reports https://github.com/FromDoppler/doppler-webapp/blob/master/src/components/Reports/Reports.test.js#L18

src/components/ChangePlan/PlanCalculator.test.js
```diff
@@ -0,0 +1,374 @@
...
+    await waitFor(() => {
+      expect(queryByText('plan_calculator.with_quarterly_discount')).toBe(null);
+      expect(queryByText('plan_calculator.with_half_yearly_discount')).toBe(null);
+      expect(queryByText('plan_calculator.with_yearly_discount')).toBe(null);
+    });
+  });
+
+  it('should calc price with discount', async () => {
+    // Arrange
+    const planServiceDouble = {
```
cbernat, line 330: Perhaps instead of redefining this in each test, you could have a function that returns a service with, all plan types available or only prepaid, and so on. To improve readability. What do you think?

lucianaDilizio: Yes, or maybe I can define only getPlans method in each test and have just once the other ones (because are always the same)... I mean.. to be clear what is the context in each test.

lucianaDilizio:
```
    const planServiceDouble = {
      ...planServiceDoubleBase,
      getPlans: () => [
        {
          type: 'prepaid',
          id: 1,
          name: '1500-CREDITS',
          credits: 1500,
          price: 15,
          featureSet: 'standard',
        },
      ],
    };
```

PR #1153, opened 2020-11-02
feat: allow deploy from master and main

cbernat:
Only two points to be modified, I will leave this pr in draft until needed. 

https://github.com/github/renaming

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -96,7 +96,10 @@ pipeline {
         }
         stage('Generate version') {
             when {
-                branch 'master'
+                anyOf {
+                    branch 'master';
+                    branch 'main'
+                }
             }
             steps {
                 // TODO: by the moment this work is being done by another Jenkins task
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -96,7 +96,7 @@ pipeline {
         }
         stage('Generate version') {
             when {
-                branch 'master'
+                branch 'main'
```
andresmoschini, line 99: What do you think about keeping both temporarily? or it is completely unnecessary?

cbernat: Is it possible? O_o If so I will, I was planning on leave this PR opened until needed.

andresmoschini: Yep, you can generate a version when the branch is `main` or when the branch is `master`, I think that the command is _oneOf_ or something like that.

cbernat:
@andresmoschini I´ve added optional configuration for master or main. I think I will merge this as it is, I've tried semantic release on dry-run and it seems to be working ok. What do you think?

PR #1165, opened 2020-11-09
feat(AgenciesForm): apply content

lucianaDilizio:
### Apply content in Agencies Form
- Apply content
- Apply submit button behavior when submit form
- Change url path
- Add agencies url in plan selection screen
- Tests


src/components/DopplerPlus/AgenciesForm.js
```diff
@@ -100,29 +105,29 @@ const AgenciesForm = ({ dependencies: { dopplerLegacyClient, appSessionRef } })
...
-                <legend>Completa el siguiente formulario de planes</legend>
+                <legend>{_('agencies.form_legend')}</legend>
                 <fieldset>
                   <FieldGroup>
                     <EmailFieldItem
                       autoFocus
                       type="email"
                       fieldName={fieldNames.email}
-                      label="Email:"
+                      label={'*' + _('forms.label_email')}
```
cbernat, line 130: How about using string literals here? to avoid the +
```suggestion
                      label={`*${ _('forms.label_email')}`}
```

src/components/form-helpers/form-helpers.js
```diff
@@ -100,20 +100,24 @@ export const FormWithCaptcha = ({
...
     // See more details in https://stackoverflow.com/questions/43488605/detect-when-challenge-window-is-closed-for-google-recaptcha
-    formikProps.setSubmitting(false);
-    const result = await verifyCaptcha();
-    formikProps.setSubmitting(true);
-    if (result.success) {
-      await originalOnSubmit(
-        { ...values, captchaResponseToken: result.captchaResponseToken },
-        formikProps,
-      );
+    if (process.env.NODE_ENV === 'test') {
```
cbernat, line 103: Nice approach to fix this issue for testing.

cbernat: You could try this way that you are doing: use the captcha form as it were a normal form for testing, or test that the captcha is fired if you use a captcha form. Perhaps a middle point approach could be, to have a flag to disable the captcha by code (instead of validating with test env) to allow both scenarios. Perhaps something like this:
```suggestion
    if (process.env.NODE_ENV === 'test' && process.env.REACT_APP_DISABLE_CAPTCHA) {
```

lucianaDilizio: I was reviewing my changes and it seems to be that this changes are not necessaries. 😄

lucianaDilizio: I have deleted the last comment. If you read it, ignore it.

andresmoschini: No leí todo, pero no está bueno leer variables de entorno en los js, tampoco verificar si el entorno es de test.

Entiendo que puede ser difícil, que tal vez no sea el momento de hacerlo y que entonces sea razonable dejar este if, pero es una ventana rota, por lo menos deberíamos dejar un comentario con un TODO.

Lo ideal sería por injección de dependencias inyectar un dummy del captcha.

cbernat: Thanks @andresmoschini  for the suggestion, are you up to this change @lucianaDilizio ?

lucianaDilizio: >  cbernat 20 minutes ago Member
> Thanks @andresmoschini for the suggestion, are you up to this change @lucianaDilizio ?

No 😅. I will try to do it here, but maybe it would be better to do this in another PR.

cbernat: If you are willing to do this in another PR perhaps all tests should be there as well.

andresmoschini: I think that it is fine doing it here, the tests are addinmg safety. 

What do you think about simply adding a TODO comment by the moment?

PR #1168, opened 2020-11-12
fix: prepare reCaptcha for tests

lucianaDilizio:
Modify captcha utils to be used as a service that must be injected. This allows to harcode the service in tests. 

```
..............
..............
describe('AgenciesForm component', () => {
  afterEach(cleanup);

  const dependencies = {
    captchaUtilsService: {
      useCaptcha: () => {
        const Captcha = () => null;
        const verifyCaptcha = async () => {
          return { success: true, captchaResponseToken: 'hardcodedResponseToken' };
        };
        const recaptchaRef = null;
        return [Captcha, verifyCaptcha, recaptchaRef];
      },
    },
  };
..............
..............
```

### DIFF
--- a/src/components/DopplerPlus/AgenciesForm.test.js
+++ b/src/components/DopplerPlus/AgenciesForm.test.js
@@ -24,6 +24,16 @@ describe('AgenciesForm component', () => {
         return { success: true };
       },
     },
+    captchaUtilsService: {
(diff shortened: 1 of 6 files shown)


andresmoschini:
Great! ❤️

cbernat:
Nice! ❤️

PR #1177, opened 2020-11-18
chore(Signup): add test

lucianaDilizio:
### Signup confirm page test
- It add a test to check that confirm page is showed when user submits the form.


cbernat:
Nice! ❤️

PR #1187, opened 2020-11-24
feat: add maintenance state for deploy using new endpoint

cbernat:
### Maintenance state for deploy
Added new endpoint in Doppler [here](https://github.com/MakingSense/Doppler/pull/7502).
Used new endpoint to put all app in maintenance mode, instead of getUserData (this makes use of error by CORS to see if user is not logged in, we tried to use this call, with this issue: app offline, all calls return error by CORS)
By default disable, and enabled only in int for now. 

image

You can see this in action in int https://cdn.fromdoppler.com/doppler-webapp/int-build2877/#/login by doing like in deploy, set the app offline, and then you will see offline page, in qa it is not enabled https://cdn.fromdoppler.com/doppler-webapp/qa-build2875/#/login


src/services/session-manager.ts
```diff
@@ -60,6 +60,14 @@ export class OnlineSessionManager implements SessionManager {
...
 
   private async update() {
+    if (process.env.REACT_APP_ALLOW_MANTENAINCE_AUTODETECT === 'true') {
+      try {
+        await this.dopplerLegacyClient.isDopplerMVCUp();
+      } catch {
+        this.updateSession({ status: 'maintenance' });
+        return;
+      }
+    }
```
lucianaDilizio, line 70: I don't understand what is the difference doing what the code below does and this. 🤔 
Isn't the same doing this with getUsetData?

cbernat: We tried this approach as you can see in the code, the issue is: when we call getUserData, we use CORS error to identify if user is not logged in, and when we tried reusing this call, we noticed that all calls to doppler from webapp returned CORS error when the server is offline. Explanation in spanish [here](https://github.com/MakingSense/Doppler/pull/7502)

cbernat: image

lucianaDilizio: Ok, maybe the other approach it could be useful when the user was already logged in the app but not for users that wants to entry for the first time. So it seems to be that this should work 👍

lucianaDilizio (commented):
👍

cbernat:
@andresmoschini the other part for maintenance screen. What do you think?

src/services/session-manager.ts
```diff
@@ -79,11 +79,7 @@ export class OnlineSessionManager implements SessionManager {
         } as AppSession, // Cast required because TS cannot resolve datahubCustomerId complexity
       );
     } catch (error) {
-      if (!!error.response && (error.response.status === 503 || error.response.status === 404)) {
-        this.updateSession({ status: 'maintenance' });
-      } else {
-        this.updateSession({ status: 'non-authenticated' });
-      }
+      this.updateSession({ status: 'non-authenticated' });
```
lucianaDilizio, line 82: 👍

src/services/session-manager.ts
cbernat: For this to work we would need to change the actual behaviour for login, it's a bit risky.

src/services/session-manager.ts
```diff
@@ -60,6 +60,14 @@ export class OnlineSessionManager implements SessionManager {
   }
 
   private async update() {
+    if (process.env.REACT_APP_ALLOW_MANTENAINCE_AUTODETECT === 'true') {
```
andresmoschini, line 63: Esto debería entrar por una configuración, no? (para no leer derecho el process digo)

src/services/session-manager.ts
andresmoschini: No convendría llamar a `isDopplerMVCUp` aquí? de esa forma, el caso feliz sería más rápido.

src/services/session-manager.ts
cbernat: Inyectarlo decís no?

src/services/session-manager.ts
cbernat: Genial me encantó 😁

src/services/session-manager.ts
andresmoschini: Si, inyectar una configuración o un servicio para leer _process_ en la menor cantidad de lugares, en lo posible solo se debería leerlo en el composition root.

src/services/pure-di.tsx
```diff
@@ -176,6 +176,7 @@ export class AppCompositionRoot implements AppServices {
           appSessionRef: this.appSessionRef as MutableRefObject<AppSession>,
           dopplerLegacyClient: this.dopplerLegacyClient,
           keepAliveMilliseconds: this.appConfiguration.dopplerLegacyKeepAliveMilliseconds,
+          maintenanceAutoDetect: process.env.REACT_APP_ALLOW_MANTENAINCE_AUTODETECT === 'true',
```
andresmoschini, line 179: I am not totally sure, but it seems that the old approach is using `this.appConfiguration` in place of directly reading `process`. 

I think that it could be useful if we want to override the complete configuration. I mean, having a single point for reading the configuration.

image


Another advantage is that you unify the parsing and typing in a single place:

```suggestion
          maintenanceAutoDetect: this.appConfiguration.allowMantenainceAutodetect,
```

```javascript
  get appConfiguration() {
    return this.singleton('appConfiguration', () => ({
      allowMantenainceAutodetect: process.env.REACT_APP_ALLOW_MANTENAINCE_AUTODETECT === 'true',
      dopplerLegacyUrl: process.env.REACT_APP_DOPPLER_LEGACY_URL as string,
      dopplerSitesUrl: process.env.REACT_APP_DOPPLER_SITES_URL as string,
```

cbernat: We have all kinds of use of `process.env.....` everywhere, I will fix this here and be on the lookout for other mishaps.

andresmoschini: > We have all kinds of use of `process.env...`

We should not. It should be reduced to the minimum.

andresmoschini:
🚀

cbernat:
@ffernandez4 @swidzinskiMS This maintenance page will be disabled until back end is deployed. When production release is being made we will need to coordinate to enable this in production.

fgchaio:
👏

PR #1214, opened 2020-12-16
fix(deploy-page): change message

cbernat:
Changed message to avoid time and inform about campaign sending.

image


cbernat:
@mrunge86 https://cdn.fromdoppler.com/doppler-webapp/build2909/#/offline

PR #1215, opened 2020-12-17
DW-617: fix: set separate unexpected error for signup only

cbernat:
### Add different unexpected error for signup

- Add contact to support link only for signup unexpected error.

image


cbernat:
@mrunge86 😄

cbernat:
@andresmoschini Previously we had this message: ¡Ouch! Detectamos un problema de conexión. Por favor inténtalo nuevamente en unos minutos.` because we agreed on removing the option to contact support in login. The issue was, that the error was shared with register also.

PR #1216, opened 2020-12-18
Improve loggly log

cbernat:
### Improve loggly log

- On retry log only last attempt
- Remove login post logged into login component (since it will log then again on retry)


andresmoschini:
👍

PR #1224, opened 2020-12-21
fix(retry): DW-616 space a bit more between retries

cbernat:
### Increate time between retries for login and signup
After post to login or signup we have a retry policy, but all attempts were made within the same second, spacing seems to be a good idea.

Before we had 200 ms fixed in retry, so for each retry:
1st retry: wait 200ms
2nd retry: wait 400ms
3rd retry: wait 600ms 

With this change:
1st: 700ms
2nd: 1.4s
3rd: 2.1s

@andresmoschini how do you feel this?


andresmoschini:
I have no idea, but I think that it is ok to try. You can check the logs to confirm if it was a good idea or not.

PR #1227, opened 2020-12-29
DM-802 - New docker registry

ms-darianbenito:
We will publish images to both registries fromdoppler (current) and dopplerdock (new).

On the first commit we are checking if we can remove some steps in the current proccess

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,7 +48,7 @@ pipeline {
                 // TODO: remove build differences based on environments to allow reducing the
                 // number of different images. Maybe `demo` and `production` cold be enought
                 sh '''docker build \
-                    -t "fromdoppler/doppler-webapp:production-commit-${GIT_COMMIT}" \\
+                    -t "local-doppler-webapp:production-commit-${GIT_COMMIT}" \\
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -39,59 +39,22 @@ pipeline {
...
-                    --build-arg version=int-${TAG_NAME}+${GIT_COMMIT} \\
-                    -f Dockerfile.swarm \\
-                    .'''
-                 sh '''docker build \
-                    -t "fromdoppler/doppler-webapp:qa-commit-${GIT_COMMIT}" \\
-                    --build-arg environment=qa \\
-                    --build-arg react_app_router=browser \\
-                    --build-arg public_url="/" \\
-                    --build-arg version=qa-${TAG_NAME}+${GIT_COMMIT} \\
+                    --build-arg version=int-v0.0.0+${GIT_COMMIT} \\
```
elsupergomez, line 50: What happend with _production_ and _demo_?

ms-darianbenito: It was a temporary change. Now, you can see all enviroments.

andresmoschini:
👍

PR #1257, opened 2021-01-25
DW-624 recaptcha fix

cbernat:
### Recaptcha execute timeout error

When submit login form, and failed for connection reasons or recaptcha was slow in responding, you could trigger recaptcha.execute serveral times, producing a timeout error. 
Adding a timeout seems to improve this behaviour, giving it more time to recaptcha to get a response from google's servers.

I will try this approach and then watch closely loggly log to make sure this is working.


andresmoschini:
👍

PR #1262, opened 2021-02-01
fix(offline-page): add timeout 30 sec to resolve no server response…

cbernat:
### Offline page: Add timeout 1 minute to resolve no server response in deploy
Analizing the logs from production during deploy: 
When we try to establish if the user is connected or not in `session-manager`, and mvc is offline, the server does not respond to `getUserData` request. I suspect that the code freezes in await, that's why we do not see the offline page. 

image

### Fix
To test this I've added a timeout of 30 seconds to `getUserData` call only, with a log entry for this case, to monitorize (just in case).

Production log shows this:
image

image

### DIFF
--- a/src/services/doppler-legacy-client.ts
+++ b/src/services/doppler-legacy-client.ts
@@ -566,7 +566,7 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
   }
 
   public async getUserData() {
-    const response = await this.axios.get('/WebApp/GetUserData');
+    const response = await this.axios.get('/WebApp/GetUserData', { timeout: 30000 });
     if (!response || !response.data) {
       throw new Error('Empty Doppler response');
     }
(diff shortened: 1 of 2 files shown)


src/services/doppler-legacy-client.ts
```diff
@@ -566,7 +566,7 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
   }
 
   public async getUserData() {
-    const response = await this.axios.get('/WebApp/GetUserData');
+    const response = await this.axios.get('/WebApp/GetUserData', { timeout: 60000 });
```
andresmoschini, line 569: Is not it too much to wait for a minute?

cbernat: This timeout works only in the worst case, think of  a deploy in which the server does not return any response ever. Or if some user has very slow connection we should allow it right? Chrome has 5 minutes to timeout a connection.  What do you suggest?

cbernat: image

lucianaDilizio: Does anybody will wait one minute till receive a response? 🤔 
I think that Andres is right, maybe the timeout could be... 5 seconds?

cbernat: In the worst case scenario only, that is no response from server or excesive slow connection. In normal everyday this number does not affect. I've put this timeout in 30 seconds.

PR #1264, opened 2021-02-03
feat(ExclusiveForm): add send volume and message.

fcoronelmakingsense:
image

With other options selected

image


src/components/DopplerPlus/ExclusiveForm.js
```diff
@@ -184,6 +219,32 @@ const ExclusiveForm = ({ dependencies: { dopplerLegacyClient, appSessionRef } })
                       </fieldset>
                     </FieldItem>
                   </FieldGroup>
+                  <div className="dp-wrapper-volume-options" style={{ margin: '30px 0' }}>
```
cbernat, line 222: Why not create a styled component for this `div`? To avoid using inline style, because this feature will be here forever. 😄

cbernat:
Could you add some test? Perhaps about validation for radio buttons, and features almost one selected.

src/components/DopplerPlus/ExclusiveForm.js
fcoronelmakingsense: because its only for this element, but i can create

src/components/DopplerPlus/ExclusiveForm.js
```diff
@@ -97,6 +128,10 @@ const ExclusiveForm = ({ dependencies: { dopplerLegacyClient, appSessionRef } })
     }
   };
 
+  const handleChange = () => {
```
lucianaDilizio, line 131: Maybe this function could have a name more specific as handleOthersCheckChange or something like that (I'm not good choosing names 😅 )

src/components/DopplerPlus/ExclusiveForm.js
```diff
@@ -18,6 +18,7 @@ import { useIntl } from 'react-intl';
 const ExclusiveForm = ({ dependencies: { dopplerLegacyClient, appSessionRef } }) => {
   const intl = useIntl();
   const _ = (id, values) => intl.formatMessage({ id: id }, values);
+  const [showMessage, setShowMessage] = useState();
```
lucianaDilizio, line 21: Here you could change `showMessage` by `showMessageBox`  and `setShowMessageBox` because is not a message, is a `textarea`.

lucianaDilizio: Maybe also it could be `messageBoxVisible`  and `setMessageBoxVisibility`

src/components/DopplerPlus/ExclusiveForm.js
fcoronelmakingsense: changed

src/components/DopplerPlus/ExclusiveForm.js
fcoronelmakingsense: changed

src/components/DopplerPlus/ExclusiveForm.js
fcoronelmakingsense: changed!

PR #1266, opened 2021-02-03
Loggly log cleaning

cbernat:
### Loggly log cleaning

- Stop login timeout error for accounts not registered (this can happen in login and does not affect anyone)
- Set all error message description as message, to avoid having all sorts of confusing fields in log


lucianaDilizio (commented):
👍

PR #1267, opened 2021-02-05
fix(session-manager): if connection is aborted leave session as is

cbernat:
### If the connection aborted by timeout leave session as it is
If the user is already logged in, and suddenly the connection is slower, it may happen that the connection is aborted due timeout. We had some entries like this, so we decided to keep the session as it is for this case.

### DIFF
--- a/src/services/session-manager.ts
+++ b/src/services/session-manager.ts
@@ -81,20 +81,29 @@ export class OnlineSessionManager implements SessionManager {
         } as AppSession, // Cast required because TS cannot resolve datahubCustomerId complexity
       );
     } catch (error) {
-      if (error.code === 'ECONNABORTED') {
-        const account =
-          this.appSessionRef.current?.status === 'authenticated'
-            ? this.appSessionRef.current.userData.user.email
-            : 'none';
+      const account =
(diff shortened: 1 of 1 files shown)


src/services/session-manager.ts
```diff
@@ -103,7 +112,6 @@ export class OnlineSessionManager implements SessionManager {
           return;
         }
       }
-      this.updateSession({ status: 'non-authenticated' });
```
andresmoschini, line 106: What happens if the timeout occurs in the first request (when the status is _unknown_ and a spinner is being shown)?

Will be the spinner visible up to the next call to _GetUserData_?

cbernat: When status is unknown, account is empty, there is no user logged, so session value would be 'non-authenticated'
We only do nothing if the connection timed out and a user is logged in. 
In the first case, status goes from 'unknown' to 'non-authenticated' when timeout fails.
image

I don´t see a way that can happen that user account is set, and it's the first time it's called. First time account shouldn´t be logged.

andresmoschini: Great explanation! I understand now, thanks!

PR #1285, opened 2021-02-18
fix(Signup): email validation with accent

fcoronelmakingsense:
- Change validation in email to dont allow accents.

image

image


src/validations.ts
```diff
@@ -1,6 +1,21 @@
...
+const accentArray = ["á","à","ã","â","é","è","ê","í","ì","î","õ","ó","ò","ô","ú","ù","û"];
+
+
+function hasAccents(
+  value: string,
+): true | false {
+  for(var i=0; i < value.length; i++){
+    for(var j=0; j < accentArray.length; j++){
+        if(value[i] === accentArray[j]){
+            return true;
```
andresmoschini, line 13: is it possible to do it declaratively?

andresmoschini:
You forgot the unit tests and the integration tests

src/validations.ts
```diff
@@ -1,6 +1,21 @@
 // eslint-disable-next-line
 const emailRegex = /^\s*((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+(\.([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])+)*)|((\x22)((((\x20|\x09)*(\x0d\x0a))?(\x20|\x09)+)?(([\x01-\x08\x0b\x0c\x0e-\x1f\x7f]|\x21|[\x23-\x5b]|[\x5d-\x7e]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(\\([\x01-\x09\x0b\x0c\x0d-\x7f]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]))))*(((\x20|\x09)*(\x0d\x0a))?(\x20|\x09)+)?(\x22)))@((([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])([a-z]|\d|-|\.|_|~|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])*([a-z]|\d|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])))\.)+(([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])|(([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])([a-z]|\d|-|\.|_|~|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])*([a-z]|[\u00A0-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF])))\s*$/i;
 const nameRegex = /^[\u00C0-\u1FFF\u2C00-\uD7FF\w][\u00C0-\u1FFF\u2C00-\uD7FF\w'`\-. ]+$/i;
+const accentArray = ["á","à","ã","â","é","è","ê","í","ì","î","õ","ó","ò","ô","ú","ù","û"];
```
cbernat, line 4: Why not a regular expression for something like detecting accents?

fcoronelmakingsense: changed

src/validations.ts
```diff
@@ -12,8 +13,12 @@ export function validateEmail(
 
   if (!emailRegex.test(value)) {
     return commonErrorKey;
+  } else {
+    const matchAccents = value.match(accentRegex);
+    if (matchAccents !== null && matchAccents[0] !== value) {
```
cbernat, line 18: Why not use test instead of match? 
```suggestion
    if (accentRegex.test(value)) {
```

cbernat: https://www.w3schools.com/jsref/jsref_regexp_test.asp

cbernat: I've talked with @fcoronelmakingsense and the aim of this regex it to detect a valid format, when it fails in detecting the whole email as correct that's when we have an accent. I demand tests for this 😄

andresmoschini, line 18: Do you think that reducing nesting is possible?

Something like this:

```suggestion
  } else if (CONDITION) {
```

It will make this code simpler and easier to read.

fcoronelmakingsense: Because with test don't work. always has true value. With match return array with the all matches

fcoronelmakingsense:
> You forgot the unit tests and the integration tests

i added now

src/validations.ts
fcoronelmakingsense: i try use something like this `if(value.match(accentRegex) !== null && value.match(accentRegex) [0] !== value)` but the validator says `value.match(accentRegex) [0]` its possibly null

src/validations.ts
```diff
@@ -12,8 +13,11 @@ export function validateEmail(
 
   if (!emailRegex.test(value)) {
     return commonErrorKey;
+  } else {
+    if(accentRegex.test(value)) {
+      return 'validation_messages.error_has_accents';
+    }
```
andresmoschini, line 19: Now, you can reduce the nesting  😀

```suggestion
  } else if(accentRegex.test(value)) {
    return 'validation_messages.error_has_accents';
  }
```

fcoronelmakingsense: fixed

PR #1293, opened 2021-02-22
[DW-751] Implement adapter and double of Field History API

andresmoschini:
Hi @FromDoppler/doppler-webapp and @FromDoppler/kreitech teams!

This is the adapter to access our Field History API (it does not exist yet).

### DIFF
--- a/src/services/doppler-api-client.test.ts
+++ b/src/services/doppler-api-client.test.ts
@@ -52,7 +52,7 @@ describe('HttpDopplerApiClient', () => {
     expect(request).toBeCalledTimes(1);
     expect(result).not.toBe(undefined);
     expect(result.success).toBe(true);
-    expect(result.value.amountSubscribers).not.toBe(undefined);
+    expect(result.success && result.value.amountSubscribers).not.toBe(undefined);
   });
 
   it('should set throw error when list does not exist', async () => {
@@ -124,7 +124,7 @@ describe('HttpDopplerApiClient', () => {
       expect(request).toBeCalledTimes(1);
       expect(result).not.toBe(undefined);
       expect(result.success).toBe(true);
-      expect(result.value.email).toEqual('test@test.com');
+      expect(result.success && result.value.email).toEqual('test@test.com');
     });
 
     it('should get a subscriber with unsubscribed by hard status', async () => {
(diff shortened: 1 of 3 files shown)


andresmoschini:
We have decided to use Date in place of strings ith ISO Date Format and I have also added tests. So, in my opinion, it is ready to review and then merge.

src/services/doppler-api-client.test.ts
```diff
@@ -531,4 +533,99 @@ describe('HttpDopplerApiClient', () => {
...
+
+    it('should call with the right parameters and parse successful response', async () => {
+      // Arrange
+      const subscriberEmail = 'a@a.com';
+      const fieldName = 'MiPermiso';
+      const fieldsResult = {
+        data: {
+          items: [
+            {
+              subscriberEmail: 'email@a.com',
```
cbernat, line 563: Here shouldn't be the `subscriberEmail`--> `a@a.com` instead of `email@a.com`?

andresmoschini: Yes, but it is the same, right?

I mean, what does the server return in `subscriberEmail` and `fieldName` is the server's responsibility. Right?

cbernat: Yes, but it's confusing to read, I'm questioning you because of this. Imagine ourselves in the future :trollface:

andresmoschini: Ok, thanks, changing...

src/services/doppler-api-client.test.ts
```diff
@@ -528,7 +530,102 @@ describe('HttpDopplerApiClient', () => {
...
+        expect(item).toMatchObject({
+          date: expect.any(Date),
+          fieldName: expect.any(String),
+          fieldType: expect.any(String),
+          originIP: expect.any(String),
+          originType: expect.any(String),
+          subscriberEmail: expect.any(String),
+          value: expect.any(String),
+        });
+      }
```
andresmoschini, line 628: Do you know a better way to do this with Jest? I mean, I would prefer to remove the foreach and using something like this:

```suggestion
      expect((result.success && result.value.items) || []).All.toMatchObject({
        date: expect.any(Date),
        fieldName: expect.any(String),
        fieldType: expect.any(String),
        originIP: expect.any(String),
        originType: expect.any(String),
        subscriberEmail: expect.any(String),
        value: expect.any(String),
      });
```

cbernat: I don´t know sorry...

PR #1296, opened 2021-02-24
chore: update GDPR history related API

andresmoschini:
As [we discussed with the team](https://makingsense.slack.com/archives/CGJ72QFQX/p1614107973070900), think that these URLs will match better our future scenarios:
* `GET /accounts/${userAccount}/subscribers/${subscriberEmail}/permissions-history` (for the complete history of all subscriber's permission fields, maybe with an option for CSV)
* `GET /accounts/${userAccount}/subscribers/${subscriberEmail}/permissions-history/${fieldName}` (for the history of a single permission field)

### DIFF
--- a/src/services/doppler-api-client.test.ts
+++ b/src/services/doppler-api-client.test.ts
@@ -534,14 +534,14 @@ describe('HttpDopplerApiClient', () => {
     });
   });
 
-  describe('getSubscriberFieldHistory', () => {
+  describe('getSubscriberPermissionHistory', () => {
     it('should get an error when the response is not valid', async () => {
       // Arrange
       const request = jest.fn(async () => {}); // Invalid response
       const dopplerApiClient = createHttpDopplerApiClient({ request });
 
       // Act
-      const result = await dopplerApiClient.getSubscriberFieldHistory({
+      const result = await dopplerApiClient.getSubscriberPermissionHistory({
         subscriberEmail: 'a@a.com',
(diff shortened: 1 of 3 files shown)


src/services/doppler-api-client.test.ts
```diff
@@ -609,7 +609,7 @@ describe('HttpDopplerApiClient', () => {
           Authorization: `token ${jwtToken}`,
         },
         method: 'GET',
-        url: `/accounts/${accountEmail}/subscribers/${subscriberEmail}/fields/${fieldName}/history`,
+        url: `/accounts/${accountEmail}/subscribers/${subscriberEmail}/permissions-history/${fieldName}`,
```
lucianaDilizio, line 612: Is it a good practice to get the information using the fieldname instead the field ID? Same with subscriberEmail and acountEmail. 🤔

lucianaDilizio:
You should update API URLs in PR description 😅

src/services/doppler-api-client.test.ts
andresmoschini: I think that it is when it is possible. For example, in Lists, the name is a kind of description, not an identifier.

But the field name is a kind of identifier, so, I think that we can use it in the URLs, it is nice and self-descriptive.

andresmoschini:
Hi @FromDoppler/kreitech, in this PR I am updating #1293

PR #1324, opened 2021-04-01
DW-823: add security headers

cbernat:
### Add security headers
Added new security headers as in security report indicated. Only 3 of 5 mentioned in the report for now.

✔️  **[Strict-Transport-Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security):** force https on first http request from browser to prevent man in the middle attacks.
❌  **[Content-Security-Policy:](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy)** opted out for now, allows to get or add content only from safe sites if configured, this is waiting for security consultant's reply. If we need to add every url for all tools we use, this would be not so convenient. More information on CSP [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)
✔️ **[X-Content-Type-Options:](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options)** The X-Content-Type-Options response HTTP header is a marker used by the server to indicate that the MIME types advertised in the Content-Type headers should not be changed and be followed. This is a way to opt out of MIME type sniffing, or, in other words, to say that the MIME types are deliberately configured.
✔️ **[Referrer-Policy:](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy)** The Referrer-Policy HTTP header controls how much referrer information (sent via the Referer header) should be included with requests. Aside from the HTTP header, you can set this policy in HTML.
❌ **[Permissions-Policy:](https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy)** Feature Policy allows web developers to selectively enable, disable, and modify the behavior of certain features and APIs in the browser. It is similar to Content Security Policy but controls features instead of security behavior. ⚠️ The Feature-Policy header has now been renamed to Permissions-Policy in the [W3C spec](https://w3c.github.io/webappsec-permissions-policy/#feature-policy-http-header-field).

Also removed unused conf files for nginx (they were used when webapp had two servers apart from the swarm we have now)

image

### Referrer-Policy
Changed referrer to `origin-when-cross-origin`
image

For same origin resources we get the full referrer url like this:
image


aferraromedina (commented):
👍

conf/site-local.conf
```diff
@@ -29,6 +29,9 @@ server {
                 expires off;
                 add_header Cache-Control "no-store";
                 add_header X-Frame-Options "DENY";
+                add_header Strict-Transport-Security "max-age=31536000; includeSubDomains" always;
+                add_header X-Content-Type-Options nosniff;
+                add_header Referrer-Policy 'no-referrer';
```
andresmoschini, line 34: Do we not use the referrer for CORS request for _GetUserData_ with Doppler Cookie?

cbernat: I will test this local, pointing to production as we spoke, to test further. There is an option to send the domain only for CORS, if needed I can use that.

cbernat: I will test this further, thanks!

cbernat: I've researched this a bit more, and `origin-when-cross-origin` seems to be the best fit.
Some examples:

image

We will not have troubles using the `referer` or `origin` header for inside the same site, and in a different site, only the origin will be sent as both the `origin` and `referer` headers.

andresmoschini: 👍

PR #1339, opened 2021-04-20
feat(Signup): add utms

fcoronelmakingsense:
- get utms from url and save in local storage the history in UtmCokies.
- add utm_source, utm_campaign, utm_medium and utm_term in the model when the user create the account.

image

### DIFF
--- a/src/App.test.js
+++ b/src/App.test.js
@@ -57,8 +57,13 @@ const dopplerSitesClientDouble = {
   getBannerData: jest.fn(async () => rejectedPromise),
 };
 
-const createJsonParse = () => {
+const createJsonParse = (item) => {
   JSON.parse = jest.fn().mockImplementationOnce(() => {
+    if (item) {
+      const array = [];
+      array.push(item);
+      return array;
+    }
     return [];
   });
 };
@@ -668,7 +673,7 @@ describe('App component', () => {
       });
     });
 
-    it('check utm parameters', async () => {
+    it('should check utm parameters are stored', async () => {
       // Arrange
       const dependencies = {
         sessionManager: createDoubleSessionManager(),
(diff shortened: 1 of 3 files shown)


andresmoschini:
Looks great! try to add tests for the new functionality.

cbernat:
👍 Some tests would be nice!

src/App.test.js
```diff
@@ -668,6 +668,46 @@ describe('App component', () => {
...
+
+      createJsonParse();
+
+      // Act
+      act(() => {
+        render(
+          <AppServicesProvider forcedServices={dependencies}>
+            <Router
+              initialEntries={[
+                '/signup?origin=testOrigin&utm_source=test&utm_campaign=testcampaign&utm_medium=testmedium&utm_term=testterm',
```
cbernat, line 687: It would be nice to see the at least two items in cookies history. Perhaps you could assume you have one entrance as @andresmoschini suggested in the daily.

fcoronelmakingsense: update with this test

src/App.test.js
```diff
@@ -653,12 +666,104 @@ describe('App component', () => {
...
+        expect(localStorageItems['UtmCookies']).toMatch('UTMCampaign');
+        expect(localStorageItems['UtmCookies']).toMatch('testcampaign');
+        expect(localStorageItems['UtmCookies']).toMatch('UTMMedium');
+        expect(localStorageItems['UtmCookies']).toMatch('testmedium');
+        expect(localStorageItems['UtmCookies']).toMatch('UTMTerm');
+        expect(localStorageItems['UtmCookies']).toMatch('testterm');
+        expect(localStorageItems['UtmCookies']).toMatch('utmsource1');
+        expect(localStorageItems['UtmCookies']).toMatch('utmcampaign1');
+        expect(localStorageItems['UtmCookies']).toMatch('utmmedium1');
+        expect(localStorageItems['UtmCookies']).toMatch('utmterm1');
```
cbernat, line 763: Could you deserialize UtmCookies and count two instances (two items)?

src/App.test.js
```diff
@@ -653,12 +666,104 @@ describe('App component', () => {
...
+        expect(localStorageItems['UtmCookies']).toMatch('UTMCampaign');
+        expect(localStorageItems['UtmCookies']).toMatch('testcampaign');
+        expect(localStorageItems['UtmCookies']).toMatch('UTMMedium');
+        expect(localStorageItems['UtmCookies']).toMatch('testmedium');
+        expect(localStorageItems['UtmCookies']).toMatch('UTMTerm');
+        expect(localStorageItems['UtmCookies']).toMatch('testterm');
+      });
+    });
+
+    it('check utm parameters', async () => {
```
cbernat, line 716: How about this?
```suggestion
    it('should accumulate utm parameters when there was a previous navigation', async () => {
```

src/App.test.js
```diff
@@ -653,12 +666,104 @@ describe('App component', () => {
...
         expect(localStorageItems['dopplerFirstOrigin.value']).toEqual('testOrigin');
         expect(localStorageItems['dopplerFirstOrigin.date']).toBeDefined();
+        expect(localStorageItems['UtmCookies']).toBeDefined();
         const dopplerOriginDate = new Date(localStorageItems['dopplerFirstOrigin.date']);
         expect(dopplerOriginDate).toBeDefined();
         expect(dopplerOriginDate.getFullYear()).toBeGreaterThan(2018);
       });
     });
 
+    it('check utm parameters', async () => {
```
cbernat, line 676: How about being a bit more descriptive? Like this:
```suggestion
    it('should check utm parameters are stored', async () => {
```

src/App.test.js
fcoronelmakingsense: i try but its a string and when try to use json.parse in the `act` doesnt work, if you have an idea i can try

src/App.test.js
```diff
@@ -653,12 +666,104 @@ describe('App component', () => {
...
+              <App locale="en" />
+            </Router>
+          </AppServicesProvider>,
+        );
+      });
+
+      // Assert
+      const localStorageItems = dependencies.localStorage.getAllItems();
+      await waitFor(() => {
+        expect(localStorageItems['UtmCookies']).toBeDefined();
```
cbernat, line 751: You could do this here:
```suggestion
        expect(localStorageItems['UtmCookies']).toBeDefined();
        const utmCookiesObject = parse(localStorageItems['UtmCookies']);
        expect(utmCookiesObject.length).toBeGreaterThan(1);
```

Where parse would be a copy of JSON.parse, before doing the mock:
```
const parse = JSON.parse;
```

fcoronelmakingsense: changed!

cbernat:
🚀

PR #1341, opened 2021-04-21
fix(style): fix gdpr history styles

ramniel:
image-20210421-150712


andresmoschini:
Thanks @ramniel! could you add a couple of screenshot to the PR description?

src/components/Reports/GDPRPermissions/PermissionExpandableRow.js
```diff
@@ -63,7 +63,7 @@ const PermissionExpandableRow = ({
               <button
                 type="button"
                 onClick={() => setExpanded(!expanded)}
-                className={`dp-expand-results ${expanded && 'dp-open-results'}`}
+                className={`dp-expand-results${expanded ? ' dp-open-results' : ''}`}
```
andresmoschini, line 66: I understand what you did, but, even when a extra space will be rendered I prefer the following approach (if it works fine):

```suggestion
                className={`dp-expand-results ${expanded ? 'dp-open-results' : ''}`}
```

In my opinion it is cleaner. What do you think?

andresmoschini:
Nice!

@GusBaamonde did you took a look?

cbernat:
This is looking nice!
https://cdn.fromdoppler.com/doppler-webapp/build3046/#/subscribers/cbernat@makingsense.com/gdpr

image

I've noticed that each time a request is done we are printing it in the console, this is not so nice...

src/components/Reports/GDPRPermissions/PermissionExpandableRow.js
ramniel: I agree with the proposed approach. It looks cleaner.

andresmoschini:
🚀

GusBaamonde:
👍

cbernat:
👍 
@ramniel could you remove api response logging in another PR?

andresmoschini:
What happened with this PR?

PR #1358, opened 2021-05-04
fix: exclude the 'consent' field in the permission list

ramniel:
image-20210421-122751

The 'consent' field is a legacy field, which behaves differently than permission-type fields, and even has no permission history.


andresmoschini:
🚀

cbernat:
👍

PR #1373, opened 2021-05-12
DW-889: fix(utmCookies): add only one entrance for each navigation

cbernat:
### Fix utmCookies record

Each time the signup component was rendered, a new entrances was added to this field utmCookies. To fix this I've added that this addition be done only dependent on the first render.

Besides it should only be registered one navigation each time the user access from outside in Signup. 
Example:
**Scenario 1**
- The user navigates to  `/signup?origin=testOrigin&utm_source=test&utm_campaign=testcampaign&utm_medium=testmedium&utm_term=testterm`
_One entrance is registered in UtmCookie and the registered flag is set to true_
- The user then navigates to `/login`
- Then he returns to `/signup` 
_No new entrance is registered_

**Scenario 2**
- The user navigates to  `/signup?origin=testOrigin&utm_source=test&utm_campaign=testcampaign&utm_medium=testmedium&utm_term=testterm`
_One entrance is registered in UtmCookie and the registered flag is set to true_
- The user closes the tab
- The user opens a new tab and types directy the url `/signup`
_Another entrance is registered in UtmCookie with the source direct and the registered flag is set to true_

### DIFF
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -58,27 +58,6 @@ function getSource(location) {
   return utmSource;
 }
 
-const manageUtmCookies = (localStorage, utmSource, utmCampaign, utmMedium, utmTerm) => {
-  let utmCookies = JSON.parse(localStorage.getItem('UtmCookies'));
-
-  if (!utmCookies) {
-    utmCookies = [];
-  }
-
-  const newItem = {
-    date: new Date().toISOString(),
-    UTMSource: utmSource,
-    UTMCampaign: utmCampaign,
-    UTMMedium: utmMedium,
-    UTMTerm: utmTerm,
-  };
-  utmCookies.push(newItem);
-
-  localStorage.setItem('UtmCookies', JSON.stringify(utmCookies));
-
-  return utmCookies;
-};
-
 /** Prepare empty values for all fields
  * It is required because in another way, the fields are not marked as touched.
  */
@@ -96,7 +75,7 @@ const getFormInitialValues = () =>
  */
 const Signup = function ({
   location,
-  dependencies: { dopplerLegacyClient, originResolver, localStorage },
+  dependencies: { dopplerLegacyClient, originResolver, localStorage, utmCookiesManager },
 }) {
   const intl = useIntl();
   const _ = (id, values) => intl.formatMessage({ id: id }, values);
@@ -110,7 +89,8 @@ const Signup = function ({
   const utmMedium = getParameter(location, 'utm_medium');
   const utmTerm = getParameter(location, 'utm_term');
 
-  const utmCookies = manageUtmCookies(localStorage, utmSource, utmCampaign, utmMedium, utmTerm);
+  utmCookiesManager.setCookieEntry(localStorage, utmSource, utmCampaign, utmMedium, utmTerm);
+  const utmCookies = utmCookiesManager.getUtmCookie(localStorage);
 
   const addExistentEmailAddress = (email) => {
     setAlreadyExistentAddresses((x) => [...x, email]);
@@ -172,7 +152,7 @@ const Signup = function ({
       utm_campaign: utmCampaign,
       utm_medium: utmMedium,
       utm_term: utmTerm,
-      utm_cookies: utmCookies.slice(-10),
+      utm_cookies: utmCookies,
     });
     if (result.success) {
       setRegisteredUser(values[fieldNames.email]);
--- a/src/services/utm-cookies-manager.js
+++ b/src/services/utm-cookies-manager.js
@@ -1,0 +1,28 @@
+export class UtmCookiesManager {
+  constructor() {
+    this.hasRegistered = false;
+  }
+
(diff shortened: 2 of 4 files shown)


src/components/Signup/Signup.js
```diff
@@ -105,12 +105,16 @@ const Signup = function ({
...
 
-  const utmCookies = manageUtmCookies(localStorage, utmSource, utmCampaign, utmMedium, utmTerm);
+  useEffect(() => {
+    utmSource = getSource(location);
+    utmCampaign = getParameter(location, 'utm_campaign');
+    utmMedium = getParameter(location, 'utm_medium');
+    utmTerm = getParameter(location, 'utm_term');
+
+    utmCookies = manageUtmCookies(localStorage, utmSource, utmCampaign, utmMedium, utmTerm);
+  }, [location]);
```
andresmoschini, line 117: It will generate a new entry for example if the language change.

Is it fine?

The alternative is using an empty array:

```suggestion
  }, []);
```

cbernat:
Test this please.

src/components/Signup/Signup.js
cbernat: Good catch! Thanks!

cbernat:
Test this please.

src/components/Signup/Signup.js
```diff
@@ -105,12 +105,20 @@ const Signup = function ({
...
 
-  const utmSource = getSource(location);
-  const utmCampaign = getParameter(location, 'utm_campaign');
-  const utmMedium = getParameter(location, 'utm_medium');
-  const utmTerm = getParameter(location, 'utm_term');
+  let utmSource,
+    utmCampaign,
+    utmMedium,
+    utmTerm,
+    utmCookies = '';
```
andresmoschini, line 112: Why did you declare these variables out of _useEffect_ scope?

cbernat: I need this variables available outside the effect.

lucianaDilizio: What if you create a states for that variables?
const [utmCampaing, setUtmCampaign] = useState('');
...
an so on...

lucianaDilizio: or... 
`const [utmList, setUtmList] = useState({utmCampaign: '', utmSource: '' .....etc})`

```
useEffect(()=>{   
   setUtmList(
     {
        utmCampaign: getParameter(location, 'utm_campaing', 
        etc
     }
   )
}, []); 
```

src/components/Signup/Signup.js
```diff
@@ -105,12 +105,31 @@ const Signup = function ({
...
+    utmSource: '',
+    utmCampaign: '',
+    utmMedium: '',
+    utmTerm: '',
+    utmCookies: [],
+  });
 
-  const utmCookies = manageUtmCookies(localStorage, utmSource, utmCampaign, utmMedium, utmTerm);
+  useEffect(() => {
+    if (localStorage.getItem('utm_cookie_registered') === 'false') {
```
andresmoschini, line 117: Creo que no es necesario este segundo elemento `utm_cookie_registered`, me parece que alcanza con `UtmCookies`:

```suggestion
    if (!localStorage.getItem('UtmCookies')) {
```

andresmoschini: Oh, I understand, `utm_cookie_registered` value is cleaned on each navigation.

In that case, why did you not use a global variable?

BTW, I am not sure that adding a new layer over the old code is a good idea, I think that based on this new expected behavior it could be thought and coded from the scratch.

cbernat: "I think that based on this new expected behavior it could be thought and coded from the scratch."
I don't quite get the meaning of this 👀

andresmoschini: Perdón, mi ingles es muy malo.

Creo que a partir de que ahora tenemos claro como es el comportamiento esperado, se puede repensar y reescribir la implementación en lugar de solo agregar el if. Creo que puede salir algo más simple.

cbernat: Something like a service that it's inyected in signup and has the responsibility of registering and administering utm parameters? Instead of having all sprear everywhere?

andresmoschini: Jeje, no lo pensé tanto, pero podría ser.

src/services/utm-cookies-manager.js
```diff
@@ -0,0 +1,46 @@
...
+export class UtmCookiesManager {
+  constructor() {
+    this.hasRegistered = false;
+  }
+
+  getParameter(location, parameter) {
+    return extractParameter(location, queryString.parse, parameter);
+  }
+
+  getSource(location, referrer) {
```
jcamposmk, line 12: you could do something like this:

`getSource(location, referrer) {
  return this.getParameter(location, 'utm_source') ?? (referrer || 'direct');
}`

cbernat: Sounds good, but I'll bring all the parameters processed into this service to avoid duplication.

cbernat:
Test this please.

PR #1374, opened 2021-05-12
feat: enable feature to all users

ramniel:
Remove the ExperimentalFeature dependency to enable the feature for all users.


cbernat:
Nice! 👍

PR #1381, opened 2021-05-18
fix: display an error message when a response error is received

ramniel:
If there was an error in the request, it was shown as if there was no history.

Before:
Screen Shot 2021-05-18 at 11 12 20

Now:
Screen Shot 2021-05-18 at 11 13 38


andresmoschini:
❤️

PR #1382, opened 2021-05-18
feat: add a link at the bottom of the report to download the permission history

ramniel:
Screen Shot 2021-05-18 at 17 29 51

### DIFF
--- a/src/components/Reports/GDPRPermissions/SubscriberGdprPermissions.js
+++ b/src/components/Reports/GDPRPermissions/SubscriberGdprPermissions.js
@@ -3,6 +3,7 @@ import { InjectAppServices } from '../../../services/pure-di';
 import { Loading } from '../../Loading/Loading';
 import { FormattedMessage, useIntl } from 'react-intl';
 import PermissionExpandableRow from './PermissionExpandableRow';
+import { DownloadLink } from './SubscriberGdpr.styles';
 
 const SubscriberGdprPermissions = ({
   subscriber,
@@ -95,6 +96,17 @@ const SubscriberGdprPermissions = ({
           </table>
         )}
       </div>
+
+      {subscriber.downloadPermissionHistoryUrl ? (
+        <DownloadLink className="dp-cta-links">
+          <a href={subscriber.downloadPermissionHistoryUrl}>
+            <i className="ms-icon icon-download"> </i>
+            <span className="m-l-6 align-middle">
+              <FormattedMessage id="subscriber_gdpr.download_permission_history" />
+            </span>
+          </a>
+        </DownloadLink>
+      ) : null}
     </div>
   );
 };
--- a/src/services/doppler-api-client.double.ts
+++ b/src/services/doppler-api-client.double.ts
@@ -95,6 +95,8 @@ const subscriber = {
   unsubscriptionComment: 'test',
   status: 'unsubscribed_by_hard',
   score: 2,
+  downloadPermissionHistoryUrl:
+    '/accounts/test@test.com/subscribers/test@test.com/permissions-history.csv',
 };
 
 // const subscriberWithNoGDPR = {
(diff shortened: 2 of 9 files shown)


src/components/Reports/GDPRPermissions/SubscriberGdprPermissions.js
```diff
@@ -95,6 +108,17 @@ const SubscriberGdprPermissions = ({
...
+      {downloadPermissionHistoryUrl && (
+        <DownloadLink className="dp-cta-links">
+          <a href={downloadPermissionHistoryUrl}>
+            <i className="ms-icon icon-download"> </i>
+            <span className="m-l-6 align-middle">
+              <FormattedMessage id="subscriber_gdpr.download_permission_history" />
+            </span>
+          </a>
+        </DownloadLink>
+      )}
```
andresmoschini, line 121: If `downloadPermissionHistoryUrl` is false, does it render `"false"`? 

Maybe using the ternary operator is a better option, see https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

src/components/Reports/GDPRPermissions/SubscriberGdprPermissions.js
```diff
@@ -1,8 +1,20 @@
...
 import { FormattedMessage, useIntl } from 'react-intl';
 import PermissionExpandableRow from './PermissionExpandableRow';
+import { searchLinkByRel } from '../../../utils';
+import { DownloadLink } from './SubscriberGdpr.styles';
+
+const permissionHistoryLinkRel = 'docs/rels/get-permission-history-csv';
+
+const getDownloadUrl = (links) => {
+  if (links) {
+    const link = searchLinkByRel(links, permissionHistoryLinkRel)[0];
```
andresmoschini, line 13: We do not consider this logic as _presentation logic_, so, in other cases, we do not do this at the _component level_.

In place of this we parse the links at the _service level_, for example: 

https://github.com/FromDoppler/doppler-webapp/blob/eb1c3ea03a97b997f705e0fd8bf72ee326a260d8/src/services/doppler-billing-api-client.ts#L71-L78

https://github.com/FromDoppler/doppler-webapp/blob/eb1c3ea03a97b997f705e0fd8bf72ee326a260d8/src/services/doppler-billing-api-client.ts#L80-L93

src/services/doppler-api-client.double.ts
```diff
@@ -95,6 +95,24 @@ const subscriber = {
...
+        '/accounts/test@test.com/subscribers/test@test.com/permissions-history.csv?_s=XpYGFksgzI6LhSXu2T4LZLlYUHzr6GN68QWvw19m1yw6LdrNklO5qpwARR4WEJVK',
+      description: "Get subscriber's complete permission history as CSV",
+      rel: '/docs/rels/get-permission-history-csv',
+    },
+    {
+      href: '/accounts/test@test.com/lists/27766804/subscribers/test@test.com',
+      description: 'Remove Subscriber from List.',
+      rel: '/docs/rels/remove-from-list',
+    },
+  ],
```
andresmoschini, line 115: I do not see the related changes in [doppler-api-client](https://github.com/FromDoppler/doppler-webapp/blob/master/src/services/doppler-api-client.ts).

So, it will not work with real data, does not it?

src/components/Reports/GDPRPermissions/SubscriberGdprPermissions.js
ramniel: You're right. Excellent article, thanks for sharing.

cbernat:
In a commit you have this
image
I think translation is not the feature per se but the link, perhaps this could be added as a chore:
`chore: add translation of history download link` 
What do you think?

cbernat:
Test this please.

ramniel:
Test this please

PR #1383, opened 2021-05-19
fix: show consent without response when receiving null permission value

ramniel:
Screen Shot 2021-05-19 at 10 33 43


andresmoschini:
❤️ 🚀

PR #1384, opened 2021-05-19
fix(Signup): send last 10 items in utm cookies

fcoronelmakingsense:
- Change in signup when create user, send the last 10 utm cookies to fix the problem when the utm cookies are a lot.

image

### DIFF
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -172,7 +172,7 @@ const Signup = function ({
       utm_campaign: utmCampaign,
       utm_medium: utmMedium,
       utm_term: utmTerm,
-      utm_cookies: utmCookies,
+      utm_cookies: utmCookies.slice(-10),
     });
     if (result.success) {
       setRegisteredUser(values[fieldNames.email]);


src/components/Signup/Signup.js
```diff
@@ -172,7 +172,7 @@ const Signup = function ({
       utm_campaign: utmCampaign,
       utm_medium: utmMedium,
       utm_term: utmTerm,
-      utm_cookies: utmCookies,
+      utm_cookies: utmCookies.slice(-10),
```
andresmoschini, line 175: Si hay menos de 10 no rompe, no?

andresmoschini: Tampoco utmCookies puede ser null, correcto?

fcoronelmakingsense: image

fcoronelmakingsense: @andresmoschini

andresmoschini:
🚀 

Si `utmCookies` no puede ser null, dale para adelante...

PR #1391, opened 2021-05-26
feat: add modification source column

ramniel:
Screen Shot 2021-05-27 at 16 29 57


ramniel:
Test this please.

src/components/Reports/GDPRPermissions/PermissionExpandableRow.test.js
```diff
@@ -175,4 +225,64 @@ describe('PermissionExpandableRow component', () => {
...
+    await waitFor(() => {
+      expect(container.querySelector('.dp-table-results')).toBeInTheDocument();
+      expect(getByText('subscriber_gdpr.modification_source')).toBeInTheDocument();
+      const tableNode = container.querySelector('.dp-table-results');
+      expect(tableNode.tHead.rows[0].cells.length).toBe(4);
+      expect(tableNode.tBodies[0].rows[0].cells.length).toBe(4);
+    });
+  });
+
+  it("shouldn't show origin column when origin type is received", async () => {
```
cbernat, line 259:
```suggestion
  it("shouldn't show origin column when origin type is not received", async () => {
```

src/components/Reports/GDPRPermissions/PermissionExpandableRow.js
```diff
@@ -118,6 +118,14 @@ const PermissionExpandableRow = ({ field, email, dependencies: { dopplerApiClien
                         <th aria-label={_('subscriber_gdpr.modification_date')} scope="col">
                           <FormattedMessage id="subscriber_gdpr.modification_date" tagName="span" />
                         </th>
+                        {permissions.some((p) => p.originType) ? (
```
lucianaDilizio, line 121: Shouldn't this column has to show always beyond having this data or not?  Is there a reason to have a permission without origin? In this case, maybe it could shows a text like `none`/`vacío`

src/components/Reports/GDPRPermissions/PermissionExpandableRow.js
```diff
@@ -130,16 +138,17 @@ const PermissionExpandableRow = ({ field, email, dependencies: { dopplerApiClien
...
                           <td>
-                            <PermissionValue key={index} value={value} />
+                            <PermissionValue value={value} />
                           </td>
                           <td>{originIP}</td>
                           <td>
-                            <FormattedDate key={index} value={date} />
+                            <FormattedDate value={date} />
                           </td>
+                          {originType ? <td>{originType}</td> : null}
```
lucianaDilizio, line 151: In this case I'd show a text like: `none`/`vacío` or maybe an empty string instead `null`

src/components/Reports/GDPRPermissions/PermissionExpandableRow.js
lucianaDilizio: If this is something that we don't have from the BE yet, I think that this is ok like that.

src/components/Reports/GDPRPermissions/PermissionExpandableRow.js
lucianaDilizio: As the comment before, if this is something that we don't have from the BE yet, I think that this is ok like that.

src/components/Reports/GDPRPermissions/PermissionExpandableRow.test.js
```diff
@@ -40,6 +40,48 @@ describe('PermissionExpandableRow component', () => {
...
+        value: {
+          items: [
+            {
+              subscriberEmail: 'test@fromdoppler.com',
+              fieldName: 'Permiso2',
+              fieldType: 'permission',
+              date: new Date('2021-02-10T15:22:00.000Z'),
+              value: 'true',
+              originIP: '181.167.226.47',
+              originType: '',
```
lucianaDilizio, line 77: If we don't receive this field yet, maybe the real case to test this, it should be without this value at all.

cbernat:
Demo and production builds are ok.
https://cdn.fromdoppler.com/doppler-webapp/build3104/#/subscribers/cbernat@makingsense.com/gdpr
https://cdn.fromdoppler.com/doppler-webapp/demo-build3104/#/subscribers/cbernat@makingsense.com/gdpr

src/components/Reports/GDPRPermissions/PermissionExpandableRow.js
cbernat: We don`t have yet this information from API so, the column should only be shown when the data is ready.

src/components/Reports/GDPRPermissions/PermissionExpandableRow.test.js
ramniel: Thanks!

src/components/Reports/GDPRPermissions/PermissionExpandableRow.js
ramniel: Thanks for the clarification!

src/components/Reports/GDPRPermissions/PermissionExpandableRow.test.js
ramniel: Sounds good to me, let's test the case when the property is not received. Thanks!

ramniel:
I see that I was missing the colon in the column name (Source:). I'll fix it

andresmoschini:
❤️

PR #1392, opened 2021-05-26
feat(Signup): add hidden field & load zoho script

jcamposmk:
add hidden field & load zoho script

### DIFF
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import { Link } from 'react-router-dom';
 import { Helmet } from 'react-helmet';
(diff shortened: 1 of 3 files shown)


src/components/Signup/Signup.js
```diff
@@ -112,6 +112,19 @@ const Signup = function ({
...
+  
+    script.src = 'https://crm.zoho.com/crm/javascript/zcga.js';
+    script.async = true;
+  
+    document.body.appendChild(script);
+  
+    return () => {
+      document.body.removeChild(script);
+    }
+  }, []);
```
andresmoschini, line 126: We are using [Helmet](https://github.com/nfl/react-helmet) for this kind of thing.

andresmoschini:
Is adding tests possible? for example one for verifying that the signup header has the script and the input and another to verify that the login page does not.

PR #1414, opened 2021-06-08
DW-937: Remove publish to docker in build-n-publish script.

cbernat:
Publish of code to nginx is done in swarm for all environments. This code is obsolete. The codes gives error, and the result is that we have cdn versioning only for integration. Then the copy fails and the process stops.

Fixes:
Remove obsolete code causing error.


andresmoschini:
👍

PR #1416, opened 2021-06-09
Dw 925 add hidden field in signup

jcamposmk:
### **Add zoho hidden field and zoho script in signup form**

**Scenario**

- The user navigates to /signup?gclid=gclidTest
- A hidden field called zc_gad is added to the signup form
- Zoho script is loaded
- Get gclid from url and save in hidden field and cookie


andresmoschini:
How is this field used?

src/components/Signup/Signup.js
```diff
@@ -234,6 +237,7 @@ const Signup = function ({
             onSubmit={onSubmit}
             validate={validate}
           >
+            <input type='hidden' id="zc_gad" name="zc_gad"/>
```
andresmoschini, line 240: How is this field used?

cbernat:
Test this please.

src/components/Signup/Signup.js
cbernat: This PR was brought into life again only to get hold of a cdn link with the script and hidden field to show to zoho, just in case they answer our question. And that's all. https://cdn.fromdoppler.com/doppler-webapp/build3180/#/signup?gclid=dfsflksdjfslkdfjs

cbernat:
I'm closing this because GCLID does not work even in zoho, so until we get any response for them, this has lower priority.

src/components/Signup/Signup.js
andresmoschini: Thanks for the clarification!

PR #1423, opened 2021-06-15
feat: add switch to enable/disable the feature

ramniel:
contact-policy-2


cbernat:
How about adding some tests?

src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -1,12 +1,14 @@
...
 import { Helmet } from 'react-helmet';
 import { useIntl } from 'react-intl';
 import HeaderSection from '../shared/HeaderSection/HeaderSection';
 import { InjectAppServices } from '../../services/pure-di';
 import { Breadcrumb, BreadcrumbItem } from '../shared/Breadcrumb/Breadcrumb';
 import { FormattedMessageMarkdown } from '../../i18n/FormattedMessageMarkdown';
+import { Switch } from '../shared/Switch/Switch';
 
 export const ContactPolicy = InjectAppServices(() => {
+  const [checked, setChecked] = useState(false);
```
cbernat, line 11: This state is refering to enabling/disabling contact policy. What do you think of adding a more descriptive state?
An example:
```suggestion
  const [active, toggleActive] = useState(false);
```

src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -1,12 +1,14 @@
...
+import React, { useState } from 'react';
 import { Helmet } from 'react-helmet';
 import { useIntl } from 'react-intl';
 import HeaderSection from '../shared/HeaderSection/HeaderSection';
 import { InjectAppServices } from '../../services/pure-di';
 import { Breadcrumb, BreadcrumbItem } from '../shared/Breadcrumb/Breadcrumb';
 import { FormattedMessageMarkdown } from '../../i18n/FormattedMessageMarkdown';
+import { Switch } from '../shared/Switch/Switch';
 
 export const ContactPolicy = InjectAppServices(() => {
```
cbernat, line 10: I don't see you are using this `InjectAppServices`
```suggestion
export const ContactPolicy = ()=> {
```

src/components/shared/Switch/Switch.js
```diff
@@ -0,0 +1,32 @@
...
+        </label>
+      </div>
+      <label htmlFor="switch">{text}</label>
+    </>
+  );
+};
+
+Switch.propTypes = {
+  id: PropTypes.string.isRequired,
+  onChange: PropTypes.func.isRequired,
```
cbernat, line 27: I'm asking but I'm not sure, can you add some tests validating these?

ramniel: The prop-type warning is printed to the test output, but the test still passes. Jest does not know what to do with the console.error because it is not an Error, and therefore passes the test since it did not throw.

cbernat:
👍

PR #1425, opened 2021-06-15
Dw 948 fix duplicate emails

jcamposmk:
### **Fix duplicate emails**
**Ticket**: https://makingsense.atlassian.net/browse/DW-948

**escenario**

- The user navigates to **/signup** and creates an account with email **'hardcoded@hardcoded.com'**
- The user confirm your account
- The user navigates to **/signup**, again. The user creates an account with email **' hardcoded@hardcoded.com'** (whitespace)
- It is not allowed to create an account, because the email already exists

### DIFF
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -153,6 +153,7 @@ const Signup = function ({
 
     const result = await dopplerLegacyClient.registerUser({
       ...values,
+      email: values[fieldNames.email].trim(),
       language: intl.locale,
       firstOrigin: originResolver.getFirstOrigin(),
       origin: originResolver.getCurrentOrigin(),
--- a/src/components/Signup/Signup.test.js
+++ b/src/components/Signup/Signup.test.js
@@ -250,4 +250,89 @@ describe('Signup', () => {
     // Assert
     await waitFor(() => expect(getByText('signup.thanks_for_registering')).toBeInTheDocument());
   });
+
+  it('should show that the email with whitespace already exists when submit', async () => {
+    // Arrange
+    const values = {
+      firstname: 'Juan',
+      lastname: 'Perez',
+      phone: '+54 223 655-8877',
+      email: ' hardcoded@hardcoded.com',
+      password: 'HarcodedPass123',
+      accept_privacy_policies: true,
+    };
+    const registerUser = jest.fn().mockImplementation((values, statusForm) => ({
+      success: false,
+      expectedError: {
+        emailAlreadyExists: true,
+      },
+    }));
+    const dependencies = {
+      ...defaultDependencies,
+      dopplerLegacyClient: {
+        ...defaultDependencies.dopplerLegacyClient,
+        registerUser,
+      },
+      utmCookiesManager: {
+        setCookieEntry: jest.fn(),
+        getUtmCookie: jest.fn().mockImplementation(() => []),
+      },
+    };
+    const location = { search: 'test', pathname: '/signup' };
+
+    // Act
+    const { container } = render(
+      <AppServicesProvider forcedServices={dependencies}>
+        <DopplerIntlProvider locale="en">
+          <Router>
+            <Signup location={location} />
+          </Router>
+        </DopplerIntlProvider>
+      </AppServicesProvider>,
+    );
+
+    act(() => {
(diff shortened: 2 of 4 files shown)


src/components/Signup/Signup.test.js
```diff
@@ -66,7 +66,7 @@ describe('Signup', () => {
     // Assert
     await waitFor(() => expect(container.querySelectorAll('.error')).toHaveLength(0));
 
-    // Act
+    // ActAct
```
andresmoschini, line 69: 🤔

src/components/form-helpers/form-helpers.js
```diff
@@ -121,7 +132,7 @@ const _FormWithCaptcha = ({
       formikProps.setErrors({
         _error: <FormattedMessageMarkdown id="validation_messages.error_unexpected_MD" />,
       });
-      setTimeout(() => formikProps.setSubmitting(false), captchaDelay);
+      timer3 = setTimeout(() => formikProps.setSubmitting(false), captchaDelay);
```
andresmoschini, line 135: ?

src/components/form-helpers/form-helpers.js
```diff
@@ -90,20 +90,31 @@ const _FormWithCaptcha = ({
...
   const originalOnSubmit = onSubmit;
   const [Captcha, verifyCaptcha] = captchaUtilsService.useCaptcha();
 
+  useEffect(() => {
+    return () => {
+      clearTimeout(timer1);
+      clearTimeout(timer2);
+      clearTimeout(timer3);
+    };
+  }, []);
```
andresmoschini, line 106: There should be a cleaner way to do this.

andresmoschini: Maybe a kind of hook? something like `useTimeout()` that avoids effects when the component has been unloaded?

src/components/Signup/Signup.test.js
cbernat: Eager to act?

src/components/form-helpers/form-helpers.js
jcamposmk: Maybe, something like this? 

**const createTimeout = useTimeout();**

inside the function it would be used as:

const verifyCaptchaAndSubmit = async (values, formikProps) => {
      createTimeout(() => formikProps.setSubmitting(true), captchaDelay);

      createTimeout((() => formikProps.setSubmitting(true), captchaDelay);

      createTimeout(() => formikProps.setSubmitting(false), captchaDelay);
}

jcamposmk: What about think ?

andresmoschini: Looks nice! 

Does `useTimeout` already exist or you will create it?

jcamposmk: I have to create it

src/components/Signup/Signup.test.js
```diff
@@ -261,29 +261,29 @@ describe('Signup', () => {
...
+        getUtmCookie: jest.fn().mockImplementation(() => []),
+      },
     };
     const location = { search: 'test', pathname: '/signup' };
 
     // Act
     const { container, getByText } = render(
       <AppServicesProvider forcedServices={dependencies}>
-        <DopplerIntlProvider locale='en'>
+        <DopplerIntlProvider locale="en">
```
cbernat, line 286: Why are you changing to ""? We try to use '' . Is due to prettier?

src/components/form-helpers/form-helpers.js
cbernat: I think that we can have this done in a separate PR.

src/components/Signup/Signup.js
```diff
@@ -153,6 +153,7 @@ const Signup = function ({
 
     const result = await dopplerLegacyClient.registerUser({
       ...values,
+      email: values[fieldNames.email].trim(),
```
cbernat, line 156: Did you test forgot password? because it happened there also according with the ticket.

jcamposmk: You're right!

cbernat:
Nice! 👍

andresmoschini:
👏👏👏

PR #1427, opened 2021-06-17
Dw 544 format numbers in plan calculator

lucianaDilizio:
### Plan Calculator: Numbers format
- Add a new function to allow format number based on configured language.
- Function also allows define between a simple intl format or compact one. 

ESP
image

ENG
image

### DIFF
--- a/src/components/ChangePlan/PlanCalculator.js
+++ b/src/components/ChangePlan/PlanCalculator.js
@@ -4,7 +4,7 @@ import { InjectAppServices } from '../../services/pure-di';
 import { Loading } from '../Loading/Loading';
 import { FormattedMessage, useIntl } from 'react-intl';
 import queryString from 'query-string';
-import { extractParameter, getPlanFee } from '../../utils';
+import { extractParameter, getPlanFee, thousandSeparatorNumber } from '../../utils';
 import { useRouteMatch, Link } from 'react-router-dom';
 
 const NavigatorTabs = ({ tabs, pathType, selectedPlanType }) => {
@@ -157,16 +157,20 @@ const PlanPriceWithoutDiscounts = ({ planData }) => {
 const PlanPricePerMonth = ({ planData }) => {
   const intl = useIntl();
   const _ = (id, values) => intl.formatMessage({ id: id }, values);
+
   return (
     <>
       <h2 className="dp-price-large">
         <span className="dp-price-large-money">US$</span>
         <span className="dp-price-large-amount">
-          {planData.discount?.discountPercentage
-            ? Math.round(
-                getPlanFee(planData.plan) * (1 - planData.discount?.discountPercentage / 100),
-              )
-            : getPlanFee(planData.plan)}
+          {thousandSeparatorNumber(
+            intl.defaultLocale,
+            planData.discount?.discountPercentage
+              ? Math.round(
+                  getPlanFee(planData.plan) * (1 - planData.discount?.discountPercentage / 100),
+                )
+              : getPlanFee(planData.plan),
+          )}
         </span>
       </h2>
       <span className="dp-for-time">{_('plan_calculator.per_month')}</span>
@@ -197,10 +201,13 @@ const PlanAgreement = ({ planData }) => {
           <strong>
             {' '}
             US$
-            {Math.round(
-              getPlanFee(planData.plan) *
-                (1 - planData.discount.discountPercentage / 100) *
-                planData.discount.monthsAmmount,
+            {thousandSeparatorNumber(
+              intl.defaultLocale,
+              Math.round(
+                getPlanFee(planData.plan) *
+                  (1 - planData.discount.discountPercentage / 100) *
+                  planData.discount.monthsAmmount,
+              ),
             )}
           </strong>
         </p>
--- a/src/utils.test.js
+++ b/src/utils.test.js
@@ -9,6 +9,8 @@ import {
   searchLinkByRel,
   logAxiosRetryError,
   addLogEntry,
+  thousandSeparatorNumber,
+  compactNumber,
 } from './utils';
 import { renderHook } from '@testing-library/react-hooks';
 import queryString from 'query-string';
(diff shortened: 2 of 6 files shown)


lucianaDilizio:
test this please

lucianaDilizio:
test this please 😄

src/utils.ts
```diff
@@ -266,3 +266,19 @@ export function searchLinkByRel(links: Link[], rel: string): Link[] {
...
+  switch (format) {
+    case 'compact':
+      return new Intl.NumberFormat('en', {
+        //@ts-ignore
+        notation: 'compact',
+        //@ts-ignore
+        compactDisplay: 'short',
+      }).format(value);
+    case 'intl':
+      return new Intl.NumberFormat(lang === 'es' ? 'de' : lang).format(value);
```
lucianaDilizio, line 280: The reason that why I'm changing to german if language is spanish, is due to for NumberFormat, group separator starts from 10000 and not from 1000 when language is spanish.

src/utils.ts
```diff
@@ -266,3 +266,19 @@ export function searchLinkByRel(links: Link[], rel: string): Link[] {
   var regularExpression = new RegExp('\\b' + escapeRegExp(rel) + '\\b');
   return links.filter((l) => regularExpression.test(l.rel));
 }
+
+export function intlFormatNumber(lang: string, format: string, value: number) {
```
cbernat, line 270: Perhaps it would be wise to change the function number, so not to be mistaken by an intl function. That's what I thought at first 😅 Something like this or any other thing:
```suggestion
export function formatNumber(lang: string, format: string, value: number) {
```

src/utils.ts
cbernat: I don`t get the idea, can you give an example?

src/utils.ts
```diff
@@ -266,3 +266,19 @@ export function searchLinkByRel(links: Link[], rel: string): Link[] {
...
+  switch (format) {
+    case 'compact':
+      return new Intl.NumberFormat('en', {
```
cbernat, line 273: For `compact` format you always use locale `en`, so the only option that really uses language option is the intl option, I think here perhaps it's best to have two separate functions. 
`compact` --> compactNumberFormat
`intl` --> languageNumberFormat (this one depends only on language)

src/utils.ts
lucianaDilizio: Sure!
For `ESP`, if you do something like: 
`new Intl.NumberFormat('es').format(1000)` it will return `1000`
and if you do something like:
`new Intl.NumberFormat('es').format(10000)` it will return `10.000`
(it has to have more than four digits to apply the group separator)

But for `DE`, if you do something like:
`new Intl.NumberFormat('de').format(1000)` it will return `1.000`
(That is what we need)

I'm not sure, but I have read that is related with something with the RAE.

src/utils.ts
lucianaDilizio: Yes, it could be separated.
I was thinking that there's no too many differences between use this function and use directly the Intl object. But I thought that maybe, in a future, we could want to add a new behavior (as return short format only for numbers greater than 9999, for example) and in this case it would be annoying to apply it in every place that we use it. 

I was thinking too, that maybe, in a future, we could also want to show a short format based on the language (for Spanish, for example, `1000` would be `1 mil`). Beyond that, it could be in separated functions anyway.

src/utils.ts
lucianaDilizio: I don't get it, this is a function that is related with i18n. It will return a number formatted based on language.

lucianaDilizio:
@jcamposmk would you like to take a look at this PR?

src/utils.ts
cbernat: Perhaps you could add a comment nearby explaining this bizarre issue.

cbernat:
Nice! 👍

src/components/ChangePlan/PlanCalculator.js
```diff
@@ -157,16 +157,20 @@ const PlanPriceWithoutDiscounts = ({ planData }) => {
...
-              )
-            : getPlanFee(planData.plan)}
+          {thousandSeparatorNumber(
+            intl.defaultLocale,
+            planData.discount?.discountPercentage
+              ? Math.round(
+                  getPlanFee(planData.plan) * (1 - planData.discount?.discountPercentage / 100),
+                )
+              : getPlanFee(planData.plan),
+          )}
```
jcamposmk, line 173: it's not necessary planData.discount?.discountPercentage again, because was already did it. 
planData.discount.discountPercentage working fine.

You could to use a constant to save the planData.discount?.discountPercentage value, as it is used on several sides. It's more simple and you use DRY. 
Maybe: 

const discountPercentage = planData.discount?.discountPercentage;

{discountPercentage
            ? Math.round(
                getPlanFee(planData.plan) * (1 - discountPercentage / 100),  etc

lucianaDilizio: Thanks for the suggestion! I have already merge this PR, but I have to apply some changes to resolve an issue with prices rounding, so I will take in count this comment!

src/utils.test.js
```diff
@@ -497,4 +499,29 @@ describe('utils', () => {
...
+
+  describe('compactNumber function', () => {
+    it('should return correct compact format for numbers (no matter what language)', () => {
+      // Assert
+      expect(compactNumber(100)).toBe('100');
+      expect(compactNumber(1000)).toBe('1K');
+      expect(compactNumber(10000)).toBe('10K');
+    });
+  });
 });
```
jcamposmk, line 527: Two phases are united: Act & Assert. 
It does not affect the operation at all, it is only a matter of organization. 
I think that @cbernat  has more knowledge about this
You can divide the phases like this:

```
describe('compactNumber function', () => {
    it('should return correct compact format for numbers (no matter what language)', () => {
      // Act
     const value1 = compactNumber(100);
     const value2 = compactNumber(1000);
     const value3 = compactNumber(10000);

      // Assert
      expect(value1).toBe('100');
      expect(value2).toBe('1K');
      expect(value3).toBe('10K');
    });
  });

```

andresmoschini: Good Idea @jcamposmk!

What is more, you can use [each](https://jestjs.io/docs/api#describeeachtablename-fn-timeout), as [we are already doing here](https://jestjs.io/docs/api#describeeachtablename-fn-timeout):

```js
describe.each`
  rawNumber  | expectedFormatted
  100        | 100
  1000       | 1K
  10000      | 10K
`('compactNumber function', ({ rawNumber, expectedFormatted }) => {
    it(`should return ${expectedFormatted} when the number is ${rawNumber} (no matter what language)`, () => {
      // Act
     const result = compactNumber(rawNumber);

      // Assert
      expect(result).toBe(expectedFormatted);
    });
  });
```

PR #1428, opened 2021-06-17
fix(console): fix console test errors

jcamposmk:
### **Fix console test errors**
Add useTimeout hook in form-helpers and clean errors in console

**escenario:** 

- Run yarn test on console
- All tests pass successfully

**Results:**

- Local
image

- Jenkins
image

### DIFF
--- a/src/hooks/useTimeout/index.js
+++ b/src/hooks/useTimeout/index.js
@@ -5,7 +5,7 @@ const useTimeout = () => {
 
   // clear the timeout.
   useEffect(() => {
-    return () => timerIds.map((timerId) => clearTimeout(timerId));
+    return () => timerIds.forEach((timerId) => clearTimeout(timerId));
   }, [timerIds]);
 
   const createTimeout = (callback, delay) => {
--- a/src/hooks/useTimeout/index.js
+++ b/src/hooks/useTimeout/index.js
@@ -1,0 +1,24 @@
+import { useCallback, useEffect, useState } from 'react';
+
+const useTimeout = () => {
+  const [timerIds, setTimerIds] = useState([]);
+
+  // clear the timeout.
+  useEffect(() => {
+    return () => timerIds.map((timerId) => clearTimeout(timerId));
+  }, [timerIds]);
+
+  const createTimeout = (callback, delay) => {
+    // Don't schedule if no delay or callback is specified.
+    if (delay === null || !callback) {
+      return;
+    }
+
+    const id = setTimeout(callback, delay);
+    setTimerIds((timerIds) => [...timerIds, id]);
+  };
+
+  return useCallback(createTimeout, []);
+};
+
+export default useTimeout;
--- a/src/hooks/useTimeout/index.test.js
+++ b/src/hooks/useTimeout/index.test.js
@@ -1,7 +1,6 @@
-import { useEffect } from 'react';
 import '@testing-library/jest-dom/extend-expect';
 import useTimeout from '.';
-import { renderHook } from '@testing-library/react-hooks';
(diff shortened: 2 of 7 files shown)


src/hooks/useTimeout/index.js
```diff
@@ -0,0 +1,24 @@
+import { useCallback, useEffect, useState } from 'react';
+
+const useTimeout = () => {
+  const [timerIds, setTimerIds] = useState([]);
+
+  // clear the timeout.
+  useEffect(() => {
+    return () => timerIds.map((timerId) => clearTimeout(timerId));
+  }, [timerIds]);
```
andresmoschini, line 9: Is it fine using `timerIds` as dependency? will it not clear existing timeouts each time you add a new one?

src/hooks/useTimeout/index.js
```diff
@@ -0,0 +1,24 @@
+import { useCallback, useEffect, useState } from 'react';
+
+const useTimeout = () => {
+  const [timerIds, setTimerIds] = useState([]);
+
+  // clear the timeout.
+  useEffect(() => {
+    return () => timerIds.map((timerId) => clearTimeout(timerId));
```
andresmoschini, line 8: I know that it is very opinionable, also, that this way is very idiomatic for some JS people.

I am not asking for a change, only starting a conversation.

In my opinion, it is better to keep functions like `map` to _map_, it means to convert each item of a list into a different item in a new list without side effects.

`clearTimeout` is a effect, for that reason, I think that `for` will be clear here.

I know that it is not for JS ecosystem, but I think that [these Eric Lippert's reflections](https://ericlippert.com/2009/05/18/foreach-vs-foreach/) applyies here.

src/hooks/useTimeout/index.js
andresmoschini: But the test shows me that it is working fine, what I am missing?

src/hooks/useTimeout/index.test.js
```diff
@@ -0,0 +1,72 @@
...
+    const delay2 = 1500;
+
+    // Act
+    const { unmount } = renderHook(() => {
+      const createTimeout = useTimeout();
+
+      useEffect(() => {
+        createTimeout(callback, delay);
+        createTimeout(callback2, delay2);
+      }, []);
```
andresmoschini, line 47: `useEffect` is not necessary here right?

```suggestion
      createTimeout(callback, delay);
      createTimeout(callback2, delay2);
```

I mean, `useTimeout` already has the `useEffect` inside, and in our real scenario, we are not using `useTimeout` inside `useEffect`.

jcamposmk: I use the useEffect in renderHook's to be able to execute the **createTimemout** function. it is the easiest way to use it.
If I don't put it, it will show me that the setTimeout has been called too many times. 

image

what I can do is simulate that these calls to **createTimeout** are made within a function, and trigger the function by clicking on a fake button.

src/hooks/useTimeout/index.js
jcamposmk: I think you're right. It causes me doubt, I will investigate

src/hooks/useTimeout/index.test.js
jcamposmk: fixed!

src/hooks/useTimeout/index.js
jcamposmk: I hadn't seen it that way, but it makes sense

jcamposmk: I guess it gives more expression to the code

cbernat:
Nice! 👍

PR #1432, opened 2021-06-18
feat: add inputs and buttons

ramniel:
contact-policy-3


src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -5,12 +5,46 @@ import HeaderSection from '../shared/HeaderSection/HeaderSection';
...
+  //TODO: refactor next block. (just to simulate form submission)
+  //BEGIN BLOCK
+  const delay = (ms) => new Promise((res) => setTimeout(res, ms));
+  const submitContactPolicyForm = async (values, { setSubmitting }) => {
+    setFormSubmitted(false);
+    await delay(1000);
+    setFormSubmitted(true);
+    setSubmitting(false);
+  };
+  //END BLOCK
```
andresmoschini, line 46: Why do you not simulate the submission in the client?

ramniel: This is only for this PR, in the next one the contact policy api will be integrated.

andresmoschini: Ok, I am fine if you do that in the following PR.

What is more, you can add a new PR with only the double, before integrating the contact policy api.

src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -5,12 +5,46 @@ import HeaderSection from '../shared/HeaderSection/HeaderSection';
...
+
+    initialValues[fieldNames.amount] = '';
+    initialValues[fieldNames.interval] = '';
+
+    return initialValues;
+  };
+
+  //TODO: refactor next block. (just to simulate form submission)
+  //BEGIN BLOCK
+  const delay = (ms) => new Promise((res) => setTimeout(res, ms));
```
jcamposmk, line 39: When you need to create a timeout, you can use useTimeout hook. 
PR: https://github.com/FromDoppler/doppler-webapp/pull/1428

This is to avoid console warnings indicating that all subscriptions must be canceled.

src/components/ContactPolicy/ContactPolicy.test.js
```diff
@@ -0,0 +1,51 @@
...
+  afterEach(cleanup);
+
+  const ContactPolicyComponent = () => (
+    <DopplerIntlProvider>
+      <ContactPolicy />
+    </DopplerIntlProvider>
+  );
+
+  it('should the inputs be disabled if the switch is not active', async () => {
+    // Arrange
```
jcamposmk, line 18:
```suggestion
    // Act
```

src/components/ContactPolicy/ContactPolicy.test.js
```diff
@@ -0,0 +1,51 @@
...
+    const inputInterval = container.querySelector('input#contact-policy-input-interval');
+
+    // Assert
+    expect(switchButton).not.toBeChecked();
+    expect(inputAmount).toBeDisabled();
+    expect(inputInterval).toBeDisabled();
+  });
+
+  it("shouldn't the inputs be disabled if the switch is active", async () => {
+    // Arrange
```
jcamposmk, line 32:
```suggestion
    // Act
```

jcamposmk: Some test could be added to verify that when the form is sent, the success message is displayed

src/i18n/en.js
```diff
@@ -135,18 +135,24 @@ const messages_en = {
...
+    save: `Save`,
     send: `Send`,
     show: `Show`,
     synchronizing: `Synchronizing`,
     ui_library_image: `https://cdn.fromdoppler.com/doppler-ui-library/${dopplerUiLibraryVersion}/img/{imageUrl}`,
     unexpected_error: `Ouch! An unexpected error occurred, please try again`,
   },
   contact_policy: {
+    amount_description: 'Do not send more than',
+    amount_unit: 'Emails',
```
jcamposmk, line 147: 'Emails' is a general text. 
It could be good to have a general section in the translations file, where the texts that are common on the platform are placed. In this way, it is avoided that the translation files grow very large when having many repeated texts. @andresmoschini

andresmoschini: In general, I agree with you.

But, in this case, `Emails` does not represent only "emails", it represents the _amount unit_, I mean, it could be "emails", "envíos", "deliveries", "contacts", "events". So, in my opinion, it should not be in a general section.

src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -5,12 +5,46 @@ import HeaderSection from '../shared/HeaderSection/HeaderSection';
...
+    const initialValues = Object.keys(fieldNames).reduce(
+      (accumulator, currentValue) => ({
+        ...accumulator,
+        [currentValue]: '',
+      }),
+      {},
+    );
+
+    initialValues[fieldNames.amount] = '';
+    initialValues[fieldNames.interval] = '';
```
lucianaDilizio, line 32: If I'm not wrong, this is not necessary. This values are initialized with the `reduce`
```
initialValues[fieldNames.amount] = '';
initialValues[fieldNames.interval] = '';
```

src/components/ContactPolicy/ContactPolicy.test.js
```diff
@@ -0,0 +1,51 @@
...
+describe('ContactPolicy component', () => {
+  afterEach(cleanup);
+
+  const ContactPolicyComponent = () => (
+    <DopplerIntlProvider>
+      <ContactPolicy />
+    </DopplerIntlProvider>
+  );
+
+  it('should the inputs be disabled if the switch is not active', async () => {
```
cbernat, line 17:
```suggestion
  it('should disable the inputs be disabled if the switch is not active', async () => {
```

src/components/ContactPolicy/ContactPolicy.test.js
```diff
@@ -0,0 +1,51 @@
...
+    const inputAmount = container.querySelector('input#contact-policy-input-amount');
+    const inputInterval = container.querySelector('input#contact-policy-input-interval');
+
+    // Assert
+    expect(switchButton).not.toBeChecked();
+    expect(inputAmount).toBeDisabled();
+    expect(inputInterval).toBeDisabled();
+  });
+
+  it("shouldn't the inputs be disabled if the switch is active", async () => {
```
cbernat, line 31:
```suggestion
  it("shouldn't disable the inputs if the switch is active", async () => {
```

src/i18n/en.js
cbernat: I agree, this text can change.

src/i18n/es.js
```diff
@@ -134,18 +134,25 @@ const messages_es = {
...
     show: `Mostrar`,
     synchronizing: `Syncronizando`,
     ui_library_image: `https://cdn.fromdoppler.com/doppler-ui-library/${dopplerUiLibraryVersion}/img/{imageUrl}`,
     unexpected_error: `¡Ouch! Ocurrio un error inesperado, intenta nuevamente`,
   },
   contact_policy: {
+    amount_description: 'No enviar mas de',
+    amount_unit: 'Emails',
+    interval_description: 'en un período de',
+    interval_unit: 'Días',
```
cbernat, line 148: I think that `Días` and `Emails` should be without capital letters. 
```suggestion
    interval_unit: 'días',
```

src/i18n/en.js
ramniel: In fact we have a (common) section where common texts are placed. We can add 'Emails' as a general text.

andresmoschini: Yes, we could. But in my opinion, it is not a `General.Email`, it is a `contact_policy.amount_unit`

src/i18n/es.js
ramniel: This comes from the design defined in the figma.
https://www.figma.com/file/YjtGdoOSCoo4UTTtGxZowV/Pol%C3%ADtica-de-Contacto?node-id=27%3A5277
Screen Shot 2021-06-22 at 14 24 47

cbernat: @mdirago is it ok like this? 🤔

cbernat: We tend to use capital letters only on Doppler entities to give them more presence, like Lists, Campaigns and so on. Here emails is referred to as a non Doppler entity, the same as days.

ramniel: Thanks for the clarification, that way the change makes more sense. Maybe @jmonacchi  can help us define what text we put.

ramniel: From https://makingsense.atlassian.net/wiki/spaces/DOP/pages/704249857/Pol+ticas+de+Contacto+V2
Text for inputs: `Enviar hasta XX Emails en un plazo de XX días`

src/components/form-helpers/form-helpers.js
```diff
@@ -526,6 +526,27 @@ export const CheckboxFieldItem = ({
...
+  onChange,
+  ...rest
+}) => (
+  <Field
+    type="number"
+    className={className}
+    name={fieldName}
+    id={id}
+    onChange={onChange}
+    placeholder={placeholder}
```
andresmoschini, line 544: These 4 fields could be part of the rest, I mean, these lines are redundant, right? or I am missing something?

```diff

export const NumberField = ({
-  className,
  fieldName,
-  placeholder,
   required,
-  id,
-  onChange,
   ...rest
 }) => (

   <Field
     type="number"
-    className={className}
     name={fieldName}
-    id={id}
-    onChange={onChange}
-    placeholder={placeholder}
     validate={createRequiredValidation(required)}
     {...rest}
```

You could even rename `fieldName` as `name` and it will be also redundant.

ramniel:
Test this please.

andresmoschini:
Test this please.

PR #1437, opened 2021-06-21
fix(PlanCalculator): fix order of plans

jcamposmk:
### **Fix order of plans**
task link: https://makingsense.atlassian.net/browse/DW-545

- The user navigates to /plan-selection/standard
- Plans are displayed in the order: by contacts, by sending and by credits

Results: 

image

### DIFF
--- a/src/components/ChangePlan/PlanCalculator.test.js
+++ b/src/components/ChangePlan/PlanCalculator.test.js
@@ -5,6 +5,7 @@ import PlanCalculator from './PlanCalculator';
 import IntlProvider from '../../i18n/DopplerIntlProvider.double-with-ids-as-values';
 import { AppServicesProvider } from '../../services/pure-di';
 import { MemoryRouter, Route } from 'react-router-dom';
+import { orderPlanTypes } from '../../utils';
 
 describe('PlanCalculator component', () => {
   afterEach(cleanup);
@@ -21,10 +22,11 @@ describe('PlanCalculator component', () => {
     },
   };
 
+  const planTypes = orderPlanTypes(['prepaid', 'subscribers', 'monthly-deliveries']);
   const planServiceDoubleBase = {
     getPlanList: () => [],
     mapCurrentPlanFromTypeOrId: () => {},
-    getPlanTypes: () => ['prepaid', 'subscribers', 'monthly-deliveries'],
+    getPlanTypes: () => planTypes,
     getBuyUrl: () => '',
   };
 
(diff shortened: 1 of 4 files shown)


src/components/ChangePlan/PlanCalculator.test.js
```diff
@@ -67,6 +69,16 @@ describe('PlanCalculator component', () => {
...
+
+      // check the order of the plan types
+      const navigatorTabs = container.querySelector('.tabs-nav');
+      const it = planTypes[Symbol.iterator]();
+      let tab = navigatorTabs.firstChild;
+      while (tab) {
+        const planType = it.next().value;
+        expect(tab).toHaveTextContent(`plan_calculator.plan_type_${planType.replace('-', '_')}`);
+        tab = tab.nextSibling;
+      }
```
andresmoschini, line 81: Is there not a high-level way of doing this? (without using `Symbol`, `next()`, and `while`)

jcamposmk: Yes. I don't know why I did it like that.It's better that way: 
```
planTypes.forEach((planType) => {
    expect(tab).toHaveTextContent(`plan_calculator.plan_type_${planType.replace('-', '_')}`);
    tab = tab.nextSibling;
});
```

src/utils.test.js
```diff
@@ -8,12 +8,77 @@ import {
...
   logAxiosRetryError,
-  addLogEntry,
+  orderPlanTypes,
 } from './utils';
 import { renderHook } from '@testing-library/react-hooks';
 import queryString from 'query-string';
 
 describe('utils', () => {
+  describe('orderPlanTypes', () => {
+    it('should sort plan types when all first plans are', () => {
```
cbernat, line 18:
```suggestion
    it('should sort plan types when all first plans are present', () => {
```

src/utils.ts
```diff
@@ -250,6 +250,26 @@ export function getPlanFee(plan: Plan): number {
   return plan.type === 'prepaid' ? (plan as PrepaidPack).price : (plan as FeaturedPlan).fee;
 }
 
+const firstPlansDefault: PlanType[] = ['subscribers', 'monthly-deliveries', 'prepaid'];
```
cbernat, line 253: How about indicating in this variable that we are talking about order or sorting?
```suggestion
const firstPlansDefaultOrder: PlanType[] = ['subscribers', 'monthly-deliveries', 'prepaid'];
```

src/utils.test.js
jcamposmk: applied!

src/utils.ts
jcamposmk: great! applied!

cbernat:
👍

PR #1438, opened 2021-06-22
test(Card): add test to Card, CardAction, CardPrice, Ribbon and CardF…

jcamposmk:
### **Add test to Card Components**

**How to test it ?**

- Run `yarn test` command in terminal
- Press p to filter by a filename regex pattern
- Write **card** and select **src/components/ChangePlan/Card.test.js**

**Results:**

image

image

### DIFF
--- a/src/components/ChangePlan/Card.js
+++ b/src/components/ChangePlan/Card.js
@@ -4,6 +4,7 @@ import classNames from 'classnames';
 import { Link } from 'react-router-dom';
 
 export const Card = ({ children, className, highlighted, ...rest }) => {
+  // TODO: double negation (!!) should be removed
   return (
     <div
       className={classNames({ 'dp-card': true, 'dp-highlighthed': !!highlighted }, className)}
@@ -24,6 +25,7 @@ export const CardAction = ({ url, children, ...rest }) => {
   );
 };
 
+// TODO: the variable name 'doNotShowSince' should be an affirmative expression. 'doNotShowSince' is hard to read
 export const CardPrice = ({ currency, children, doNotShowSince }) => {
   const intl = useIntl();
   const _ = (id, values) => intl.formatMessage({ id: id }, values);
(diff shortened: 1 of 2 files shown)


src/components/ChangePlan/Card.js
```diff
@@ -4,6 +4,7 @@ import classNames from 'classnames';
 import { Link } from 'react-router-dom';
 
 export const Card = ({ children, className, highlighted, ...rest }) => {
+  // TODO: double negation (!!) should be removed
```
lucianaDilizio, line 7: 👍

src/components/ChangePlan/Card.js
```diff
@@ -24,6 +25,7 @@ export const CardAction = ({ url, children, ...rest }) => {
   );
 };
 
+// TODO: the variable name 'doNotShowSince' should be an affirmative expression. 'doNotShowSince' is hard to read
```
lucianaDilizio, line 28: 👍

src/components/ChangePlan/Card.js
```diff
@@ -40,7 +42,7 @@ export const CardPrice = ({ currency, children, doNotShowSince }) => {
   );
 };
 
-export const Ribbon = ({ children, position = 'top-right', content }) => {
+export const Ribbon = ({ position = 'top-right', content }) => {
```
lucianaDilizio, line 45: 👍

andresmoschini:
👏 👏 👏 

@jcamposmk, maybe we could add a new step in the test to verify that the test coverage has not been reduced or that it is not lower that an arbitrary number. Then, we could increase the number gradually.

jcamposmk:
That would be great!
I don't know if technically it can be done, but it may also be interesting to add a process to verify if there is a test for new developments. It is currently a process that is being done manually. I don't know how feasible it is

PR #1439, opened 2021-06-23
feat: add contact policy api double

ramniel:
Integrate contact-policy API (https://apis.fromdoppler.com/contact-policies/swagger/index.html) with double

### DIFF
--- a/src/services/doppler-contact-policy-api-client.test.ts
+++ b/src/services/doppler-contact-policy-api-client.test.ts
@@ -1,0 +1,15 @@
+import { cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+describe('HttpDopplerContactPolicyApiClient', () => {
+  afterEach(cleanup);
+
+  //TODO: Implement real tests
+  it('dummy test', () => {
+    // Act
+    const n = null;
+
+    // Assert
+    expect(n).toBeNull();
+  });
+});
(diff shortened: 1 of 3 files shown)


src/services/doppler-contact-policy-api-client.test.ts
```diff
@@ -0,0 +1,15 @@
+import { cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+describe('HttpDopplerContactPolicyApiClient', () => {
+  afterEach(cleanup);
+
+  //TODO: Implement real tests
```
andresmoschini, line 7: Will it be done before or after the merge? 😛

andresmoschini:
Looks fine, I would integrate this with the UI in this same PR, but if you prefer to split in this way, it is ok for me

src/services/doppler-contact-policy-api-client.test.ts
ramniel: After

andresmoschini: 😕 Mmmm, in that case, I think that you could remove the file, it is totally unuseful at the moment.

Or..., you can add tests before the merge, you can mock Axios response and test the URLs requested, the mappers, how you deal with the errors, etc.

jcamposmk:
:+1:

PR #1441, opened 2021-06-24
[DW-991] apply transition UI in tabs

lucianaDilizio:
### Transition UI
- Apply transition UI when change tab. 

transition

### DIFF
--- a/src/components/ChangePlan/PlanCalculator.js
+++ b/src/components/ChangePlan/PlanCalculator.js
@@ -6,6 +6,7 @@ import { FormattedMessage, useIntl } from 'react-intl';
 import queryString from 'query-string';
 import { extractParameter, getPlanFee, thousandSeparatorNumber } from '../../utils';
 import { useRouteMatch, Link } from 'react-router-dom';
+import useTimeout from '../../hooks/useTimeout';
 
 const NavigatorTabs = ({ tabs, pathType, selectedPlanType }) => {
   const intl = useIntl();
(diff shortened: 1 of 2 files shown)


src/components/ChangePlan/PlanCalculator.js
```diff
@@ -343,6 +344,11 @@ const PlanCalculator = ({ location, dependencies: { planService, appSessionRef }
     fetchData();
   }, [actionTypes.INIT, appSessionRef, planService, planType, pathType]);
 
+  useEffect(() => {
+    setActiveClass('');
+    setTimeout(() => setActiveClass('active'));
+  }, [planType]);
+
```
jcamposmk, line 351: You can you useTimeout() hook. 
PR https://github.com/FromDoppler/doppler-webapp/pull/1428
this to avoid introducing warnings in the console and in the tests about unsubscription

lucianaDilizio: I tried, but I had some troubles to use hooks inside a hook. Maybe I'm not using it well 🤔

lucianaDilizio: Well... it's not so bad. I should do something like: 
```
useEffect(() => {
    setActiveClass('');
    createTimeout(() => setActiveClass('active'));
  }, [planType, createTimeout]);
```

I'm not convinced adding createTimeout as a parameter in useEffect hook, but it works.

jcamposmk: you would be using a function (createTimeout) inside a hook. The hook would be `useTimeout ()`

jcamposmk: similar to how useState is used, but instead of returning this:

`const [myState, setMyStet] = useState ()`

it would return this:

`const createTimeout = useTimeout ();`

silsanchez8 (commented):
👍

GusBaamonde:
:+1:

cbernat:
👍

PR #1444, opened 2021-06-25
chore: add utm_content property to CreateUser

aferraromedina:
@jcamposmk Could you help me to create a small test for this PR?

### DIFF
--- a/src/components/Signup/Signup.js
+++ b/src/components/Signup/Signup.js
@@ -89,6 +89,7 @@ const Signup = function ({
   const utmMedium = getParameter(location, 'utm_medium');
   const utmTerm = getParameter(location, 'utm_term');
   const gclid = getParameter(location, 'gclid');
+  const utmContent = getParameter(location, 'utm_content');
 
   utmCookiesManager.setCookieEntry({
     storage: localStorage,
@@ -97,6 +98,7 @@ const Signup = function ({
     UTMMedium: utmMedium,
     UTMTerm: utmTerm,
     gclid,
+    UTMContent: utmContent,
   });
   const utmCookies = utmCookiesManager.getUtmCookie(localStorage);
 
@@ -164,6 +166,7 @@ const Signup = function ({
       utm_term: utmTerm,
       utm_cookies: utmCookies,
       gclid,
+      utm_content: utmContent,
     });
     if (result.success) {
       setRegisteredUser(values[fieldNames.email]);
--- a/src/components/Signup/Signup.test.js
+++ b/src/components/Signup/Signup.test.js
@@ -277,6 +277,7 @@ describe('Signup', () => {
         getUtmCookie: jest.fn().mockImplementation(() => []),
       },
     };
+
     const location = { search: 'test', pathname: '/signup' };
 
     // Act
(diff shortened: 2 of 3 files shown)


src/components/Signup/Signup.test.js
```diff
@@ -333,6 +333,7 @@ describe('Signup', () => {
       utm_medium: null,
       utm_term: null,
       gclid: null,
+      utm_content: null
```
jcamposmk, line 336: This paramter should not be null because `location` has the query param `utm_ content`.
you are rendering `signup` with: `const location = { search: 'test', pathname: '/signup/?utm_content=test-utem-content' };`

jcamposmk: You can do two tests: the **first** with the query param `utm_content` and the **second** without it.

src/components/Signup/Signup.js
```diff
@@ -97,6 +98,7 @@ const Signup = function ({
     UTMMedium: utmMedium,
     UTMTerm: utmTerm,
     gclid,
+    UTMContent: utmContent,
```
jcamposmk, line 101: agregarla en los utm cookie nosé si sea necesario

src/components/Signup/Signup.test.js
```diff
@@ -332,6 +334,7 @@ describe('Signup', () => {
       utm_medium: null,
       utm_term: null,
       gclid: null,
+      utm_content: 'test-utm-content'
     });
   });
 });
```
jcamposmk, line 340: También sería bueno agregar otro test donde no tenga el parámetro utm_content, es decir con: 
const search = '';
const location = { search, pathname: `/signup/` };

src/components/Signup/Signup.js
cbernat: Yes we do need this in utm cookies. By requirement.

jcamposmk:
:+1:

src/components/Signup/Signup.test.js
```diff
@@ -277,7 +277,97 @@ describe('Signup', () => {
...
+        registerUser,
+      },
+      utmCookiesManager: {
+        setCookieEntry: jest.fn(),
+        getUtmCookie: jest.fn().mockImplementation(() => []),
+      },
+    };
+
+    const search = '';
+    const location = { search, pathname: `/signup/${search}` };
```
cbernat, line 370:
```suggestion
    const location = { pathname: `/signup/` };
```

cbernat: @aferraromedina I think this will work too, can you try?

src/components/Signup/Signup.test.js
```diff
@@ -277,7 +277,97 @@ describe('Signup', () => {
         getUtmCookie: jest.fn().mockImplementation(() => []),
       },
     };
-    const location = { search: 'test', pathname: '/signup' };
+
+    const search = '?utm_content=test-utm-content';
```
cbernat, line 281: I think this section to validate utm should be added in a separate test, right now I see this code as part of this test: `should show that the email with whitespace already exists when submit` 🤔

src/components/Signup/Signup.test.js
aferraromedina: yes!

src/components/Signup/Signup.test.js
```diff
@@ -332,6 +333,95 @@ describe('Signup', () => {
       utm_medium: null,
       utm_term: null,
       gclid: null,
+      utm_content: null,
+    });
+  });
+
+  it('should show that utm_content param is inluded in request', async () => {
```
cbernat, line 340:
```suggestion
  it('should show that utm_content param is included in request', async () => {
```

src/components/Signup/Signup.test.js
cbernat: Before we had 2 tests, one for utm_content empty and another one with utm_content null, now I see only one O_o

aferraromedina: yes, empty utm_content test was equal to the existing test "should show that the email with whitespace already exists when submit" so i deleted it.

cbernat:
I think this is ready to merge @aferraromedina 🚀

PR #1445, opened 2021-06-25
DBD-10 big query template

kapoc3:
Component for bigquery control panel  with test

Notes
 
- with out styles, only use ui library components
- The component is in diferent route, in the next PR the url component be http://localhost:3000/integrations/big-query
- The test include all actions of buttons in screen and all posible way
- The code not include the moks for service, in next PR include this changes

Initial screen
image

when the input text is not email 
image

when thie input text is empty 
image

when the input text is a email 
image

When changes Saved
image

when button remove click
image

image


src/i18n/es.js
```diff
@@ -25,9 +25,9 @@ const subjectBlockedAccountInvalidPassword = `?subject=Cuenta%20bloqueada%20por%
 const urlControlPanel = `${urlDopplerLegacy}/ControlPanel`;
 const urlBuyMonthly = `${urlControlPanel}/AccountPreferences/UpgradeAccount?Plan=monthly`;
 const urlSiteTracking = `${urlControlPanel}/CampaignsPreferences/SiteTrackingSettings`;
-const urlSiteFromSignup = `${urlSite}/`;
-const urlSiteFromLogin = `${urlSite}/`;
-const urlSiteFromForgot = `${urlSite}/`;
+const urlSiteFromSignup = `${urlSite}/?utm_source=app&utm_medium=landing&utm_campaign=signup`;
+const urlSiteFromLogin = `${urlSite}/?utm_source=app&utm_medium=landing&utm_campaign=login`;
+const urlSiteFromForgot = `${urlSite}/?utm_source=app&utm_medium=landing&utm_campaign=restablecimiento-contrasenia`;
```
andresmoschini, line 30: Is it fine?

src/components/Integrations/BigQuery/ControlPanel/ControlPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+const ControlPanel: React.FunctionComponent<IControlPanelProps> = (props) => {
+  const intl = useIntl();
+  const i18 = useCallback(
+    (id: string, values?: any) => {
+      return intl.formatMessage({ id: id }, values);
+    },
+    [intl],
+  );
+
+  const [values, setValues] = useState<string[]>([]);
```
andresmoschini, line 22: What are these values? emails?

In that case, what do you think about renaming it?

```suggestion
  const [emails, setEmails] = useState<string[]>([]);
```

src/i18n/es.js
kapoc3: I got these changes when I did the rebase

src/components/Integrations/BigQuery/ControlPanel/MultiInputContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+}
+
+const MultiInput = (props: MultiInputProps) => {
+  const intl = useIntl();
+  const i18 = useCallback(
+    (id: string, values?: any) => {
+      return intl.formatMessage({ id: id }, values);
+    },
+    [intl],
+  );
```
andresmoschini, line 54: If I am not wrong, in general, we call this function `_`, right?

src/components/Integrations/BigQuery/ControlPanel/MultiInputContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+    onRemove();
+  };
+
+  const handleAdd = () => {
+    let hasError = false;
+    const newInputs = inputConfigs?.map((input) => {
+      if (input.value === '') {
+        input.errorMessage = i18('big_query.plus_error_message_empty');
+        hasError = true;
+      } else if (!/^[a-zA-Z0-9]+@[a-zA-Z0-9]+\.[A-Za-z]+$/.test(input.value)) {
```
andresmoschini, line 96: We should already have an email regular expression defined.

src/components/Integrations/BigQuery/ControlPanel/MultiInputContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+      } else {
+        input.className = 'style_not_empty';
+        input.errorMessage = '';
+      }
+
+      return input;
+    });
+
+    if (hasError) {
+      setInputConfigs([...newInputs]);
```
andresmoschini, line 108: Why do you clone this array? is it useful in any way?

```suggestion
      setInputConfigs(newInputs);
```

src/i18n/es.js
```diff
@@ -354,14 +378,15 @@ visitantes poseen un Email que Doppler ha identificado y cuántos no. ¿Necesita
...
     upgrade_account_free_MD: `
-Accede a **Reportes detallados** para entender el comportamiento de los
-visitantes de tu Sitio Web o E-commerce. Descubre cuáles son las páginas más
-visitadas, cuántos visitantes ya están en tus Listas de Suscriptores y cuántos no.
-¿Quieres saber más? Presiona [HELP](${urlHelpAdvancedReports}).
+Contrata un Plan Pago y accede a Reportes detallados sobre el
+comportamiento de los usuarios en tu Sitio Web o E-commerce.
+Descubre cuáles son las páginas más visitadas, cuántos visitantes
+poseen un Email que Doppler ha identificado y cuántos no.
+¿Necesitas ayuda? Presiona [HELP](${urlHelpAdvancedReports}).
```
andresmoschini, line 385: Are these changes fine?

src/components/Integrations/BigQuery/ControlPanel/ControlPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+    },
+    [setMessage, setSaved],
+  );
+
+  const onClickBtnHandler = useCallback(() => {
+    setMessage(true);
+    setSaved(false);
+  }, [setMessage, setSaved, values]);
+
+  const MessageMemo = useMemo(() => {
```
andresmoschini, line 44: What is the advantage of memoized it?

src/components/Integrations/BigQuery/ControlPanel/ControlPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+      if (isSaved) {
+        setMessage(isSaved);
+        setSaved(isSaved);
+      } else {
+        setMessage(isSaved);
+        setSaved(isSaved);
+      }
+    },
+    [setMessage, setSaved],
+  );
```
andresmoschini, line 37: What do you think about having two differnet functions, one for "save" and another one "cancel"?

src/i18n/es.js
andresmoschini: Mmm, probably you have resolved conflicts in a wrong way.

src/components/Integrations/BigQuery/ControlPanel/MultiInputContainer.tsx
```diff
@@ -0,0 +1,143 @@
...
+const uniqueId = function (name: String) {
+  return (
+    name +
+    '-xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
+      let r = (Math.random() * 16) | 0,
+        v = c === 'x' ? r : (r & 0x3) | 0x8;
+      return v.toString(16);
+    })
+  );
+};
```
andresmoschini, line 21: What do you think about moving this to `Utils`?

src/components/Integrations/BigQuery/ControlPanel/MultiInputContainer.tsx
cbernat: We do have a regex and it is here https://github.com/FromDoppler/doppler-webapp/blob/eb1c3ea03a97b997f705e0fd8bf72ee326a260d8/src/validations.ts#L2

src/components/Integrations/BigQuery/ControlPanel/index.ts
```diff
@@ -0,0 +1,2 @@
+import ControlPanel from './ControlPanel';
```
cbernat, line 1: How about avoiding the use of typescript components as we did last time? that way there is only one file for each component.

src/i18n/es.js
cbernat: We've removed all utm parameters from this site, this shouldn`t be as a change.

kapoc3: ok I'm going to check what could have happened

src/components/Integrations/BigQuery/ControlPanel/MultiInputContainer.tsx
jcamposmk: yes, an example: `const _ = (id, values) => intl.formatMessage({ id: id }, values);`

src/components/Integrations/BigQuery/ControlPanel/ControlPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+              <span>{i18('big_query.plus_step_one')}</span>
+              <strong>
+                <FormattedMessageMarkdown id="big_query.plus_body_step_one_HD" />
+              </strong>
+              <p className="p-t-36">{i18('big_query.plus_step_one_paragraph')}</p>
+            </div>
+
+            <div className="dp-rowflex p-t-48">
+              <ul className="field-group">
+                <MultiInputContainer
```
cbernat, line 112: When we use inputs and handle submit, we do this in the context of a form. For example [agencies form](https://github.com/FromDoppler/doppler-webapp/blob/master/src/components/DopplerPlus/AgenciesForm.js) or you can take a look at login component or signup. To handle forms we use formik, and we have a pre defined way to work with formik. You can see all form helpers [here](https://github.com/FromDoppler/doppler-webapp/blob/master/src/components/form-helpers/form-helpers.js) Also this type of form you are using is being worked in another team, @ramniel is on it, perhaps you can wait for this to be ready.

src/components/Integrations/BigQuery/ControlPanel/ControlPanel.tsx
jcamposmk: I do not understand. If `isSaved` is `true` run this:
```
setMessage(isSaved);
setSaved(isSaved);
```
but if it is false it also executes the same. It would be better to remove the if-else block and just write:
```
(isSaved: boolean) => {
    setMessage(isSaved);
    setSaved(isSaved);
}
```

I think `setSaved` and `setMessage` are not necessary to add them to the **dependency array**.

src/components/Integrations/BigQuery/ControlPanel/Input.tsx
```diff
@@ -25,9 +25,8 @@ const Input = (props: InputProps) => {
   const ButtonRemove = useMemo(() => {
     if (className === 'style_not_empty') {
       return (
-        <button type="button" onClick={handleRemove}>
-          {' '}
-          -{' '}
+        <button type="button" onClick={handleRemove} data-testid="btn-remove">
```
jcamposmk, line 28: why `data-testid` attribute? you can react elements with `getByText` in the tests, or you can container.querySelector

src/components/Integrations/BigQuery/ControlPanel/index.ts
lucianaDilizio: I'm agree. In this project we don't usually use typescript in components. I think that it would be better to keep it in this way.

src/components/Integrations/BigQuery/ControlPanel/MultiInputContainer.tsx
jcamposmk: We use formik, maybe you can use FieldArray
https://formik.org/docs/api/fieldarray

jcamposmk: **formik** provides you helpers to dynamically add / remove fields

src/components/Integrations/BigQuery/ControlPanel/MultiInputContainer.tsx
lucianaDilizio: It is necessary to use `useCallback` here?

src/components/Integrations/BigQuery/ControlPanel/ControlPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+    [setMessage, setSaved],
+  );
+
+  const onClickBtnHandler = useCallback(() => {
+    setMessage(true);
+    setSaved(false);
+  }, [setMessage, setSaved, values]);
+
+  const MessageMemo = useMemo(() => {
+    let messageText;
```
jcamposmk, line 45:
```suggestion
    const messageText = <p>{_(saved ? `big_query.plus_message_saved` : 'big_query.plus_message_remember' )}</p>;
```

src/components/Integrations/BigQuery/ControlPanel/ControlPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+    if (saved) {
+      messageText = <p>{i18('big_query.plus_message_saved')}</p>;
+    } else {
+      messageText = <p>{i18('big_query.plus_message_remember')}</p>;
+    }
+
+    if (message === true) {
```
lucianaDilizio, line 52: This is the same than just `if (message) {`

src/components/Integrations/BigQuery/ControlPanel/ControlPanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+      return (
+        <div className="p-t-96" data-testid="message-info">
+          <div className={`dp-wrap-message dp-wrap-${saved ? 'success' : 'info'}`}>
+            <span className="dp-message-icon"></span>
+            <div className="dp-content-message">{messageText}</div>
+          </div>
+        </div>
+      );
+    }
```
jcamposmk, line 61: It's not necessary `if(necessary === true)`. You can simply do: `if(message)` or `if(!message)`.

jcamposmk: remove data-testid, it's not necessary. There are other ways

src/components/Integrations/BigQuery/ControlPanel/ControlPanel.tsx
kapoc3: Suggest accepted and applied !!!

kapoc3: > What do you think about having two differnet functions, one for "save" and another one "cancel"?

yes i agree with you !!!

kapoc3:
this PR continue in #1467

PR #1449, opened 2021-06-28
fix(ForgotPassword): add dependency doppler sites client

fcoronelmakingsense:
fix tests in forgot password adding the dependency of doppler sites client to get the default banner.


jcamposmk:
still missing signup: 
image

fcoronelmakingsense:
fixed

PR #1452, opened 2021-06-29
test(ChangePlan): add test to change plan components

jcamposmk:
### **Add ChangePlan component test**
Task: https://makingsense.atlassian.net/browse/DW-993

**How to test it?** 

Run `yarn test` command in terminal
Press `p` to filter by a filename regex pattern
Write card and select `src/components/ChangePlan/ChangePlan.test.js`

**Result:**

image


src/components/ChangePlan/ChangePlan.test.js
```diff
@@ -0,0 +1,442 @@
...
+describe('BigDataBullet component', () => {
+  it('should render BigDataBullet', async () => {
+    // Arrange
+    const text = 'fake test';
+
+    // Act
+    const { getByText } = render(<BigDataBullet>{text}</BigDataBullet>);
+
+    // Assert
+    expect(getByText(text)).toBeInTheDocument();
```
cbernat, line 56: Don`t we have a special bullet class for big query also?

jcamposmk: these are all who are:
- StarBullet
- BasicBullet
- BigDataBullet

PR #1455, opened 2021-06-30
DW-992-allow-plan-upgrade

silsanchez8:
## Modifications:

- fix(change-plan): Always show button to upgrade current plan. _( Also when current plan is the highest plan)_

- fix(plan-service): Remove unused `deadEnd `property from plan-service, tests and code.

image

### DIFF
--- a/src/components/ChangePlan/ChangePlan.js
+++ b/src/components/ChangePlan/ChangePlan.js
@@ -178,7 +178,7 @@ export const CardWithPrice = ({ path, showFeatures, currentPlanType, promoCode }
       <CardPrice doNotShowSince={path.current} currency="US$">
         {thousandSeparatorNumber(intl.defaultLocale, path.minimumFee)}
       </CardPrice>
-      {path.current && !path.deadEnd ? (
+      {path.current ? (
         <>
           {/* TODO: it's not necessary. Can be used 'search' and add promoCode
           https://reactrouter.com/web/api/Link */}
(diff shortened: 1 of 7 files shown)


jcamposmk:
:+1:

src/components/ChangePlan/ChangePlan.js
```diff
@@ -178,7 +178,7 @@ export const CardWithPrice = ({ path, showFeatures, currentPlanType, promoCode }
       <CardPrice doNotShowSince={path.current} currency="US$">
         {thousandSeparatorNumber(intl.defaultLocale, path.minimumFee)}
       </CardPrice>
-      {path.current && !path.deadEnd ? (
```
cbernat, line 181: Do we use `deadEnd` property for anyting else?

silsanchez8: I'll use this property in my next ticket, when I have to detect which is the highest plan in the list. In case this property is true, I will show the new UI for that case. 😉

lucianaDilizio:
💪

src/components/ChangePlan/ChangePlan.js
silsanchez8: I was deeply analizing it and It's really not necessary to mantein `deadEnd `property. I have the user plan session id, so I send it to BannerUpgrade component as a property, and It lets me compare values and check if there is just one plan in the list and if it's the current session plan. So, I should remove `deadEnd` property, are you agree? @cbernat  @lucianaDilizio

cbernat: @silsanchez8 Then in your next PR you could refactor what you need into the new UI in plan calculator. And remove what is not being used anymore.

cbernat:
Please check if these tests are no longer valid: https://github.com/FromDoppler/doppler-webapp/blob/83fced39f7e3bfcd730cb41dc272bdf07eb95c02/src/components/ChangePlan/ChangePlan.test.js#L148

cbernat:
Very neat @silsanchez8 ! How about changing the second commit into a refactor one?
image
`refactor(planservice): remove unused deadEnd property ... `

PR #1456, opened 2021-07-01
feat: add function to update account settings and tests

ramniel:
- [x] add tests to get and update account settings.
- [x] add function to update account settings.
- [x] add doubles for updating account settings.

### DIFF
--- a/src/services/doppler-contact-policy-api-client.test.ts
+++ b/src/services/doppler-contact-policy-api-client.test.ts
@@ -1,15 +1,124 @@
-import { cleanup } from '@testing-library/react';
-import '@testing-library/jest-dom/extend-expect';
+import { AxiosStatic } from 'axios';
+import { RefObject } from 'react';
+import { AppSession } from './app-session';
+import { DopplerLegacyUserData } from './doppler-legacy-client';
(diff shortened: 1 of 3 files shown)


cbernat:
Why not integrate the get into the component before adding more loose method into an unused api?

src/services/doppler-contact-policy-api-client.test.ts
```diff
@@ -1,15 +1,121 @@
...
+    const dopplerContactPolicyApiClient = createHttpDopplerContactPolicyApiClient({ request });
+
+    // Act
+    const result = await dopplerContactPolicyApiClient.getAccountSettings(emailAccount);
+
+    // Assert
+    expect(request).toBeCalledTimes(1);
+    expect(result).not.toBe(undefined);
+    expect(result.success).toBe(true);
+    expect(result.value.excludedSubscribersLists).not.toBe(undefined);
```
cbernat, line 63: How about validating this other values as well?
  ```
  active: true,
  emailsAmountByInterval: 20,
  intervalInDays: 7,
  ```

ramniel:
> Why not integrate the get into the component before adding more loose method into an unused api?

The idea is to define the service with the tests in this PR and then the integration in the next one.

cbernat:
👍

PR #1459, opened 2021-07-01
fix(Jenkinsfile): tag name was empty so now we use the function that …

cbernat:
### Added the same approach as in hello microservice

Modified the function for tag detection like in [hello microservice](https://github.com/FromDoppler/hello-microservice/blob/c6151e852de2c1bd2600546339b5b0ab2d949a1a/Jenkinsfile#L105-L107). I will test this to see if it works. 

**Context**

This seems to be not working from a very long time, `${TAG_NAME}` variable is empty in jenkins file, It was like this in previous job webapp-next and now the issue is in the new job also.

image

image

@ms-darianbenito @andresmoschini could you take a look?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -135,5 +135,5 @@ def boolean isVersionTag(String tag) {
 // https://stackoverflow.com/questions/56030364/buildingtag-always-returns-false
 // workaround https://issues.jenkins-ci.org/browse/JENKINS-55987
 def String readCurrentTag() {
-    return sh(returnStdout: true, script: "git describe --tags --match v?*.?*.?* --abbrev=0 --exact-match || echo ''").trim()
+    return sh(returnStdout: true, script: 'echo ${TAG_NAME}').trim()
 }


Jenkinsfile
```diff
@@ -47,6 +47,9 @@ pipeline {
             }
         }
         stage('Build final version images') {
+            environment {
+                TAG_VERSION = readCurrentTag();
```
andresmoschini, line 51: I think that it is fine, but the current implementation of `readCurrectTag` has some issues.

Please, review this one: https://github.com/FromDoppler/hello-microservice/blob/c6151e852de2c1bd2600546339b5b0ab2d949a1a/Jenkinsfile#L105-L107

In https://stackoverflow.com/questions/56030364/buildingtag-always-returns-false you can get more context.

andresmoschini: Haha, in my implementation I am using `TAG_NAME` 😛

Are you sure that it is empty?

cbernat: You can check in the log in here http://docker.fromdoppler.com:8080/job/FromDoppler/job/doppler-webapp/job/master/ in the steps that use `TAG_NAME` 
image

cbernat: And in the screen captures in the description of this PR I added the log for pipeline where you see the `TAG_NAME` is empty.
image

andresmoschini: En realidad, eso no debería ejecutarse, porque es master, no un tag.

andresmoschini: `master` no debería generar _final versions_, solo los _tags_ deben hacerlo.

andresmoschini: Es lo que te comentaba en el chat: https://makingsense.slack.com/archives/CPU4QDR6V/p1625226387003400?thread_ts=1625167475.001300&cid=CPU4QDR6V

cbernat: I've added the code as in hello microservice, the idea is test this to see if it works this way.

PR #1460, opened 2021-07-02
[DO-64] - Remove all push related code

ms-darianbenito:
Related to [DO-64](https://makingsense.atlassian.net/browse/DO-64)

We will create a dedicated push server for each environment.


cbernat:
Are you sure you wish to remove this form here? perhaps in the future we will need to add push in Doppler right?.

.env
```diff
@@ -14,6 +14,5 @@ REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 NODE_ICU_DATA=node_modules/full-icu
 REACT_APP_ROUTER=browser
 REACT_APP_DOPPLER_UI_LIBRARY_VERSION = v3.89.1
-REACT_APP_DATA_HUB_URL= https://hub.fromdoppler.com
```
cbernat, line 17: Please do not remove this since we use it for datahub. (to track)

public/index.html
```diff
@@ -52,11 +52,6 @@
       })(window,document,'script','dataLayer','GTM-TMMV2DF');
     </script>
     <!-- End Google Tag Manager -->
-    <script
```
cbernat, line 55: Please leave this as it is. We also use this to track, this is not related only to push.

ms-darianbenito: Good catch! Thanks!

ms-darianbenito:
> Are you sure you wish to remove this form here? perhaps in the future we will need to add push in Doppler right?.

Ceci, I will talk with @hrocca about that. Thanks for the ask!

hrocca:
> > Are you sure you wish to remove this form here? perhaps in the future we will need to add push in Doppler right?.
> 
> Ceci, I will talk with @hrocca about that. Thanks for the ask!

if necessary, we could add it again, but now we have environments for specific subscription, including production. (pushint, pushqa and pushprod in fromdoppler.net)

public/index.html
ms-darianbenito: Done!

.env
ms-darianbenito: Done!

cbernat:
👍

PR #1480, opened 2021-07-14
feat(big-query): add bigquery client and getEmailsData method

kapoc3:
Add client for big query with getEmailsData method for bigquery authorization page fetch data

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -6,6 +6,7 @@ REACT_APP_USE_DOPPLER_LEGACY_LOGIN=false
 REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 REACT_APP_SHOPIFY_URL= https://sfy.fromdoppler.com:4433
+REACT_APP_BIGQUERY_URL= https://apisint.fromdoppler.net/big-query
 REACT_APP_DOPPLER_API_URL= http://newapiqa.fromdoppler.net
 REACT_APP_REPORTS_URL= https://reportsint.fromdoppler.net:400
 REACT_APP_OFFLINE=false
--- a/.env.int
+++ b/.env.int
@@ -6,6 +6,7 @@ REACT_APP_USE_DOPPLER_LEGACY_LOGIN=false
 REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 REACT_APP_SHOPIFY_URL= https://sfy.fromdoppler.com
+REACT_APP_BIGQUERY_URL= https://apisint.fromdoppler.net/big-query
 REACT_APP_DOPPLER_API_URL= https://newapiint.fromdoppler.net
 REACT_APP_REPORTS_URL= https://reportsint.fromdoppler.net:400
 REACT_APP_OFFLINE= false
--- a/.env.production
+++ b/.env.production
@@ -6,6 +6,7 @@ REACT_APP_USE_DOPPLER_LEGACY_LOGIN=false
 REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 REACT_APP_SHOPIFY_URL= https://sfy.fromdoppler.com
+REACT_APP_BIGQUERY_URL= https://apis.fromdoppler.com/big-query
 REACT_APP_DOPPLER_API_URL= https://restapi.fromdoppler.com
 REACT_APP_REPORTS_URL= https://reports2.fromdoppler.com
 REACT_APP_OFFLINE= false
--- a/.env.qa
+++ b/.env.qa
@@ -6,6 +6,7 @@ REACT_APP_USE_DOPPLER_LEGACY_LOGIN=false
 REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 REACT_APP_SHOPIFY_URL= https://sfy.fromdoppler.com
+REACT_APP_BIGQUERY_URL= https://apisqa.fromdoppler.net/big-query
 REACT_APP_DOPPLER_API_URL= https://newapiqa.fromdoppler.net
 REACT_APP_REPORTS_URL= https://reportsqa.fromdoppler.net:400
 REACT_APP_OFFLINE= false
--- a/src/components/Integrations/BigQuery/ControlPanel/AuthorizationForm.js
+++ b/src/components/Integrations/BigQuery/ControlPanel/AuthorizationForm.js
@@ -50,7 +50,7 @@ export const AuthorizationForm = ({ emails }) => {
       validateOnBlur={false}
     >
       {({ values, validateField }) => (
-        <Form>
+        <Form aria-label="form">
           <ul className="field-group">
             <FieldArray
               name="emails"
--- a/src/components/Integrations/BigQuery/ControlPanel/AuthorizationPage.js
+++ b/src/components/Integrations/BigQuery/ControlPanel/AuthorizationPage.js
@@ -1,21 +1,23 @@
 import React, { useState, useEffect } from 'react';
 import { Loading } from '../../../Loading/Loading';
-import useTimeout from '../../../../hooks/useTimeout';
 import { AuthorizationForm } from './AuthorizationForm';
+import { InjectAppServices } from '../../../../services/pure-di';
 
-export const AuthorizationPage = () => {
-  const [data, setData] = useState([]);
-  const [loading, setLoading] = useState(false);
-  const createTimeout = useTimeout();
+const AuthorizationLayout = ({ dependencies: { bigQueryClient } }) => {
+  const [data, setData] = useState(['']);
+  const [loading, setLoading] = useState(true);
 
   useEffect(() => {
-    const fetchInfo = () => ['email1@gmail.com', 'email2@gmail.com'];
-    setLoading(true);
-    createTimeout(() => {
-      setData(fetchInfo());
+    const fetchData = async () => {
+      const result = await bigQueryClient.getEmailsData();
+      if (result.success) {
+        setData(result.value.emails);
+      }
       setLoading(false);
-    }, 2000);
-  }, [createTimeout]);
+    };
+
+    fetchData();
+  }, [bigQueryClient]);
 
   if (loading) {
     return <Loading page></Loading>;
@@ -30,3 +32,5 @@ export const AuthorizationPage = () => {
     </div>
   );
 };
+
+export const AuthorizationPage = InjectAppServices(AuthorizationLayout);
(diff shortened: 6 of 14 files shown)


.env.development
```diff
@@ -6,6 +6,7 @@ REACT_APP_USE_DOPPLER_LEGACY_LOGIN=false
 REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 REACT_APP_SHOPIFY_URL= https://sfy.fromdoppler.com:4433
+REACT_APP_BIGQUERY_URL= https://bqy.fromdoppler.com:4433
```
andresmoschini, line 9:
```suggestion
REACT_APP_BIGQUERY_URL= https://apisint.fromdoppler.net/bigquery
```

.env.int
```diff
@@ -6,6 +6,7 @@ REACT_APP_USE_DOPPLER_LEGACY_LOGIN=false
 REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 REACT_APP_SHOPIFY_URL= https://sfy.fromdoppler.com
+REACT_APP_BIGQUERY_URL= https://bqy.fromdoppler.com:4433
```
andresmoschini, line 9:
```suggestion
REACT_APP_BIGQUERY_URL= https://apisint.fromdoppler.net/big-query
```

.env.production
```diff
@@ -6,6 +6,7 @@ REACT_APP_USE_DOPPLER_LEGACY_LOGIN=false
 REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 REACT_APP_SHOPIFY_URL= https://sfy.fromdoppler.com
+REACT_APP_BIGQUERY_URL= https://bqy.fromdoppler.com:4433
```
andresmoschini, line 9:
```suggestion
REACT_APP_BIGQUERY_URL= https://apis.fromdoppler.com/big-query
```

.env.qa
```diff
@@ -6,6 +6,7 @@ REACT_APP_USE_DOPPLER_LEGACY_LOGIN=false
 REACT_APP_USE_DOPPLER_LEGACY_SIGNUP=false
 REACT_APP_USE_DOPPLER_LEGACY_FORGOTPASSWORD=false
 REACT_APP_SHOPIFY_URL= https://sfy.fromdoppler.com
+REACT_APP_BIGQUERY_URL= https://bqy.fromdoppler.com:4433
```
andresmoschini, line 9:
```suggestion
REACT_APP_BIGQUERY_URL= https://apisqa.fromdoppler.net/big-query
```

src/components/Integrations/BigQuery/ControlPanel/AuthorizationForm.js
```diff
@@ -7,7 +7,7 @@ export const AuthorizationForm = ({ emails }) => {
   const [errorMessage, setErrorMessage] = useState('');
   const intl = useIntl();
   const _ = (id, values) => intl.formatMessage({ id: id }, values);
-  const initialValues = [...emails, ''];
+  const initialValues = [...emails, ''] || [''];
```
andresmoschini, line 10: Esto está raro, en que casos `[...emails, '']` podría ser _falsy_?

kapoc3: si, esto era parte de una prueba, se elimina el opcional vacio, por que siempre se carga el valor desde el fetch inicial en el caso de no existir datos ya se envia el array vacio

.env.development
kapoc3: change accepted and applied

.env.int
kapoc3: change accepted and applied

.env.production
kapoc3: change accepted and applied

.env.qa
kapoc3: change accepted and applied

src/services/generic-client.ts
```diff
@@ -0,0 +1,44 @@
+import { AxiosInstance, AxiosStatic } from 'axios';
+import { AppSession } from './app-session';
+import { RefObject } from 'react';
+
+interface ApiConnectionData {
+  jwtToken: string;
+}
+
+export class GenericClient {
```
andresmoschini, line 9: Good idea @kapoc3, it is nice! 

In my opinion, I think that this kind of code refactoring should be in a different PR, but I think that it is acceptable.

@cbernat @jcamposmk what do you think?

BTW, what do you think about naming it _HttpGenericClient_ or _HttpBaseClient_?

```suggestion
export class HttpBaseClient {
```

src/services/big-query-client.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  private mapEmails(response: any): EmailList {
+    return {
+      emails: response.emails,
+    };
+  }
+  public async getEmailsData(idUser: String): Promise<ResultWithoutExpectedErrors<EmailList>> {
+    try {
+      const { jwtToken } = super.getApiConnectionData();
+      const response = await this.axios.request({
```
andresmoschini, line 24: Why do you use `super` for using `getApiConnectionData` and `this` for using `axios`?

Would it work?

```suggestion
      const { jwtToken } = this.getApiConnectionData();
      const response = await this.axios.request({
```

src/services/big-query-client.ts
```diff
@@ -0,0 +1,41 @@
...
+        method: 'GET',
+        url: `/big-query/emails`,
```
andresmoschini, line 26: The URL should include the accountname (email) because it helps a lot in the security implementation of our APIs:

```suggestion
      const { email, jwtToken } = super.getApiConnectionData();
      const response = await this.axios.request({
        method: 'GET',
        url: `/big-query/{email}/emails`,
```

andresmoschini, line 26: What do you think about a more descriptive name for this URL?

For example: 

```suggestion
        url: `/big-query/allowed-emails`,
```

src/services/big-query-client.ts
```diff
@@ -0,0 +1,41 @@
...
+        params: { idUser: idUser },
+        headers: { Authorization: `token ${jwtToken}` },
+      });
+
+      return this.responseCache.getCachedOrMap(this.getEmailsData, response, (r) => {
+        return r.data && r.data.length
+          ? { success: true, value: r.data.map(this.mapEmails) }
+          : { success: true, value: [''] };
```
andresmoschini, line 34: I am not sure if using `responseCache` is useful here.

The idea behind this is to use the same object's reference when the response is the same.

It will work, but I think that it is not necessary here. I have no a strong opinion about it.

src/services/generic-client.ts
andresmoschini: On the other hand, in general, I prefer composition over inheritance, is there a way of doing this using composition?

src/services/big-query-client.ts
kapoc3: in this case is because the method implementation exist on the father class, with this it works but it is not readable that the method is implemented in the parent...

andresmoschini: But `this` allows to override it, and it is desirable, right?

src/services/big-query-client.ts
kapoc3: all apis methods must always have that url param?

kapoc3: change accepted and applied

andresmoschini: if the result is related to data of an specific account, yes.

We can talk about the reason.

src/services/big-query-client.ts
kapoc3: the cache was removed

src/services/generic-client.ts
kapoc3: i can change inheritance for composition way, do you prefeer composition?

andresmoschini: In general, I prefer composition, but in this case, I am not sure if composition applies. So, if you prefer inheritance, I am fine.

src/App.js
```diff
@@ -28,7 +28,7 @@ import InvoicesList from './components/InvoicesList/InvoicesList';
 import ExclusiveForm from './components/DopplerPlus/ExclusiveForm';
 import { ContactPolicy } from './components/ContactPolicy/ContactPolicy';
 import { Promotional } from './components/Integrations/BigQuery/Promotional/Promotional';
-import { AuthorizationPage } from './components/Integrations/BigQuery/ControlPanel/AuthorizationPage';
+import AuthorizationPage from './components/Integrations/BigQuery/ControlPanel/AuthorizationPage';
```
andresmoschini, line 31: I think that we are trying to avoid usage of _default exports_, see the rationale here: https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/

andresmoschini:
In my opinion, it is fine, but I would like to see @cbernat or @jcamposmk review here.

src/App.js
kapoc3: when using the dependency injector is the way to export the class in the solution, i can make this change, but it would not be consistent with the rest of the components of the solution, do you agree with this change

Posible change 
App.js
imagen

new way to export 

imagen


do you agree with this change ?

src/services/big-query-client.ts
kapoc3: suggested accepted and applied

src/services/big-query-client.ts
kapoc3: the email was incorporated in the query string

src/App.js
andresmoschini: Oh, I understand. @cbernat @jcamposmk what is your opinion?

src/services/big-query-client.ts
andresmoschini: Thanks!

Just a minor correction, you have incorporated the email in the path:

```
          userinfo       host      port
          ┌──┴───┐ ┌──────┴──────┐ ┌┴┐
  https://john.doe@www.example.com:123/forum/questions/?tag=networking&order=newest#top
  └─┬─┘   └───────────┬──────────────┘└───────┬───────┘ └───────────┬─────────────┘ └┬┘
  scheme          authority                  path                 query           fragment
```

src/App.js
cbernat: Why not do this?
```
export const AuthorizationPage = InjectAppServices ( ( {dependencies: bigQueryClient .... 
```

cbernat:
👍

src/components/Integrations/BigQuery/ControlPanel/AuthorizationPage.js
```diff
@@ -1,21 +1,24 @@
 import React, { useState, useEffect } from 'react';
 import { Loading } from '../../../Loading/Loading';
-import useTimeout from '../../../../hooks/useTimeout';
 import { AuthorizationForm } from './AuthorizationForm';
+import { InjectAppServices } from '../../../../services/pure-di';
 
-export const AuthorizationPage = () => {
-  const [data, setData] = useState([]);
+const AuthorizationPage = ({ dependencies: { bigQueryClient } }) => {
```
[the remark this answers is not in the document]
kapoc3: Solved !!

jcamposmk:
In my opinion it's fine, the tests can be done in a better way without using timeout. Tomorrow I meet with Cristian

PR #1481, opened 2021-07-14
DW-1019-fix-client-manager-menu

silsanchez8:
## Modifications:

* fix(HeaderUserMenu): fix client manager user menu:
image

And if there is a user of client manager :
image

### DIFF
--- a/src/components/Header/HeaderUserMenu/HeaderUserMenu.js
+++ b/src/components/Header/HeaderUserMenu/HeaderUserMenu.js
@@ -34,76 +34,103 @@ const HeaderUserMenu = ({ user }) => {
         </header>
         <div className="user-plan--container">
           <div className="user-plan--type">
-            {user.plan.isSubscribers || user.plan.isMonthlyByEmail ? (
+            {!user.hasClientManager && (user.plan.isSubscribers || user.plan.isMonthlyByEmail) ? (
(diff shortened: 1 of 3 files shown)


src/components/Header/HeaderUserMenu/HeaderUserMenu.js
```diff
@@ -34,76 +34,111 @@ const HeaderUserMenu = ({ user }) => {
...
+                {user.plan.isSubscribers || user.plan.isMonthlyByEmail ? (
+                  <p className="user-plan--monthly-text">
+                    <strong>{user.plan.planName}</strong> ({user.plan.maxSubscribers}{' '}
+                    {user.plan.itemDescription})
+                  </p>
+                ) : (
+                  <p className="user-plan--monthly-text">{_('header.plan_prepaid')}</p>
+                )}
+                {user.plan.buttonUrl && user.plan.pendingFreeUpgrade ? (
+                  <button onClick={() => toggleModal(true)} className="user-plan">
```
jcamposmk, line 48: este es el mismo boton que el de abajo ?

silsanchez8: Junior, that's right, it's the same button, but this code exists since two years. @cbernat  What do you think about merge this two conditionals?: 
image

PR #1483, opened 2021-07-14
feat(utils): add getFormInitialValues utility

jcamposmk:
### **Add getFormInitialValues to utilities**
task: https://makingsense.atlassian.net/browse/DW-1038

Add utility getFormInitialValues to prepare initial form values

**tests:** 
image

### DIFF
--- a/src/utils.test.js
+++ b/src/utils.test.js
@@ -11,13 +11,14 @@ import {
   thousandSeparatorNumber,
   compactNumber,
   orderPlanTypes,
+  getFormInitialValues,
 } from './utils';
 import { renderHook } from '@testing-library/react-hooks';
 import queryString from 'query-string';
 
 describe('utils', () => {
   describe('orderPlanTypes', () => {
-    it('should sort plan types when all first plans are', () => {
+    it('should sort plan types when all first plans are present', () => {
       // Arrange
       const planTypes = ['demo', 'prepaid', 'free', 'monthly-deliveries', 'subscribers'];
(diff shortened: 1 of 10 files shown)


src/utils.test.js
```diff
@@ -565,40 +566,63 @@ describe('utils', () => {
...
-    it(`should return ${expectedFormatted} when the number is ${rawNumber} and language is "${language}"`, () => {
-      // Act
-      const result = thousandSeparatorNumber(language, rawNumber);
+  describe('thousandSeparatorNumber function', () => {
+    it('should return correct intl format for numbers in ESP', () => {
+      // Assert
+      expect(thousandSeparatorNumber('es', 100)).toBe('100');
+      expect(thousandSeparatorNumber('es', 1000)).toBe('1.000');
+      expect(thousandSeparatorNumber('es', 10000)).toBe('10.000');
+    });
```
andresmoschini, line 575: These changes seems to not be related to `getFormInitialValues`, what I am missing?

andresmoschini:
Great!

Very minor:

In my opinion, it is not a feature, because it does not affect the final user.

In my opinion in a part of a refactor. By the way, what do you think about completing the refactor in this PR replacing the duplicated code with this new function?

src/utils.test.js
jcamposmk: I don't understand why it shows them as that code was altered. But it happened when I ran the prettier. these changes are not related.

jcamposmk:
it can be done, in this way it would give a context of how to use it.

PR #1494, opened 2021-07-16
feat(BigQuery): add saveEmailsData to big query http client

kapoc3:
Method for save data emails added on http bigquery client.

### DIFF
--- a/src/services/big-query-client.ts
+++ b/src/services/big-query-client.ts
@@ -1,12 +1,14 @@
-import { ResultWithoutExpectedErrors } from '../doppler-types';
 import { HttpBaseClient } from './http-base-client';
+import { ResultWithoutExpectedErrors, EmptyResultWithoutExpectedErrors } from '../doppler-types';
 
+export type SaveEmailsResult = EmptyResultWithoutExpectedErrors;
 export interface EmailList {
   emails: String[];
 }
 
 export interface BigQueryClient {
   getEmailsData(): Promise<ResultWithoutExpectedErrors<EmailList>>;
+  saveEmailsData(payload: EmailList): Promise<SaveEmailsResult>;
 }
 
 export class HttpBigQueryClient extends HttpBaseClient implements BigQueryClient {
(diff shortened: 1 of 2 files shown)


src/services/big-query-client.ts
```diff
@@ -1,12 +1,15 @@
-import { ResultWithoutExpectedErrors } from '../doppler-types';
 import { HttpBaseClient } from './http-base-client';
+import { ResultWithoutExpectedErrors, EmptyResultWithoutExpectedErrors } from '../doppler-types';
 
+export type SaveEmailsResult = EmptyResultWithoutExpectedErrors;
 export interface EmailList {
+  idUser?: String;
```
andresmoschini, line 6: ¿Por qué agregás este campo?

kapoc3: olvide removerlo del cliente, solucionado!!

src/services/big-query-client.ts
```diff
@@ -35,4 +37,18 @@ export class HttpBigQueryClient extends HttpBaseClient implements BigQueryClient
       return { success: false, error: error };
     }
   }
+
+  public async saveEmailsData(payload: EmailList): Promise<SaveEmailsResult> {
+    const { jwtToken, email } = this.getApiConnectionData(this.clientName);
+    const response = await this.axios.post(`/big-query/${email}/save-emails`, {
```
andresmoschini, line 43: In my opinion, it should be a `PUT` against the same URL:

```suggestion
    const response = await this.axios.post(`/big-query/${email}/allowed-emails`, {
```

In my opinion, it is the same resource that you obtain with the `GET`, also, it is a kind of update, an idempotent operation, so `PUT` fits much better.

kapoc3: suggest accepted and applied!!!

cbernat:
@kapoc3 it would be nice making use of this method in what you have already in the view right?

kapoc3:
> 
> 
> @kapoc3 it would be nice making use of this method in what you have already in the view right?

@cbernat when the junior's component be integrate with the template, i can generate a new pr with the integration of button save

cbernat:
> > @kapoc3 it would be nice making use of this method in what you have already in the view right?
> 
> @cbernat when the junior's component be integrate with the template, i can generate a new pr with the integration of button save

But you have a small implementation of your own that can be tested right?

cbernat:
Can you add all pages related to big query here? https://app.fromdoppler.com/new-features we have this page for pending development, to access them easily in the meantime.

cbernat:
> > > @kapoc3 it would be nice making use of this method in what you have already in the view right?
> > 
> > 
> > @cbernat when the junior's component be integrate with the template, i can generate a new pr with the integration of button save
> 
> But you have a small implementation of your own that can be tested right?

I've seen that big query page has not submit button yet 🤦 So ignore this comment.

cbernat:
👍 🚀

PR #1496, opened 2021-07-19
feat: disable save button while updating or no changes

ramniel:

### DIFF
--- a/src/components/ContactPolicy/ContactPolicy.js
+++ b/src/components/ContactPolicy/ContactPolicy.js
@@ -4,7 +4,7 @@ import { FormattedMessage, useIntl } from 'react-intl';
 import HeaderSection from '../shared/HeaderSection/HeaderSection';
 import { Breadcrumb, BreadcrumbItem } from '../shared/Breadcrumb/Breadcrumb';
 import { FormattedMessageMarkdown } from '../../i18n/FormattedMessageMarkdown';
-import { FieldGroup, NumberField, SubmitButton, SwitchField } from '../form-helpers/form-helpers';
+import { FieldGroup, NumberField, SwitchField } from '../form-helpers/form-helpers';
 import { Form, Formik } from 'formik';
 import { InjectAppServices } from '../../services/pure-di';
 import { Loading } from '../Loading/Loading';
@@ -42,14 +42,17 @@ export const ContactPolicy = InjectAppServices(
       }
     }, [isContactPolicyEnabled, dopplerContactPolicyApiClient, appSessionRef]);
 
-    const submitContactPolicyForm = async (values, { setSubmitting }) => {
+    const submitContactPolicyForm = async (values, { setSubmitting, resetForm }) => {
       setFormSubmitted(false);
       try {
         const { success } = await dopplerContactPolicyApiClient.updateAccountSettings(values);
-        if (!success) {
+        if (success) {
+          setFormSubmitted(true);
+          resetForm({ values });
+          setSettings(values);
+        } else {
           console.log('Error updating account settings');
         }
-        setFormSubmitted(success);
       } finally {
         setSubmitting(false);
       }
@@ -86,8 +89,10 @@ export const ContactPolicy = InjectAppServices(
                       ...getFormInitialValues(fieldNames),
                       ...settings,
                     }}
+                    validateOnBlur={false}
+                    enableReinitialize={true}
                   >
-                    {({ values }) => (
+                    {({ values, isSubmitting, isValid, dirty }) => (
                       <Form className="dp-contact-policy-form">
                         <fieldset>
                           <legend>{_('contact_policy.title')}</legend>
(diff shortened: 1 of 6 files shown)


src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -43,10 +44,12 @@ export const ContactPolicy = InjectAppServices(
     }, [isContactPolicyEnabled, dopplerContactPolicyApiClient, appSessionRef]);
 
     const submitContactPolicyForm = async (values, { setSubmitting }) => {
+      setSubmitDisabled(true);
       setFormSubmitted(false);
       try {
         const { success } = await dopplerContactPolicyApiClient.updateAccountSettings(values);
         if (!success) {
+          setSubmitDisabled(false);
```
jcamposmk, line 52: should this be in the `finally` block?

src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -151,7 +169,16 @@ export const ContactPolicy = InjectAppServices(
                               </a>
 
                               <span className="align-button m-l-24">
-                                <SubmitButton>{_('common.save')}</SubmitButton>
+                                <button
+                                  type="submit"
+                                  disabled={submitDisabled}
```
jcamposmk, line 174: Have you tried using isValid? isValid tells you if the form is valid, it is based on the validations that you pass it

src/components/ContactPolicy/ContactPolicy.test.js
```diff
@@ -1,6 +1,11 @@
 import React from 'react';
-import { cleanup, fireEvent, render, waitFor } from '@testing-library/react';
-import { act } from 'react-dom/test-utils';
+import {
+  cleanup,
```
jcamposmk, line 3: Just to clarify. Some testing frameworks already cleanup automatically, including jest and mocha

src/components/ContactPolicy/ContactPolicy.test.js
```diff
@@ -77,16 +82,19 @@ describe('ContactPolicy component', () => {
...
-    });
+    const switchButton = container.querySelector('input#contact-policy-switch');
+    fireEvent.click(switchButton);
+    expect(switchButton).toBeChecked();
+
+    fireEvent.click(getByText('common.save'));
+    expect(await findByText('contact_policy.success_msg'));
   });
 
   it('should disable the inputs if the switch is not active', async () => {
```
jcamposmk, line 100: The test says inputs should be disabled, but I don't see this being verified

src/components/ContactPolicy/ContactPolicy.js
ramniel: no, only enabled in case of error, otherwise it remains disabled

jcamposmk: Ok. But, then it would be necessary to reload the page again to be able to correct the error and send the form again?

ramniel (commented):
...

src/components/ContactPolicy/ContactPolicy.js
ramniel: no, the idea is to keep the latest changes

src/components/ContactPolicy/ContactPolicy.test.js
ramniel: The switch is disabled by default, therefore it is verified that the inputs are disabled

src/components/ContactPolicy/ContactPolicy.js
ramniel: Thanks. Let's try it

src/components/ContactPolicy/ContactPolicy.js
jcamposmk: How then can the user submit the form again if the button is disabled?

src/components/ContactPolicy/ContactPolicy.test.js
jcamposmk: I can see is that it is verifying that once the switch is clicked, the inputs are enabled. But you also have to make the checks that before clicking, the inputs are disabled

src/components/ContactPolicy/ContactPolicy.js
ramniel: The flow is as follows:

- By default the Save button is disabled.
- The user modifies some value in the inputs or the switch.
- The save button is enabled.
- Changes are saved (The button is disabled while saving and the button is animated)
    - If an error occurs, the button is enabled so that the user can try again.
    - If it is saved successfully, the button remains disabled until the user modifies a value.

src/components/ContactPolicy/ContactPolicy.test.js
ramniel: You mean that in this test `"shouldn't disable the inputs if the switch is active"`, right?
Yes, it would be good to include that previous check.

jcamposmk: Yes, it's better

PR #1500, opened 2021-07-20
feat: validate input fields range and show error messages

ramniel:

### DIFF
--- a/src/components/ContactPolicy/ContactPolicy.js
+++ b/src/components/ContactPolicy/ContactPolicy.js
@@ -1,11 +1,11 @@
 import React, { useEffect, useState } from 'react';
 import { Helmet } from 'react-helmet';
-import { FormattedMessage, useIntl } from 'react-intl';
+import { useIntl } from 'react-intl';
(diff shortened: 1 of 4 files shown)


src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -50,14 +75,42 @@ export const ContactPolicy = InjectAppServices(
...
 
+    const validate = (values) => {
+      setError(false);
+      setFormSubmitted(false);
+      const errors = {};
+
+      const amountIsEmpty = values.emailsAmountByInterval === '';
+      const intervalIsEmpty = values.intervalInDays === '';
+
+      if (amountIsEmpty || intervalIsEmpty) {
```
jcamposmk, line 95: Formik handles 2 types of validations: at the form-level and at the field-level. In case of form-level validations, you can pass a validation function to it. You don't need to handle these validations manually. 
doc: https://formik.org/docs/guides/validation

jcamposmk:
When you can, we connect and see it

PR #1501, opened 2021-07-20
feat(checkout): checkout summary page

cbernat:
### Upgrade for old doppler

We need to track in google tag manager when the user upgrades from free to paid succesfully. To do this, we will track a redirection to a page. In Upgrade we will do a redirect to this page hidden and then redirect as usual. 
This page in the future will be the summary for checkout process for new plans. 
This page has two ways of operation:
- When legacy parameter is present it will redirect to old doppler with redirect parameter or by default to `Campaigns/Draft`
- When legacy parameter is not present it will remain in this page. 

Pending:

- [ ] Add tests

### DIFF
--- a/src/components/ChangePlan/Checkout/CheckoutSummary.js
+++ b/src/components/ChangePlan/Checkout/CheckoutSummary.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import { extractParameter } from '../../../utils';
+import queryString from 'query-string';
+import SafeRedirect from '../../SafeRedirect';
+
+export const CheckoutSummary = ({ location }) => {
+  const redirect = extractParameter(location, queryString.parse, 'redirect');
+  const legacy = extractParameter(location, queryString.parse, 'legacy');
+
+  if (legacy) {
+    if (redirect) {
+      return <SafeRedirect to={redirect} />;
+    }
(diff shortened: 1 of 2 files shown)


src/components/ChangePlan/Checkout/CheckoutSummary.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { extractParameter } from '../../../utils';
+import queryString from 'query-string';
+import SafeRedirect from '../../SafeRedirect';
+
+export const CheckoutSummary = ({ location }) => {
```
jcamposmk, line 6: The implementation is a bit complicated for what the component does.What do you think about this:

``` 
export const CheckoutSummary = ({ location }) => {
  const redirect = extractParameter(location, queryString.parse, 'redirect');
  const legacy = extractParameter(location, queryString.parse, 'legacy');

  if(legacy) {
     if(redirect) {
       return <SafeRedirect to={redirect} />
     }
     return <SafeRedirect to="/Campaigns/Draft" />
  }
  
  return ('Normal checkout flow for new plans should go here.');
};
```

cbernat:
I will merge this without tests for now, because I need to insert this into Upgrade in old Doppler ASAP.

PR #1503, opened 2021-07-21
[DAT-500] - fix: use 'const' instead of 'var'

jhoffman-ms:
Use ```const``` instead of ```var```

**Task:** [DAT-500](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-500)


cbernat:
👍  🚀 Ready to merge.

PR #1504, opened 2021-07-21
fix: improve tests

ramniel:
With the help of @jcamposmk, consulting the documentation of the [react testing library](https://testing-library.com) and analyzing some articles about the [author](https://kentcdodds.com/blog) of the library, we found some recommendations on tests, which may be useful. We would like to know your opinions on this.

- There is no need to call the [cleanup](https://testing-library.com/docs/react-testing-library/api/#cleanup) function as it is called automatically for Jest and other test frameworks.

- The use of [screen](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen) is recommended:
The benefit of using [screen](https://testing-library.com/docs/queries/about/#screen) is you no longer need to keep the render call destructure up-to-date as you add/remove the [queries](https://testing-library.com/docs/queries/about/#using-queries) you need. You only need to type screen. and let your editor's magic autocomplete take care of the rest.

   *_It seems like the returned object from the render function could be framework specific but the screen object will be framework agnostic, making your tests more resilient and consistent across projects._ 

- Wrapping things in [`act` unnecessarily](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#wrapping-things-in-act-unnecessarily). Some queries and functions are already wrapped in `act`

- Use [`*byRole`](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-byrole-most-of-the-time) most of the time: 
There's not much you can't get with this (if you can't, it's possible your UI is inaccessible).
[Queries priority](https://testing-library.com/docs/queries/about/#priority)

- Do not use `waitFor` to wait for items that can be queried with [`find*`](https://testing-library.com/docs/dom-testing-library/api-async#findby-queries)
   (`find*` queries use `waitFor` under the hood)

- Do not put [multiple assertions](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback) in a single `waitFor` callback
By putting a single assertion, we can wait for the UI to settle to the state we want to assert on, and also fail faster if one of the assertions do end up failing.

- Use [@testing-library/user-event](https://testing-library.com/docs/ecosystem-user-event)
Is a [package](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event) that's built on top of `fireEvent`, but it provides several methods that resemble the user interactions more closely. It provides more advanced simulation of browser interactions than the built-in `fireEvent` method.

[Common mistakes](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library)
[Cheatsheet](https://testing-library.com/docs/react-testing-library/cheatsheet)

Other changes:
- The [`data-testid`](https://testing-library.com/docs/queries/bytestid) attribute was added to the `Loading` component to [obtain the component](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change#whats-with-the-data-testid-query) by this attribute and not by the name of the class. This way we make sure that the test keeps working no matter if the class name changes.
   **_In production_**: You can compile this attribute away with `babel-plugin-react-remove-properties`.

- The [aria-label](https://www.aditus.io/aria/aria-label) attribute was added to the form and the inputs fields to differentiate them by their role and [accessibility name](https://es.reactjs.org/docs/accessibility.html) 

[Using aria-label attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute)

### DIFF
--- a/src/components/ContactPolicy/ContactPolicy.test.js
+++ b/src/components/ContactPolicy/ContactPolicy.test.js
@@ -1,11 +1,6 @@
 import React from 'react';
-import {
-  fireEvent,
-  screen,
-  render,
-  waitFor,
-  waitForElementToBeRemoved,
-} from '@testing-library/react';
+import { render, screen, waitForElementToBeRemoved } from '@testing-library/react';
+import user from '@testing-library/user-event';
 import '@testing-library/jest-dom/extend-expect';
 import DopplerIntlProvider from '../../i18n/DopplerIntlProvider.double-with-ids-as-values';
 import { ContactPolicy } from './ContactPolicy';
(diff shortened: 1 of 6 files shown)


src/components/ContactPolicy/ContactPolicy.test.js
```diff
@@ -53,100 +50,316 @@ describe('ContactPolicy component', () => {
...
-    await waitFor(() => {
-      expect(container.querySelector('input#contact-policy-switch')).not.toBeChecked();
-      expect(container.querySelector('input#contact-policy-input-amount').value).toEqual(
-        `${settingsDouble().emailsAmountByInterval}`,
-      );
-      expect(container.querySelector('input#contact-policy-input-interval').value).toEqual(
-        `${settingsDouble().intervalInDays}`,
-      );
-    });
+    await waitForElementToBeRemoved(screen.getByTestId('wrapper-loading'));
```
cbernat, line 65: So instead of having to wait, you now wait on the spinner to be removed, it's clearer.

cbernat:
The changes seem reasonable, perhaps you could update this PR and merge it. And also be on the lookout for tests reviews to apply this way of doing thinks, from now on 😅

PR #1509, opened 2021-07-23
refactor: add tag cloud component to big query

jcamposmk:
### **Add tag cloud component to big query**

**scenario:** 

- the user navigate to /integrations/big-query/plus
- the user can see the authorization form for google account
- the user adds tag with enter event or click to the + button
- when the user add invalid tag (no email) shown error message
- when the user add tag valid but the tag already exist, shown error message
- when the user submit form, shown alert with sent data

image

tests: 
image


cbernat:
Test this please.

cbernat:
👍

PR #1510, opened 2021-07-23
feat: show success message and remove it after 3 seconds

ramniel:

### DIFF
--- a/src/components/ContactPolicy/ContactPolicy.js
+++ b/src/components/ContactPolicy/ContactPolicy.js
@@ -5,11 +5,12 @@ import HeaderSection from '../shared/HeaderSection/HeaderSection';
 import { Breadcrumb, BreadcrumbItem } from '../shared/Breadcrumb/Breadcrumb';
 import { FormattedMessageMarkdown } from '../../i18n/FormattedMessageMarkdown';
 import { FieldGroup, IconMessage, NumberField, SwitchField } from '../form-helpers/form-helpers';
-import { useFormikContext, Form, Formik } from 'formik';
+import { Form, Formik } from 'formik';
 import { InjectAppServices } from '../../services/pure-di';
 import { Loading } from '../Loading/Loading';
-import { getFormInitialValues } from '../../utils';
+import { getFormInitialValues, successMessageDelay } from '../../utils';
 import { Prompt } from 'react-router-dom';
+import { ShowLikeFlash } from '../shared/ShowLikeFlash/ShowLikeFlash';
 
 export const ContactPolicy = InjectAppServices(
   ({
@@ -30,8 +31,7 @@ export const ContactPolicy = InjectAppServices(
       intervalInDays: 'intervalInDays',
     };
 
-    const FieldItemMessage = () => {
-      const { errors } = useFormikContext();
+    const FieldItemMessage = ({ errors }) => {
       let message = {};
       if (errors.message) {
         message.text = errors.message;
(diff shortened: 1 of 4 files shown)


src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -42,14 +43,17 @@ export const ContactPolicy = InjectAppServices(
...
+        message.delay = 3000;
       } else {
         return null;
       }
 
       return (
-        <li className="field-item">
-          <IconMessage {...message} className="bounceIn" />
-        </li>
+        <Flash delay={message.delay}>
```
cbernat, line 52: How about renaming this component into something like this?:
`ShowLikeFlash`
`FlashInAndOut`
`FlashAnimate`
Something that indicates that the component has the responsibility to show and hide at simple sight.

src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -42,14 +43,17 @@ export const ContactPolicy = InjectAppServices(
       } else if (formSubmitted) {
         message.text = _('contact_policy.success_msg');
         message.type = 'success';
+        message.delay = 3000;
```
cbernat, line 46: It would be nice to have this delay set as a global variable for success delay somewhere, to avoid repeating this value each time we use a success message. Or have this delay included inside the `Flash` component. What do you think?

src/components/ContactPolicy/ContactPolicy.js
jcamposmk: Quizás pueda usarse un custom hook para ese componente. De tal forma que pueda ser customizado dónde se use. Algo así como lo hace https://www.npmjs.com/package/react-notifications-component

src/components/ContactPolicy/ContactPolicy.js
jcamposmk: Se pudiese crear eso de la variable global y que se utilice por defecto si no se le específica un delay custom ...

src/components/ContactPolicy/ContactPolicy.js
ramniel: ShowLikeFlash sounds good

src/components/ContactPolicy/ContactPolicy.js
ramniel: I think it would be nice to have it set as a global variable.

src/utils.ts
```diff
@@ -319,3 +319,6 @@ export const getFormInitialValues = (fieldNames: object) =>
     (accumulator, currentValue) => ({ ...accumulator, [currentValue]: '' }),
     {},
   );
+
+// Time (in milliseconds ) that the message is displayed on the screen.
+export const successMessageDelay = 3000;
```
jcamposmk, line 324: @cbernat should it be uppercase or lowercase?

jcamposmk:
@ramniel what is the `FieldItemMessage` method ? 
hooks must be called at the top level https://es.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

src/utils.ts
cbernat: We use it this way, I see it ok.

jcamposmk:
Perhaps you can build a component for success notifications in another PR. To standardize the types of notifications

jcamposmk:
:+1:

ramniel:
> Perhaps you can build a component for success notifications in another PR. To standardize the types of notifications. What do you think about this @ramniel

It is a good idea!

ramniel:
> @ramniel what is the `FieldItemMessage` method ?
> hooks must be called at the top level https://es.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

Good catch! This method is to show the error or success message when it exists. In this case we can pass the errors as arguments and not have to use the hook. Thanks!

PR #1516, opened 2021-07-26
feat: [DAT-488] Upgrade plan form - Add confirmation message

nreal:
# Background
We add a new modal when the user upgrades a new plan 
Example:
Capture2


src/components/UpgradePlanForm/ConfirmationMessage.test.js
```diff
@@ -0,0 +1,33 @@
...
+    const { getByText } = render(
+      <IntlProvider>
+        <ConfirmationMessage isSubscriber={true} amountSubscribers={1}></ConfirmationMessage>
+      </IntlProvider>,
+    );
+
+    // Assert
+    expect(getByText('upgradePlanForm.confirmation_title')).toBeInTheDocument();
+    expect(getByText('upgradePlanForm.confirmation_subtitle_contact')).toBeInTheDocument();
+  });
```
cbernat, line 32: This test is duplicated 🤔 You must've meaning to add a test for non subscriber account right?

src/components/UpgradePlanForm/UpgradePlanForm.js
```diff
@@ -78,7 +91,7 @@ const UpgradePlanForm = ({
     return errors;
   };
 
-  return !state.isLoading ? (
+  return !state.isLoading && !confirmationMessage ? (
```
cbernat, line 94: How about having this logic?
```
return state.isLoading? (<Loading>):
  (confirmationMessage? (
    <ConfirmationMessage
      isSubscriber={isSubscriber}
      amountSubscribers={amountSubscribers}
    ></ConfirmationMessage>
  ): 
  .... {/*The rest of the code here*/}
  )

src/components/UpgradePlanForm/ConfirmationMessage.test.js
nreal: Yes, I changed this test for non subscribers accounts, done

src/components/UpgradePlanForm/UpgradePlanForm.js
nreal: yes is working, done

cbernat:
Very nice @nreal ! One minor detail, could you merge the commits into one? We try to make the commit messages to be readable in the release notes.
Instead of this: 

image
 
 How about fixing the commit message to something like this:
`feat(upgrade-plan): add confirmation message for upgrade popup`

cbernat:
🚀

PR #1517, opened 2021-07-27
[DAT-496] - Change the "Province/State" field to use "Select"

jhoffman-ms:
Change the "Province/State" field to use "Select" instead of "Text"

**Before:**

image

**After:**

image

**Task:** [DAT-496](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-496)

### DIFF
--- a/src/services/static-data-client.ts
+++ b/src/services/static-data-client.ts
@@ -2,6 +2,7 @@ import { AxiosInstance, AxiosStatic } from 'axios';
 
 export interface StaticDataClient {
   getIndustriesData(language: string): Promise<any>;
+  getStatesData(country: string, language: string): Promise<any>;
 }
 
 export class HttpStaticDataClient implements StaticDataClient {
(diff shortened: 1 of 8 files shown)


jcamposmk:
:+1:

src/services/static-data-client.ts
```diff
@@ -26,4 +27,31 @@ export class HttpStaticDataClient implements StaticDataClient {
       return { success: false, error: error };
     }
   }
+
+  public async getStatesData(country: string, language: string): Promise<any> {
+    try {
+      const response = await this.axios.request({
+        method: 'GET',
+        url: this.baseUrl + `/${language === 'es' ? 'states-es-v1' : 'states-en-v1'}.json`,
+        headers: { 'Cache-Control': 'no-store,max-age=0' },
```
cbernat, line 36: As we talked this should not be neccesary right now, we will see in the future.

cbernat:
👍

src/services/static-data-client.ts
jhoffman-ms: Done

PR #1522, opened 2021-07-28
feat: add cloud tag compound field integration to contact policy

jcamposmk:
Add cloud tags to contact policy

**scenario:**

- The user navigates to /sending-preferences/contact-policy
- The user enable switch
- When user click Add List button, a fake subscriber is added

**disabled**
image
 
 **enabled**
image

**limit exceeded**
image


src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -123,6 +130,20 @@ export const ContactPolicy = InjectAppServices(
...
       return errors;
     };
+    
+    const getSubscriberToAdd = () => {
+      const subscriberId = Math.floor(Math.random() * 21);
+      return {
+        id: subscriberId,
+        [labelKey]: `subscriber${subscriberId}`,
+      };
+    };
```
cbernat, line 140: Here the this that we will be adding into the cloudTag is a List, and not a Subscriber.
```suggestion
    const getListToAdd = () => {
      const listId = Math.floor(Math.random() * 21);
      return {
        id: listId,
        [labelKey]: `List${listId}`,
      };
    };
```

cbernat: Sometimes we use List or SubscribersList to define a List.

src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -123,6 +130,20 @@ export const ContactPolicy = InjectAppServices(
...
+    
+    const selectSubscriberFromList = (addTag) => {
+      // TODO: actions needed to add to list
+      const subscriberToAdd = getSubscriberToAdd();
+      addTag(subscriberToAdd);
+    };
```
cbernat, line 146: The same here, we are adding lists and not subscribers.

cbernat:
Test this please.

src/components/ContactPolicy/ContactPolicy.test.js
```diff
@@ -31,7 +30,7 @@ describe('ContactPolicy component', () => {
     active: false,
     emailsAmountByInterval: 20,
     intervalInDays: 7,
-    excludedSubscribersLists: [],
+    excludedSubscribersLists,
```
cbernat, line 33: How about a test to simulate having one excluded list, and check the tag is created?

cbernat:
Nice! Just a small suggestion in tests if you see it fit.

src/components/ContactPolicy/ContactPolicy.test.js
cbernat: As we talked by chat this is already tested inside the component for each CloudTagField, so it's ok as it is.

cbernat:
👍 🚀

PR #1533, opened 2021-08-05
DW-1080-Fix-decimals

silsanchez8:
## Modifications:

* fix(PlanAgreement): Add function to fix unit price decimals

image

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -295,6 +295,8 @@ export const thousandSeparatorNumber = (lang: string, value: number) =>
     minimumFractionDigits: value % 1 ? 2 : 0,
   }).format(value);
 
+export const unitPriceDecimals = (value: number) => value.toFixed(5);
+
 export const compactNumber = new Intl.NumberFormat('en', {
   //@ts-ignore
   notation: 'compact',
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -295,6 +295,8 @@ export const thousandSeparatorNumber = (lang: string, value: number) =>
     minimumFractionDigits: value % 1 ? 2 : 0,
   }).format(value);
 
+export const unitPriceDecimals = (value: number) => value.toFixed(5);
```
jcamposmk, line 298: an additional parameter could be added, indicating the number of decimal to format ... 
If you try to format this number `1.3456789.toFixed(5)`, it returns this: `1.34568`. Does a rounding. That would be fine? or should it return `1.34567`?

silsanchez8: @jcamposmk  I did it using 5 decimals because all unit prices have these quantity. It's ok if rounds a number with 5 decimals, like `1.34568`

cbernat: Did you check the values against the ones in sites or old Doppler just to be sure?

silsanchez8: Yes, I checked it comparing with old Doppler. Also, I have to check sites, but when I developed that section I've checked old Doppler too. :)

PR #1535, opened 2021-08-05
feat: add list selection page component

ramniel:

### DIFF
--- a/src/components/ContactPolicy/ContactPolicy.js
+++ b/src/components/ContactPolicy/ContactPolicy.js
@@ -17,8 +17,10 @@ import { Loading } from '../Loading/Loading';
 import { getFormInitialValues, successMessageDelay } from '../../utils';
 import { Prompt } from 'react-router-dom';
 import { ShowLikeFlash } from '../shared/ShowLikeFlash/ShowLikeFlash';
-import { CloudTagCompoundField } from '../form-helpers/CloudTagCompoundField';
 import { Promotional } from '../shared/Promotional/Promotional';
+import { CloudTagCompoundField } from '../form-helpers/CloudTagCompoundField';
+import Modal from '../Modal/Modal';
+import { SubscriberListSelector } from '../SubscriberListSelector/SubscriberListSelector';
 
 const maxItems = 10;
 const limitExceededMessageKey = 'contact_policy.tooltip_max_limit_exceeded';
@@ -40,6 +42,7 @@ export const ContactPolicy = InjectAppServices(
     const [settings, setSettings] = useState({});
     const [formSubmitted, setFormSubmitted] = useState(false);
     const [error, setError] = useState(false);
+    const [modalIsOpen, setModalIsOpen] = useState(false);
     const intl = useIntl();
     const _ = (id, values) => intl.formatMessage({ id: id }, values);
 
@@ -144,20 +147,7 @@ export const ContactPolicy = InjectAppServices(
       setFormSubmitted(false);
     };
 
-    const getSubscribersListToAdd = () => {
-      const subscribersListId = Math.floor(Math.random() * 21);
-      return {
-        id: subscribersListId,
-        [labelKey]: `subscribers-list${subscribersListId}`,
-      };
-    };
-
-    const selectSubscribersList = (addList) => {
-      hideMessage();
-      // TODO: actions needed to add to list
-      const subscribersListToAdd = getSubscribersListToAdd();
-      addList(subscribersListToAdd);
-    };
+    const toggleModal = (isOpen) => setModalIsOpen(isOpen);
 
     if (loading) {
       return <Loading page />;
@@ -258,7 +248,7 @@ export const ContactPolicy = InjectAppServices(
                               messageKeys={{
                                 tagLimitExceeded: limitExceededMessageKey,
                               }}
-                              render={(addList) => {
+                              render={() => {
                                 const maxLimitReached =
                                   values[fieldNames.excludedSubscribersLists].length === maxItems;
                                 return (
@@ -271,10 +261,10 @@ export const ContactPolicy = InjectAppServices(
                                       className="dp-button dp-add-list"
                                       disabled={!values[fieldNames.active] || maxLimitReached}
                                       aria-label="add tag"
-                                      onClick={() => selectSubscribersList(addList)}
+                                      onClick={() => toggleModal(true)}
                                     >
                                       <span>+</span>
-                                      {_('contact_policy.add_list')}
+                                      {_('contact_policy.select_lists')}
                                     </button>
                                   </WrapInTooltip>
                                 );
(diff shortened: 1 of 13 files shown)


src/components/ListSelectionPage/ListSelectionPage.js
```diff
@@ -0,0 +1,88 @@
...
+            <img
+              src={_('common.ui_library_image', { imageUrl: 'check-grey.svg' })}
+              alt="check list"
+            />
+          </th>
+          <th>Nombre de la lista</th>
+          <th>suscriptores</th>
+        </tr>
+      </thead>
+      <tbody>{rows}</tbody>
```
jcamposmk, line 57: It seems better to do it directly here: 
```
lists.map((list) => (
    <ListRow list={list} key={list.id} />);
)
```

I don't think there is any performance improvement, but it helps us separate presentation from events and how the data is prepared for display. Also, you avoid creating a variable

src/components/ListSelectionPage/ListSelectionPage.js
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
+import { useIntl } from 'react-intl';
+import { InjectAppServices } from '../../services/pure-di';
+import { Loading } from '../Loading/Loading';
+import { SubscriberListState } from '../../services/shopify-client';
+import { Promotional } from '../ContactPolicy/Promotional';
+import InfiniteScroll from 'react-infinite-scroll-component';
+import { useFetch } from '../../hooks/useFetch';
+
+const listsPerPage = 20;
```
jcamposmk, line 10: Normally I declare the variables in Uppercase and outside the component, when their value is known before the execution of the app. Probably later, you want to standardize so that the tables to select lists work with this same value. It may be a good idea to raise the variable to a level where it can have a greater meaning and change its name.
I think in this article they talk a little about that https://www.codigofuente.org/variables-y-constantes-en-javascript/

src/components/ListSelectionPage/ListSelectionPage.js
```diff
@@ -0,0 +1,88 @@
...
+      </thead>
+      <tbody>{rows}</tbody>
+    </table>
+  );
+};
+
+export const ListSelectionPage = InjectAppServices(({ dependencies: { dopplerApiClient } }) => {
+  const fetchData = (page) =>
+    dopplerApiClient.getSubscribersLists(page, listsPerPage, SubscriberListState.ready);
+  const { loading, items, hasMoreItems, loadMoreItems } = useFetch(fetchData);
```
jcamposmk, line 65: I think this hook is good. But I don't think useFetch is a proper name. It is very generic for the specialization it has. 
A name that is made with the page and infinite scroll would be better. Wich will be ?

src/components/ListSelectionPage/ListSelectionPage.js
```diff
@@ -0,0 +1,88 @@
...
+          <th>Nombre de la lista</th>
+          <th>suscriptores</th>
+        </tr>
+      </thead>
+      <tbody>{rows}</tbody>
+    </table>
+  );
+};
+
+export const ListSelectionPage = InjectAppServices(({ dependencies: { dopplerApiClient } }) => {
```
jcamposmk, line 62: is this a new view or just a component to be used in another view? the page suffix doesn't know if it's okay here. In any case, ListSelectionPage is a very general name for the list of subscribers

jcamposmk:
Every time you upload a PR, you also have to upload the tests. Especially since normally we do not download the branch and test it locally, so having the tests provides us with documentation on how what you are creating behaves in certain cases.

src/components/ListSelectionPage/ListSelectionPage.js
ramniel: How about 'useInfinitePaged'?

src/components/ListSelectionPage/ListSelectionPage.js
ramniel: It is a component to be used in another view. We can name it 'SubscriberListSelector'.

src/components/ListSelectionPage/ListSelectionPage.js
ramniel: Nice article, thanks. It would be nice to share it with the team. The idea here is to follow the `camelCase` format that is used to declare global variables in utils: `export const successMessageDelay = 3000;`

src/components/ListSelectionPage/ListSelectionPage.js
jcamposmk: I would think that you should call `SubscriberLists` and receive a prop that indicates the opening mode: if it is selectable or not. What do you think ? because this list could be used in another view.

jcamposmk:
For a next time, it might be better to divide the task into 2 parts: `1. the creation of the service` and `2. the code that uses the service with everything else`. It becomes easier for both review and development

src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -19,7 +19,10 @@ import { Prompt } from 'react-router-dom';
 import { ShowLikeFlash } from '../shared/ShowLikeFlash/ShowLikeFlash';
 import { Promotional } from './Promotional';
 import { CloudTagCompoundField } from '../form-helpers/CloudTagCompoundField';
+import Modal from '../Modal/Modal';
+import { SubscriberListSelector } from '../SubscriberListSelector/SubscriberListSelector';
 
+const listsPerPage = 20;
```
jcamposmk, line 25: @andresmoschini esta variable debería estar en mayúscula o en minúscula? 
Por lo general trato que las variables cuyo contenido no cambia y de la cual se conoce su valor antes de la ejecución de la APP y su uso es global, estén en mayúsculas y con snake case: LIST_PER_PAGE = 20; tienes alguna opinion fuerte con respecto a esto ?

andresmoschini: No recuerdo como lo estamos haciendo en WebApp. Una alternativa sería ponerlo en un archivo de configuración, otra podría ser no pasar ese parámetro (ya que la API tiene un valor por defecto)

jcamposmk: Quizás deberiamos tener un archivo `constants` en `/src` dónde se coloquen cosas generales como estas. Queda un poco más organizado, el más parecido que he visto donde se está haciendo algo como esto es en utils.js.

andresmoschini: Ya tenemos esos archivos, por ejemplo https://github.com/FromDoppler/doppler-webapp/blob/master/.env

Estos valores componen un singleton `appConfiguration` que puede inyectarse en los demás servicios: https://github.com/FromDoppler/doppler-webapp/blob/master/src/services/pure-di.tsx#L98-L123

Igualmente, en este caso no creo que tenga sentido moverlo ahí. Podemos simplemente dejar que el server decida el tamaño de las páginas.

PR #1537, opened 2021-08-06
fix: [DAT-516] Fix text button for last plan form

nreal:
# Background
For last plan popup, we fix text of button, change "Actualizar Plan" by "Enviar"

Before
image

After
image


jcamposmk:
:+1:

PR #1552, opened 2021-08-14
feat: integrate no list component and error message

ramniel:
- El usuario no tiene listas de suscriptores creadas: Mostrar modal con la opción de crear listas.
- Ocurre un error al cargar las listas: Mostrar mensaje de error en el modal.

### DIFF
--- a/src/components/ContactPolicy/ContactPolicy.js
+++ b/src/components/ContactPolicy/ContactPolicy.js
@@ -43,6 +43,7 @@ export const ContactPolicy = InjectAppServices(
     const [formSubmitted, setFormSubmitted] = useState(false);
     const [error, setError] = useState(false);
     const [modalIsOpen, setModalIsOpen] = useState(false);
+    const [modalClassName, setModalClassName] = useState('dp-modal-exclude-list');
     const [selectedLists, setSelectedLists] = useState([]);
     const intl = useIntl();
     const _ = (id, values) => intl.formatMessage({ id: id }, values);
@@ -196,7 +197,10 @@ export const ContactPolicy = InjectAppServices(
               >
                 {({ values, errors, isSubmitting, isValid, dirty, setFieldValue }) => (
                   <>
-                    <Prompt when={dirty} message={_('common.unsaved_changes_message')} />
+                    <Prompt
+                      when={dirty && !modalIsOpen}
+                      message={_('common.unsaved_changes_message')}
+                    />
                     <Form className="dp-contact-policy-form" aria-label="settings">
                       <fieldset>
                         <legend>{_('contact_policy.title')}</legend>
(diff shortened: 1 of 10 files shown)


src/components/SubscriberListSelector/NoList/NoList.js
```diff
@@ -0,0 +1,45 @@
...
+export const NoList = ({ onCancel }) => {
+  const intl = useIntl();
+  const _ = (id, values) => intl.formatMessage({ id: id }, values);
+
+  return (
+    <div className="dp-container" data-testid="no-list">
+      <div className="dp-rowflex">
+        <div className="col-sm-12">
+          <div className="dp-icon-modal">
+            <img src={_('common.ui_library_image', { imageUrl: 'lists.svg' })} alt="icon" />
```
jcamposmk, line 14: instead of alt = "icon", can you replace it with something more descriptive?

src/components/SubscriberListSelector/NoList/NoList.js
```diff
@@ -0,0 +1,45 @@
...
+            <FormattedMessageMarkdown id={'subscriber_list_selector.no_list.description_MD'} />
+            <span className="dp-cta-paragraph">
+              {_('subscriber_list_selector.no_list.strong_text')}
+            </span>
+            <hr />
+            <div className="dp-cta-modal">
+              <button
+                type="button"
+                className="dp-button button-medium primary-grey"
+                onClick={() => onCancel()}
```
jcamposmk, line 29:
```suggestion
onClick={onCancel}
```

src/components/SubscriberListSelector/NoList/NoList.js
```diff
@@ -0,0 +1,45 @@
+import { useIntl } from 'react-intl';
+import { FormattedMessageMarkdown } from '../../../i18n/FormattedMessageMarkdown';
+import React from 'react';
+
+export const NoList = ({ onCancel }) => {
```
jcamposmk, line 5: I think this component can be general. Might need to be used in another context. I think to this component should have props: `onCancel, onAction, logo, title, description, paragraph, cancelButtonText, actionButtonText` and should be in `shared` folder.

src/components/SubscriberListSelector/NoList/NoList.js
ramniel: sure, we can replace it by `alt="lists logo" `

src/components/SubscriberListSelector/NoList/NoList.js
ramniel: Sounds good to me, we would have to change the name too, a more generic name, what names can you think of?

ramniel: What do you think of these names: `Confirmation`, `ConfirmationBox` or `ConfirmationDialog`? @jcamposmk @cbernat

jcamposmk: I like something like <NoResultsFound /> . ConfirmationBox it sounds very tied to the implementation to me.

PR #1562, opened 2021-08-17
Dw 1017 add promotional advanced reports

jcamposmk:
### **Add advanced reports view**
Task: https://makingsense.atlassian.net/browse/DW-1017

**Scenario:** 

-  The user navigate with free account
- The following view is shown to the user: 

image

**Instead of:**
 
image


src/components/shared/Promotional/Promotional.js
```diff
@@ -50,7 +47,7 @@ export const Promotional = ({
           </div>
           <div className="col-lg-6 col-md-12">
             <figure className="dp-img-promotion">
-              <img src={_('common.ui_library_image', { imageUrl: previewUrl })} alt={title} />
+              <img src={previewUrl} alt={title} />
```
cbernat, line 50: We try to use images as styled components. Like for example [here](https://github.com/FromDoppler/doppler-webapp/blob/master/src/components/Integrations/Shopify/Shopify.styles.js#L7) I suppose it cannot be applied in this code because of the parameter, can it?

jcamposmk: I think it can. I will try

cbernat: It has no styles but it is coherent with other images in the app.

cbernat:
🚀 Nice!

PR #1564, opened 2021-08-18
feat: add after remove callback (optional)

jcamposmk:
### **Add callback to CloudTag component**
task: https://makingsense.atlassian.net/browse/DW-1094

**Description**
When `CloudTagCompoundField` component is used, can be passed as prop a callback afterRemove function.


cbernat:
👍

PR #1569, opened 2021-08-19
feat(bigquery): refactor AuthorizationPage with promotional component

kapoc3:
1. refactor to delete promotional component from bigquery for generic promotional componet
2. authorization page evaluate if plan of user is free or plus
3. when user not is plus show the promotional screen
4. add method for know the user plan
5. migrate test from promotional component to authorization page 
6. unify url for bigquery module
imagen


src/components/Integrations/BigQuery/AuthorizationPage.js
```diff
@@ -0,0 +1,72 @@
...
+      alert('error');
+    }
+  };
+
+  if (loading) {
+    return <Loading page />;
+  }
+
+  //TODO implement variable for validate if show promotional or control panel
+  if (userPlan.plan.planName === 'Plus Account') {
```
jcamposmk, line 46: What would happen if the planName changes? shoufd be like this `if(userPlan.plan.planName === 'Plus Account') `@cbernat

jcamposmk: Maybe you can to do something like this: 
```

if(is not a plus plan) {
    return 'promotional component'
}

return "authorization form"

```
it is easier to read

src/components/Integrations/BigQuery/AuthorizationPage.test.js
```diff
@@ -0,0 +1,111 @@
...
+          bigQueryClient: bigQueryClientDouble,
+        }}
+      >
+        <IntlProvider>
+          <AuthorizationPage />
+        </IntlProvider>
+      </AppServicesProvider>,
+    );
+
+    //act
```
jcamposmk, line 103:
```suggestion
assert
```

src/components/Integrations/BigQuery/AuthorizationPage.js
cbernat, line 46: The differentiator here should be if the user has the functionality enabled, not a Plus Account, remember that we will enable this for some beta tester users.

cbernat: You should validate if the feature is enabled with the user api, like this -> https://github.com/FromDoppler/doppler-webapp/blob/master/src/components/ContactPolicy/ContactPolicy.js#L80

jcamposmk:
:+1:

src/components/Integrations/BigQuery/AuthorizationPage.js
kapoc3: this functionality is a example of the real funcionality, 

Suggest accepted and applied.

cbernat:
Nice! 👍

PR #1570, opened 2021-08-19
chore: add husky into gitignore

cbernat:
### Add husky folder into gitignore

This is only for local use, so we should not upload anything from this folder. 
@jcamposmk is it ok this way? You are the expert in this 😄 

image


jcamposmk:
:+1:

PR #1581, opened 2021-08-26
[DAT-542] - Integrate "Payment method" section with the APIs (account-plans and billing-user)

jhoffman-ms:
Intregated the "Payment method" section with different apis to get information about payment and discounts.

**Task:** [DAT-542](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-542)


src/components/ChangePlan/Checkout/PaymentMethod/PaymentMethod.js
```diff
@@ -4,150 +4,188 @@ import { useIntl } from 'react-intl';
...
+const amexDescription = 'amex';
 
-  useEffect(() => {
-    const fetchData = async () => {
-      const sessionPlan = appSessionRef.current.userData.user;
-      const discounts = fakeAccountPlanDiscounts;
+const getCreditCardBrand = (ccType) => {
+  // check for American Express
+  if (ccType === creditCardType.amex) {
+    return amexDescription;
```
cbernat, line 48: Why are you doing this? It's a bit weird...

jcamposmk: I think it's because of react-credit-card, I'm not sure. 
image

jcamposmk: this function maybe not be required. I think it would be enough just to pass `ccType`, can it be @jhoffman-ms  ?

jhoffman-ms: Yes, exactly. The reason of this is the API returns "American Express" and the credit-card component supports 'amex'

jhoffman-ms: > this function maybe not be required. I think it would be enough just to pass `ccType`, can it be @jhoffman-ms ?

The function is neccesary because the API returns "American Express" and the component needs 'amex'

jcamposmk:
:+1:

src/components/ChangePlan/Checkout/PaymentMethod/PaymentMethod.js
cbernat: Thanks for the clarification!.

PR #1596, opened 2021-09-01
Dw 1101 change promotional gif of contact policy

jcamposmk:
**Add GIF to Promotional Contact Policy**

**scenario:** 

- The user navigates with free account to `/sending-preferences/contact-policy`
- The user sees the promotional component of contact policy with the new GIF


cbernat:
Supposing this image will be used only in contact policies, wouldn´t it be better to add this in the same folder as the component? As part of the component.

jcamposmk:
there are several approaches ... I prefer to add it in the same folder of the component, along with the styles and the tests. and in the img folder add the ones that are reused. it is more scalable.
it would be nice to apply this same approach in the project. Especially in the services folder. it will surely continue to increase more and more
image

cbernat:
> there are several approaches ... I prefer to add it in the same folder of the component, along with the styles and the tests. and in the img folder add the ones that are reused. it is more scalable.
> it would be nice to apply this same approach in the project. Especially in the services folder. it will surely continue to increase more and more
> image

This we could do, with patience. Perhaps  whenever we need to modify a service we could move it into a folder, with the double and the test. What do you think?

cbernat:
👍 🚀

jcamposmk:
> > there are several approaches ... I prefer to add it in the same folder of the component, along with the styles and the tests. and in the img folder add the ones that are reused. it is more scalable.
> > it would be nice to apply this same approach in the project. Especially in the services folder. it will surely continue to increase more and more
> > image
> 
> This we could do, with patience. Perhaps whenever we need to modify a service we could move it into a folder, with the double and the test. What do you think?

Yes, I agree :+1:  slowly

PR #1599, opened 2021-09-01
refactor: move buy link to up in header menu

jcamposmk:
### **Change the location of the upgrade link**

**Result:**

image


cbernat:
👍

src/components/Header/HeaderUserMenu/HeaderUserMenu.test.js
```diff
@@ -75,8 +75,11 @@ describe('Header user menu', () => {
     // Arrange
     userData.plan = {
       clientManager: {},
+      hasClientManager: false,
+      isSubscribers: true,
```
cbernat, line 79: I'm not sure about this flags, with `clientManager` property already here should suffice.

PR #1603, opened 2021-09-03
chore(gitignore): Add husky content only into gitignore

fgchaio:
I was having issues running the local environment because the .husky folder is added into the .gitignore file.
With these changes, only the content will be ignored.


.gitignore
```diff
@@ -24,5 +24,4 @@ yarn-error.log*
 
 /.vscode
 /.eslintcache
-.idea
-/.husky/
+.idea
```
cbernat, line 27: Why not this?
```suggestion
.idea
/.husky/**/**.*
```

fgchaio: Because if I do it like that git wants to add the folder content into de project.

cbernat: But with your changes, I remove the .husky folder, then do a `yarn install` and when I try to commit, husky files and folder appear. 
image

fgchaio:
I'm closing this PR. This doesn't seem to be an issue to others. I will update yarn to the latest version and it might solve it.

PR #1612, opened 2021-09-07
fix(BannerUpgrade): fix object links

silsanchez8:
## Modifications: https://makingsense.atlassian.net/browse/DW-1135
* fix(BannerUpgrade): fix object links

image

image

image


cbernat:
Perhaps it would be nice if we could add some test to validate that the link url is not #. Perhaps add this in an existent test.

cbernat:
Tested for free and plus and it´s ok https://cdn.fromdoppler.com/doppler-webapp/int-build3840/#/plan-selection. 😄

jcamposmk:
:+1:

PR #1615, opened 2021-09-08
DW-1134-fix-monthly-deliveries-plan-tabs

silsanchez8:
## Modifications : https://makingsense.atlassian.net/browse/DW-1134
* fix(plan-service): Fix monthly deliveries tabs. Do not show Subscribers tab.

image

### DIFF
--- a/src/services/plan-service.test.js
+++ b/src/services/plan-service.test.js
@@ -260,7 +260,7 @@ describe('Doppler plan client', () => {
     var types = planService.getPlanTypes(currentPlan, 'standard', planList);
 
     // Assert
-    expect(types.length).toBe(2);
+    expect(types.length).toBe(1);
   });
 
   it('should get correct plans for free user - selected path plus, type subscribers', async () => {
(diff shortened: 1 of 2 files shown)


src/services/plan-service.test.js
```diff
@@ -351,7 +351,7 @@ describe('Doppler plan client', () => {
     const plans = await planService.getPlans(currentPlan, 'standard', 'subscribers', planList);
 
     // Assert
-    expect(plans.length).toBe(2);
+    expect(plans.length).toBe(0);
```
cbernat, line 354: I think this test is no longer valid or you could change the name. 
Instead of 
`it('should get higher subscriber plans when user type: monthly-deliveries', async () => {`
`it('should get no subscriber plan when user type: monthly-deliveries', async () => {`
What do you think? remove it or change the name? 🤔

silsanchez8: Maybe it could be better to rename the test, so we can be sure that subscriber plan is not going to be showed when plan user type is monthly deliveries 😉

cbernat:
👍

PR #1617, opened 2021-09-09
DW-1141: fix(PlanCalculator): for monthly plans there shouldn't be discounts

cbernat:
### Don`t allow discounts for monthly plans

It seems that in DB there are some discounts for monthly plans, that is not correct. But also it is not correct to do the mapping. So I've removed the mapping and then added a test as documentation for the future. 

image


jcamposmk:
:+1:

PR #1618, opened 2021-09-09
fix(breadcrumbs): Fix breadcrumb form links

silsanchez8:
## Modifications : https://makingsense.atlassian.net/browse/DW-1146

_fix(breadcrumbs): Fix breadcrumb form links in:_
- _Agencies Form_
- _Upgrade Suggestion Form_
- _Exclusive Form_
image


src/i18n/en.js
```diff
@@ -38,7 +38,7 @@ const urlUpgradePlan = `${urlDopplerLegacy}/ControlPanel/AccountPreferences/Upgr
 const messages_en = {
   agencies: {
     breadcrumb: 'Plans',
-    breadcrumb_url: `${urlBuyMonthly}`,
+    breadcrumb_url: `${urlPlanSelection}`,
```
cbernat, line 41: Perhaps this url and the breadcrumb text should be in common, like the text `Control Panel` and some of its urls https://github.com/FromDoppler/doppler-webapp/blob/master/src/i18n/en.js#L225
Something like this:
```
common: {
  ...
  plan_text: 'Plans',
  plan_url: `${urlPlanSelection}`,
  ...
```
What do you think?

cbernat:
👍

PR #1627, opened 2021-09-13
DW-1148-cost-per-email-credits-plan

silsanchez8:
## Modifications: https://makingsense.atlassian.net/browse/DW-1148

* feat(PlanAgreement): Add cost per mail in prepaid plans
* Fix some Creditcard test wich was using an all function called _replaceAll_

image

### DIFF
--- a/src/components/ChangePlan/PlanCalculator/PlanAgreement/PlanAgreement.js
+++ b/src/components/ChangePlan/PlanCalculator/PlanAgreement/PlanAgreement.js
@@ -40,7 +40,7 @@ export const PlanAgreement = ({ planData }) => {
         </p>
       ) : null}
 
-      {planData.plan.type === 'monthly-deliveries' ? (
+      {planData.plan.type === 'monthly-deliveries' || planData.plan.type === 'prepaid' ? (
         <p className="dp-cost-per-email">
           {_('plan_calculator.cost_per_email')}{' '}
           <span className="dp-price-large-money">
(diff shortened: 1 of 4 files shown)


src/components/ChangePlan/PlanCalculator/PlanAgreement/PlanAgreement.js
```diff
@@ -40,7 +40,7 @@ export const PlanAgreement = ({ planData }) => {
         </p>
       ) : null}
 
-      {planData.plan.type === 'monthly-deliveries' ? (
+      {planData.plan.type !== 'subscribers' ? (
```
cbernat, line 43: So the value by email is mapped into the same field? I would prefer being explicit here, like this:
```suggestion
      {planData.plan.type === 'monthly-deliveries' && planData.plan.type === 'prepaid'  ? (
```
We have several plans, `subscribers`, `monthly-deliveries` `agencies`, `prepaid`, when I read planData.plan.type === 'monthly-deliveries' I have to think twice to understand that you are not including agencies or exactly to which plans you are pointing out.

cbernat:
Would you mind adding a test to reinforce this mapping please? 😄

silsanchez8:
> Would you mind adding a test to reinforce this mapping please? 😄
Ceci, there is already a test for theses cases: 
Do you think it's enough or maybe we have to create a new one?
Prepaid:
image

Subscribers:
image

Emails:

image

cbernat:
🚀

PR #1628, opened 2021-09-14
refactor: fix spinner on reports page

jcamposmk:
### Fix spinner on reports page

**scenario:**

- The user navigates to /reports page when a verificated domain
- The user sees a spinner and after the user sees a spinner for each box with a placeholder content (fake data)

**Before:**
image

**After:**
image

### DIFF
--- a/src/components/Reports/Reports.js
+++ b/src/components/Reports/Reports.js
@@ -2,7 +2,6 @@ import React, { useEffect, useState } from 'react';
 import { FormattedMessage } from 'react-intl';
 import { Helmet } from 'react-helmet';
 import { InjectAppServices } from '../../services/pure-di';
-import { Loading } from '../Loading/Loading';
 import { BoxMessage } from '../styles/messages';
 import { addDays, getStartOfDate } from '../../utils';
 import ReportsFilters from './ReportsFilters/ReportsFilters';
@@ -35,7 +34,7 @@ const Reports = ({ dependencies: { datahubClient } }) => {
     dailyView: periodSelectedDaysDefault === 1,
   });
 
-  const [totalVisits, setTotalVisits] = useState({});
+  const [totalVisits, setTotalVisits] = useState({ loading: true });
 
   const changeDomain = async (name) => {
     const domainFound = state.domains.find((item) => item.name === name);
@@ -113,11 +112,9 @@ const Reports = ({ dependencies: { datahubClient } }) => {
             changePeriod={changePeriod}
             isEnableWeeks
           />
-          {state.loading ? (
-            <Loading />
-          ) : state.domains ? (
+          {state.loading || (!state.loading && state.domains) ? (
             <section className="dp-container">
-              {!state.domainSelected.verified_date ? (
+              {state.domainSelected && !state.domainSelected.verified_date ? (
                 <BoxMessage className="dp-msj-error bounceIn" spaceTopBottom>
                   <span>
                     <FormattedMessageMarkdown id="reports_filters.domain_not_verified_MD" />
@@ -132,7 +129,7 @@ const Reports = ({ dependencies: { datahubClient } }) => {
                     today={state.dailyView}
                     emailFilter={'without_email'}
                     visits={totalVisits.withoutEmail}
-                    loading={totalVisits.loading}
+                    loading={state.loading || totalVisits.loading}
                   />
                 </div>
                 <div className="col-lg-6 col-md-6 col-sm-12 m-b-24">
@@ -142,13 +139,13 @@ const Reports = ({ dependencies: { datahubClient } }) => {
                     today={state.dailyView}
                     emailFilter={'with_email'}
                     visits={totalVisits.withEmail}
-                    loading={totalVisits.loading}
+                    loading={state.loading || totalVisits.loading}
                   />
                 </div>
                 {!state.dailyView ? (
                   <div className="col-sm-12 m-b-24">
                     <ReportsDailyVisits
-                      domainName={state.domainSelected.name}
+                      domainName={state.domainSelected?.name}
                       dateFrom={state.dateFrom}
                       dateTo={state.dateTo}
                     />
@@ -156,7 +153,7 @@ const Reports = ({ dependencies: { datahubClient } }) => {
                 ) : null}
                 <div className="col-sm-12 m-b-24">
                   <ReportsTrafficSources
-                    domainName={state.domainSelected.name}
+                    domainName={state.domainSelected?.name}
                     dateFrom={state.dateFrom}
                     dateTo={state.dateTo}
                   />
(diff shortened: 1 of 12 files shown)


src/components/Reports/Reports.js
```diff
@@ -142,37 +139,37 @@ const Reports = ({ dependencies: { datahubClient } }) => {
...
                   <div className="col-sm-12 m-b-24">
                     <ReportsDailyVisits
-                      domainName={state.domainSelected.name}
+                      domainName={state.domainSelected?.name}
                       dateFrom={state.dateFrom}
                       dateTo={state.dateTo}
                     />
                   </div>
-                ) : null}
+                )}
```
andresmoschini, line 153: Are you sure about replacing the ternary operator with a boolean operation?

What happens when it evaluates as false?

See https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

src/components/Reports/Reports.js
```diff
@@ -142,37 +139,37 @@ const Reports = ({ dependencies: { datahubClient } }) => {
...
-                    loading={totalVisits.loading}
+                    loading={state.loading || totalVisits.loading}
                   />
                 </div>
-                {!state.dailyView ? (
+                {!state.dailyView && (
                   <div className="col-sm-12 m-b-24">
                     <ReportsDailyVisits
-                      domainName={state.domainSelected.name}
+                      domainName={state.domainSelected?.name}
```
andresmoschini, line 148: Is it a possible scenario? I mean is rendering this element fine when the _state.domainSelected_ does not have value?

andresmoschini:
image

It changes the system behavior, so, it is not a refactoring.

In my opinion, the commit prefix should be `fix:`

src/components/Reports/Reports.js
andresmoschini: Also, these changes are clearly a refactor, it is not related to changing behavior. 

In my opinion, it is better to split these kinds of changes of behavioral changes in different commits or even different PRs. See https://medium.com/@kentbeck_7670/bs-changes-e574bc396aaa

Maybe, the [refactoring is required to do the behavior change](https://twitter.com/KentBeck/status/250733358307500032). In this case, you can have a single PR with two commits: the first for the refactoring and the second for the behavior.

src/components/Reports/Reports.test.js
```diff
@@ -73,51 +68,42 @@ describe('Reports page', () => {
...
     // Act
-    const { getByText, container } = render(
-      <AppServicesProvider
-        forcedServices={{
-          datahubClient: datahubClientDouble,
-          appConfiguration: { dopplerLegacyUrl: 'http://test.localhost' },
-        }}
-      >
+    const { queryByText, container } = render(
+      <AppServicesProvider forcedServices={dependencies}>
```
andresmoschini, line 72: image

How does this change work?

Before:

```js
forcedServices = {
  datahubClient: { 
      getAccountDomains: ... // 👈
  },
  appConfiguration: { dopplerLegacyUrl: 'http://test.localhost' },
};
```

Now:

```js
forcedServices = { 
      getAccountDomains: ... // 👈
  };
```

So, the `forcedServices` receives now a totally different structure, is it fine?

src/components/Reports/ReportsBox/ReportsBox.js
```diff
@@ -8,38 +8,31 @@ import { Loading } from '../../Loading/Loading';
...
  */
 const ReportsBox = ({ dateFrom, dateTo, emailFilter, today, visits, loading }) => {
+  const withEmail = 'with_email' === emailFilter;
+  const totalVisits = visits ?? 0;
+  const descriptionKey = withEmail
+    ? 'reports_box.visits_description_with_email'
+    : 'reports_box.visits_description_without_emails';
+  const titleKey = withEmail
+    ? 'reports_box.visits_with_email'
+    : 'reports_box.visits_without_emails';
```
andresmoschini, line 18: I suppose that these changes are part of the behavior change. It would be nice to have them in a clean commit to allow the reviewers to see what was changed at a glance.

src/components/Reports/Reports.test.js
andresmoschini: It seems that these changes are a kind of refactoring, it would be nice to have them in a different PR to easily understand the impact of the change and to be used as a reference in further changes.

src/components/Reports/ReportsFilters/ReportsFilters.js
```diff
@@ -4,6 +4,8 @@ import * as S from './ReportsFilters.styles';
 import { FormattedMessageMarkdown } from '../../../i18n/FormattedMessageMarkdown';
 import HeaderSection from '../../shared/HeaderSection/HeaderSection';
 
+export const placeholderDate = '--/--/----';
```
andresmoschini, line 7: It seems to be a different change, maybe a different behavior commit?

andresmoschini:
Nice work @jcamposmk!

❤️ We were waiting for a long time to have this behavior.

src/components/Reports/Reports.js
jcamposmk: [More details](https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored)

image

andresmoschini: So, Kent Dodds worries are wrong?

src/components/Reports/Reports.js
jcamposmk: yes, it's possible. Because the `<ReportsDailyVisits />` is shown when the domains are being searched, in that case the `state.domainSelected` is `null`

src/components/Reports/Reports.js
andresmoschini: In my opinion, this discussion is not related to the change with the spinners. 

This change should have its own PR. Because in another way, tomorrow I apply another change and mixed with it I refactor it again based on Doods opinion 😛, the next day you change it again and so.

src/components/Reports/ReportsFilters/ReportsFilters.js
jcamposmk: it can go in the same commit of the use of spinners, because it is a placeholder text that is used while looking for the last visit

src/components/Reports/Reports.js
andresmoschini: Thanks for the clarification! now `ReportsDailyVisits` is ready to deal with `domainName = null` and before it did not broke because these elements were not being rendered when loading was true.

This is a central part of your change, not only a refactoring.

src/components/Reports/ReportsHoursVisits/ReportsHoursVisits.js
```diff
@@ -27,51 +28,69 @@ const FormatWeekDayIndex = ({ value, format }) => {
...
+    { byWeekDayAndHour: createEmptyWeekDayHoursMatrix(), max: 0 },
+  );
+
+  return processedVisits;
+};
+
+const fakeData = getProcessedVisits({
+  success: true,
+  value: prepareFakeVisitsWeekdayHoursData(),
+});
```
cbernat, line 55: This is the placeholder data right? For loading only?

jcamposmk: Yes. it is ...

cbernat:
Tested in production with mfantini account, and it's working wonderfully. https://cdn.fromdoppler.com/doppler-webapp/build3881/#/reports Good job @jcamposmk ! 🎉

PR #1630, opened 2021-09-17
feat: filter plans by contacts for users with a prepaid account

jcamposmk:
### filter plans by contacts for users with a prepaid account
task: https://makingsense.atlassian.net/browse/DW-1140

**scenarios:** 

**The user has more subscribers than what was offered in the account per contact**

- The user navigates to `/plan-selection/standard/prepaid` with a prepaid account
- The user don't see the contact tab

Example when the user has 120.000 subscribers: 
image

**The user has less subscribers than what was offered in the account per contact**

- The user navigates to `/plan-selection/standard/prepaid` with a prepaid account
- The user sees the contact tab

Example when the user has 1600 subscribers:
image

**The user has the same number of subscribers as the maximum offered in the accounts per contact**

- The user navigates to `/plan-selection/standard/prepaid` with a prepaid account
- The user sees the contact tab, but the slider is at maximum

Example when the user has 100000 subscribers (100k)
image

### DIFF
--- a/src/components/ChangePlan/PlanCalculator/PlanCalculator.js
+++ b/src/components/ChangePlan/PlanCalculator/PlanCalculator.js
@@ -82,6 +82,7 @@ const PlanCalculator = ({ location, dependencies: { planService, appSessionRef }
         sessionPlan.planType,
         sessionPlan.idPlan,
         sessionPlan.planSubscription,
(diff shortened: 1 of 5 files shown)


cbernat:
Wonderful! 🚀

PR #1631, opened 2021-09-17
fix(CloudTags): temporarily remove dp-recently-add class

cbernat:
### Remove green glow
Remove green glow from the last tag, that did not indicate nothing in particular to avoid confusion. 

This must be replaced in the future by adding the glow for the last unsaved tags added. 
[Jira ticket link](https://makingsense.atlassian.net/browse/DW-1161)

Before
image
After
image


jcamposmk:
:+1:

PR #1632, opened 2021-09-17
DW-1159-exclusive-clients

silsanchez8:
## Modifications: https://makingsense.atlassian.net/browse/DW-1159

* Fix(ChangePlan): redirect to upgradeform if exclusive clien
redirect

### DIFF
--- a/src/components/ChangePlan/ChangePlan.js
+++ b/src/components/ChangePlan/ChangePlan.js
@@ -6,7 +6,7 @@ import queryString from 'query-string';
 import { extractParameter, thousandSeparatorNumber } from '../../utils';
 import { InjectAppServices } from '../../services/pure-di';
 import { Loading } from '../Loading/Loading';
-import { Link } from 'react-router-dom';
+import { Link, Redirect } from 'react-router-dom';
 import Collapse from '@kunukn/react-collapse';
 import * as S from './ChangePlan.styles';
 import { FAQ } from '../FAQ';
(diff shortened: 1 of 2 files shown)


src/components/ChangePlan/ChangePlan.test.js
```diff
@@ -442,4 +453,50 @@ describe('ChangePlan component', () => {
...
   });
+  it('should redirect to UpgradeForm when user has exclusive plan', async () => {
+    // Arrange
+    const paths = [{}];
+    const dependencies = {
+      appSessionRef: {
+        current: {
+          userData: {
+            user: {
+              plan: {},
```
cbernat, line 464: You could add the exclusive plan here and let the planservice do de mapping right? Perhaps not, you tell me. 😄

src/components/ChangePlan/ChangePlan.js
```diff
@@ -267,6 +267,8 @@ const ChangePlan = ({ location, dependencies: { planService, appSessionRef } })
       </Helmet>
       {state.loading ? (
         <Loading page />
+      ) : state.currentPlan?.type === 'exclusive' ? (
```
jcamposmk, line 270: Maybe something like this will be better, It's more simple: 
image

src/components/ChangePlan/ChangePlan.test.js
silsanchez8: In don't understand how to do it, tomorrow I will call you 📲

PR #1640, opened 2021-09-21
chore(bigquery): remove notification messages for demo in production

kapoc3:
remove validation for notification messages.

--> this change will be included when the messages logic be implement


jcamposmk:
:+1:

PR #1646, opened 2021-09-30
DW-1121: fix(SiteTrackingRequired): Remove intermediate screen

cbernat:
Before if a paid user never accesed into advanced reports he saw this:

image

Now he will see the second screen directly:

image


jcamposmk:
:+1:

PR #1648, opened 2021-10-01
fix(zoho-chat): remove zoho chat

cbernat:
### Remove zoho chat
In the app we can see an error in zoho script, that we do not use, so I removed it.

image

```
zoho-chat.js:43 Uncaught TypeError: Cannot read properties of null (reading 'style')
    at zoho-chat.js:43
```


jcamposmk:
:+1:

PR #1708, opened 2021-11-03
feat(ControlPanel): Add connection status for integrations

fgchaio:
Add connection status for native integrations and advanced preferences. The idea is to show the user at a first glance which integrations are currently active. Also, the status should be ordered by "ALERT", "CONNECTED", "NOT CONNECTED", and lastly "no status".
image

TODO: add tests

Looking forward to your reviews!!


src/components/ControlPanel/ControlPanel.js
```diff
@@ -25,16 +44,36 @@ export const ControlPanel = InjectAppServices(({ dependencies: { controlPanelSer
...
-                  ))}
+                  {controlPanelSections[indexSection].boxes
+                    .sort(sortByStatus)
+                    .map((box, indexBox) => (
+                      <ControlPanelBox
+                        box={box}
+                        key={`box-${indexBox}`}
+                        disabled={!!box.disabled}
+                        hidden={!!box.hidden}
+                      />
```
andresmoschini, line 75: I understand that it is out of the scope of the PR, but maybe for a future PR, what do you think about that simply read `disabled` and `hidden` properties inside `ControlPanelBox`?:

```suggestion
                      <ControlPanelBox
                        box={box}
                        key={`box-${indexBox}`}
                      />
```

src/components/ControlPanel/ControlPanelBox/ControlPanelBox.js
```diff
@@ -20,12 +23,28 @@ export const ControlPanelBox = ({ box, disabled = false, hidden = false }) => {
...
+          <>
+            {box.status ? (
+              <S.StatusImage
+                src={
+                  box.status === 'connected'
+                    ? connected
+                    : box.status === 'alert'
+                    ? connection_alert
+                    : disconnected
+                }
```
andresmoschini, line 35: What do you think about doing this mapping before rendering?

```suggestion
                src={statusImage}
```

src/components/ControlPanel/ControlPanelBox/ControlPanelBox.js
```diff
@@ -20,12 +23,28 @@ export const ControlPanelBox = ({ box, disabled = false, hidden = false }) => {
...
+                alt=""
```
andresmoschini, line 36: What do you think about using the status as `alt`?

```suggestion
                alt={box.status}
```

src/components/ControlPanel/ControlPanel.js
fgchaio: I did it like that to explicitly set a default value for those properties. But I guess it's not much of a difference. I will remove it in another PR 😉

fgchaio:
@andresmoschini for now I will leave this PR like this and apply the refactor using the useReducer when I add the request for the integration statuses.

PR #1710, opened 2021-11-03
feat: [DAT-598] Integrate with POST agreement service

nreal:
# Background
We need to send this information to create new agreement to the user
```
{
    total: 23
    discountId: "2"
    planId: "3"
}
```


jhoffman-ms:
👍

PR #1711, opened 2021-11-03
[DAT-675] - The free user cannot select payment method

jhoffman-ms:
Fix issue when the free user has not select payment method. 

Now when the user has not payment method the "Payment method' section is show in update mode and with 'Credit Card' option selected and also initialized the properties.

**Task:** [DAT-675](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-675)


bamaro-dev (commented):
✔️

PR #1718, opened 2021-11-04
feat(checkout): Apply background styles regarding UI library

jhoffman-ms:
Apply background styles regarding UI library

**Before:**

image

**After:**

image


nreal:
👍

PR #1725, opened 2021-11-08
fix (checkout): When the user changes the country the state was left with the previous value

jhoffman-ms:
Fix issue when the user changes the country, the state was left with the previous value instead of the the empty value

**Bug:** [DAT-685](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-685)


bamaro-dev (commented):
Great!

PR #1736, opened 2021-11-11
[DOI-350] Modify control panel styles

fgchaio:
After reviewing it with Mariana we concluded on a few small changes in the style.
Also, I'm disabling the "showStatus" property to publish it like this and keep working on the connection status.


jcamposmk:
:+1:

PR #1749, opened 2021-11-16
fix: fix space between amount & currency

jcamposmk:
Space between currency and amount in plan calculator

**When a discount is selected**

Captura de Pantalla 2021-11-16 a la(s) 10 41 06 a  m

**When plan type is for emails**
Captura de Pantalla 2021-11-16 a la(s) 10 43 14 a  m


andresmoschini:
👍

PR #1765, opened 2021-11-29
fix: replace shopify url on int a qa environment

blancfabian:
Replace shopify url on int a qa environment, this is for testing the new shopify app.


blancfabian:
> Wow! Will it start to use the new integration? what happens with current users?

The idea is to only target the new application in Int and QA environments, we will not change the production one until we have finished all the tests and VN Studio has finished the current users data migration.

andresmoschini:
👍

PR #1798, opened 2021-12-15
feat(checkout): Use useQueryParams hooks to get the parameters

jhoffman-ms:
Use useQueryParams hooks to get the parameters


jcamposmk:
👍 💯

PR #1853, opened 2022-02-03
[DAT-718] - Change logic to redirect new checkout

jhoffman-ms:
Change logic to redirect new checkout. Now only redirect to the new checkout if the user is free, the country is different to 'AR', 'MX, and 'CO" and the buy is by credits

**Task:** [DAT-718](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-718)


jcamposmk:
Good Job! 👍

PR #1854, opened 2022-02-04
chore: conventional commits

jcamposmk:
## **Conventional Commits**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-766

### **Use cases**

1. **Example commit message:** `add new component`  ❌
Captura de Pantalla 2022-02-04 a la(s) 8 12 56 a  m

2.  **Example commit message:** `foo: add new component` ❌
Captura de Pantalla 2022-02-04 a la(s) 8 18 56 a  m

3. **Example commit message:** `feat:add new component` ❌
Captura de Pantalla 2022-02-04 a la(s) 8 21 01 a  m

4. **Example commit message:** `feat: add new component` ✅✅
Captura de Pantalla 2022-02-04 a la(s) 8 22 04 a  m

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+    extends: ['@commitlint/config-conventional'],
+};
--- a/package.json
+++ b/package.json
@@ -149,6 +149,8 @@
     "comment": "Resolution to avoid test warning: [...] Please update the following components: SideEffect(NullComponent)"
   },
   "devDependencies": {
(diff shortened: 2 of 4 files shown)


andresmoschini:
👍 

It is a different approach that we are using on other projects where we are using `gitlint`: https://github.com/FromDoppler/doppler-editors-webapp/blob/main/Jenkinsfile#L6

I am not comfortable with pre-commit hooks (and Husky), but I think that it is good that different teams follow different approaches. Eventually, they could converge (or not).

🚀

PR #1859, opened 2022-02-07
fix(checkout-summary): Fix issue when redirect checkout summary from old checkout

jhoffman-ms:
Fix issue when redirect checkout summary from old checkout

**Bug:** [DAT-780](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-780)


jcamposmk:
👍

PR #1872, opened 2022-02-17
fix: remove warning message when there is not promocode

jcamposmk:
## **Fix warning message in Plan Calculator**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-804

Remove warning message when there is not promocode


jhoffman-ms:
👍

PR #1905, opened 2022-03-16
[DAT-832] - Allow purchases plan by credits, emails or contacts for free users and that their country is not in AR/MX/CO

jhoffman-ms:
Allow purchases plan by credits, emails or contacts for free users and that their country is not in AR/MX/CO

**Task:** [DAT-832](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-832)


jcamposmk:
👍

PR #1909, opened 2022-03-18
[No-Task] - Remove an unnecessary service call in the Transfer component

jhoffman-ms:
Remove an unnecessary service call in the Transfer component

**Task:** No task


jcamposmk:
👍

PR #1918, opened 2022-03-29
[DAT-860] - fix issue when the user selects the discount from the PlanCalculator page

jhoffman-ms:
Fix issue when the user selects the discount from the PlanCalculator page

Bug: [DAT-860](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-860)


jcamposmk:
👍

PR #1957, opened 2022-04-29
[DOI-505] Fix anchor links in control panel

fgchaio:
Kudos to @jcamposmk, he gave me a huge help and came up with this solution.
The anchor links weren't making focus when accessing from outside the control panel.

Thanks a lot Junior!!

### DIFF
--- a/src/components/ControlPanel/ControlPanel.js
+++ b/src/components/ControlPanel/ControlPanel.js
@@ -15,6 +15,7 @@ import {
   CONTROL_PANEL_SECTIONS_ACTIONS,
   controlPanelSectionsReducer,
 } from './reducers/controlPanelSectionsReducer';
+import useHashScrollHandler from '../../hooks/useHashScrollHandler';
 
 const sortByStatus = (a, b) => {
   return a.status
@@ -70,6 +71,8 @@ export const ControlPanel = InjectAppServices(
 
       fetchData();
     }, [controlPanelService, dopplerUserApiClient]);
+
+    useHashScrollHandler();
 
     if (loading) {
       return <Loading page />;
--- a/src/hooks/useHashScrollHandler/index.js
+++ b/src/hooks/useHashScrollHandler/index.js
@@ -1,0 +1,18 @@
+import { useEffect } from 'react';
+
+const useHashScrollHandler = () => {
+  useEffect(() => {
+    const { hash } = window.location;
+    if (hash) {
+      setTimeout(() => {
+        const id = hash.replace('#', '');
+        const element = document.getElementById(id);
+        if (element) {
+          element.scrollIntoView();
+        }
+      }, 0);
+    }
+  }, []);
+};
+
+export default useHashScrollHandler;


src/components/ControlPanel/ControlPanel.js
```diff
@@ -71,6 +73,16 @@ export const ControlPanel = InjectAppServices(
       fetchData();
     }, [controlPanelService, dopplerUserApiClient]);
 
+    useEffect(() => {
+      createTimeout(() => {
+        if (window.location.hash) {
+          const currentHash = window.location.hash;
+          window.location.hash = '';
+          window.location.hash = currentHash;
```
andresmoschini, line 81: It is really weird.

Looks like a patch, in that case, could you add a code comment about what it is doing and why?

In general, using the global object is not a good idea, we are ready to inject it as a dependency if it is inevitable.

Could you add a test for it?

andresmoschini: > The anchor links weren't making focus when accessing from outside the control panel.

If I am not wrong, it forces navigation (slow and expensive), I think that we should deal with anchors in javascript. There should be a library or hook for that.

fgchaio: Thanks @andresmoschini, I will contact you on monday to talk about this and find a better approach.

andresmoschini: Probably we could find useful information in some of these links:

* https://github.com/remix-run/react-router/issues/394
* https://github.com/rafgraph/react-router-hash-link/issues/13#issuecomment-533251165
* https://gajus.medium.com/making-the-anchor-links-work-in-spa-applications-618ba2c6954a
* https://github.com/rafgraph/react-router-hash-link
* https://github.com/gabergg/react-scrollable-anchor

andresmoschini: We talk with @fgchaio.

We agree that [Gajus' patch](https://gajus.medium.com/making-the-anchor-links-work-in-spa-applications-618ba2c6954a) could be the more general solution but we do not know what bugs could it have.

So, @fgchaio will move his code to a hook in order to make this clear, and also reusable, and he will also change the `window.location.hash` manipulation by `scrollToElement()`.

Since he is only forcing the code execution at the end of the event loop, probably using `useTimeout` hook is not necessary and he could use `setTimeout` directly. What do you think @jcamposmk?

andresmoschini: We talk with @fgchaio.

We agree that [Gajus' patch](https://gajus.medium.com/making-the-anchor-links-work-in-spa-applications-618ba2c6954a) could be the more general solution but we do not know what bugs could it have.

So, @fgchaio will move his code to a hook in order to make this clear, and also reusable, and he will also change the `window.location.hash` manipulation by `scrollToElement()`.

Since he is only forcing the code execution at the end of the event loop, probably using `useTimeout` hook is not necessary and he could use `setTimeout` directly. What do you think @jcamposmk?

jcamposmk: sounds good! 👍

andresmoschini:
❤️

PR #1971, opened 2022-05-11
feat: add user email by default in InvoiceRecipients

jcamposmk:
### **Add user email by default at InvoiceRecipients**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-925

**Description**
When the user does not have added recipients. The email user is used by default.


jhoffman-ms:
👍

PR #2003, opened 2022-06-10
[DAT-977] - Update security code field in mercado pago

manu3210:
Update security code field in mercado pago because we have an issue when the field type is `tel`

Task: [DAT-977](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-977)


jhoffman-ms:
👍

PR #2007, opened 2022-06-14
feat: add user email by default to invoice recipients

jcamposmk:
### **Add user email by default at InvoiceRecipients**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-925

**Description**
When the user does not have added recipients. The email user is used by default.

### DIFF
--- a/src/components/Plans/Checkout/PurchaseSummary/InvoiceRecipients/index.js
+++ b/src/components/Plans/Checkout/PurchaseSummary/InvoiceRecipients/index.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { useIntl, FormattedMessage } from 'react-intl';
 import { FieldGroup, FieldItem, SubmitButton } from '../../../../form-helpers/form-helpers';
 import { Form, Formik } from 'formik';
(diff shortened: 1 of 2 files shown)


jhoffman-ms:
👍

src/components/Plans/Checkout/PurchaseSummary/InvoiceRecipients/index.js
```diff
@@ -17,42 +17,59 @@ export const InvoiceRecipients = InjectAppServices(
...
-          setRecipients(response.value.recipients);
+        const { success, value } = await dopplerBillingUserApiClient.getInvoiceRecipientsData();
+        if (success) {
+          if (value.recipients?.length > 0) {
+            setRecipients(value.recipients);
+          } else {
+            submitEditRecipients({
+              [fieldNames.editRecipients]: [userEmail],
+            });
+          }
```
andresmoschini, line 58: I did not read the logic, but, in general, we are trying to reduce nesting. What do you think about this:

```suggestion
        const { success, value } = await dopplerBillingUserApiClient.getInvoiceRecipientsData();
        if (success && value.recipients?.length > 0) {
          setRecipients(value.recipients);
        } else if (success) {
          submitEditRecipients({
            [fieldNames.editRecipients]: [userEmail],
          });
```

jcamposmk: it's ok! better!

PR #2024, opened 2022-07-05
[DAT-998] - Integrate the upgrade between plans (only for emails) with the new checkout

jhoffman-ms:
Integrate the upgrade between plans (only for emails) with the new checkout

Now the user can do upgrade between plan, currently only supports for montly plan.

**Task:** [DAT-998](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-998)


jcamposmk:
👍

PR #2026, opened 2022-07-07
[DAT-1014] - Create component to get data to build MaxSubscribersValidationForm

diegoamatte:
Jira ticket: [DAT-1014](https://makingsense.atlassian.net/browse/DAT-1014)
Related to: [DAT-999](https://makingsense.atlassian.net/browse/DAT-999) and [DAT-774](https://makingsense.atlassian.net/browse/DAT-774)

This PR is a previous task to build this form that already exists in doppler, but not in the webapp.
subscribers-form

The intention is to create a component to get the form questions from the backend.


src/services/doppler-legacy-client.ts
```diff
@@ -1012,4 +1067,12 @@ export class HttpDopplerLegacyClient implements DopplerLegacyClient {
     };
     // return { success: false };
   }
+  public async getMaxSubscribersData(): Promise<MaxSubscribersData> {
+    const response = await this.axios.get('/sendmaxsubscribersemail/getmaxsubscribersdata');
+    if (!response?.data) {
+      throw new Error('Empty Doppler response');
+    }
```
andresmoschini, line 1074: Is it an expected scenario?

diegoamatte: I think that is not a common scenario, but it could be posible since the questions are retrieved from the database. 
What do you think?

andresmoschini: Let's talk

PR #2028, opened 2022-07-11
fix: enable button when change the suscription type user

jcamposmk:
### **Enable buy button when change the suscription type user**
**Jira Ticket**: https://makingsense.atlassian.net/browse/DAT-1019

**Description**
When the user has a monthly subscription type and he wants a trimestral subscription type, the buy continue button is enabled.

**The Bug**
Grabación de pantalla 2022-07-11 a la(s) 8 32 14 a


jhoffman-ms:
👍

PR #2053, opened 2022-08-03
fix: fix error when the user has account type query param

jcamposmk:
### **Fix accountType**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-1054

**Before**: The account type query param is lost
Grabación de pantalla 2022-08-03 a la(s) 11 10 32 a (1)

**After**
Grabación de pantalla 2022-08-03 a la(s) 11 15 00 a


jhoffman-ms:
👍

PR #2067, opened 2022-08-11
chore: add clarification comment in getReferrerHostname

andresmoschini:


andresmoschini:
@MatiasRoy-Rodriguez thanks for the advice!

I think that it could help.

MatiasRoy-Rodriguez:
> @MatiasRoy-Rodriguez thanks for the advice!
> 
> I think that it could help.

Perfect 💪

jcamposmk:
👍

PR #2068, opened 2022-08-11
refactor: remove use of "/features" endpoint

leoslopez:
It's not necessary to use this endpoint anymore because now the features are assumed like true for all users.
[A ticket](https://makingsense.atlassian.net/browse/DOP-1097) was created and associated with the related code.

related to [DOP-1080](https://makingsense.atlassian.net/browse/DOP-1080)


andresmoschini:
👍

mmosquera (approved):

PR #2125, opened 2022-09-19
fix: fix cannot read null from null in signup page

jcamposmk:
### **Fix 'Cannot read property length from null' in Signup page**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-1146

**Description**

When **utmCookies** is null, the user cannot be registered.


andresmoschini:
👍

PR #2127, opened 2022-09-22
fix: use query params when the user doesn't accept cookies

jcamposmk:
### **Fix** the last click utms 
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-1161

**Description**
Use query params when the user doesn't accept cookies

example: 

- The user navigates to `/signup?utm_source=fromdoppler&utm_medium=email&utm_campaign=demoday-septiembre2022` 
- The user doesn's accept cookies
- The user creates an account
- The request sends the fields UTMSource:fromdoppler, UTMMedium:email, UTMCampaign:demoday-septiembre2022 and UTMCookies: null (because the user doesn't accept cookies)


jhoffman-ms:
👍

PR #2131, opened 2022-09-27
[DAT-1168] - Disabled the continue button when the user only change the renovation type in the update between plans

jhoffman-ms:
Disabled the continue button when the user only change the renovation type in the update between plans.

**Task:** [DAT-1168](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-1168)


jcamposmk:
👍

PR #2134, opened 2022-09-28
[DAT-1169] - Disabled the renovation type component when the current plan of user is the same that it selected

jhoffman-ms:
Disabled the renovation type component when the current plan of user is the same that it selected.

**Task:** [DAT-1169](https://makingsense.atlassian.net/jira/software/projects/DAT/boards/62?selectedIssue=DAT-1169)


jcamposmk:
👍

PR #2135, opened 2022-09-28
fix: initialize only selected style guide features

andresmoschini:
Hi team!

We have a conflict between some _Style Guide MFE_ behavior and _Menu MFE_ behavior.

With this change, we are removing the unwanted _Style Guide MFE_ behavior. See https://github.com/FromDoppler/doppler-style-guide/pull/155

Fixes: [DE-796](https://makingsense.atlassian.net/browse/DE-796)


Cromeror (commented):
🚀

PR #2141, opened 2022-10-11
fix: avoid to try to translate control panel URLs

andresmoschini:
It fixes these errors:

image

### DIFF
--- a/src/components/ControlPanel/ControlPanelBox/ControlPanelBox.js
+++ b/src/components/ControlPanel/ControlPanelBox/ControlPanelBox.js
@@ -45,7 +45,7 @@ export const ControlPanelBox = ({ box }) => {
             <S.Link
               target={box.targetBlank ? '_blank' : '_self'}
               className="dp-white"
-              href={_(box.linkUrl)}
+              href={box.linkUrl}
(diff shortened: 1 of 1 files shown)


andresmoschini:
@fgchaio @jcamposmk could you review this change?

fgchaio:
I think there are some links that requiere translation.

andresmoschini:
> I think there are some links that requiere translation.

image

I see...

But in that case, it is not the right way to do it.

Maybe we can:
* Translate it here in `control-panel-service.ts`
* Use two different properties, one for URLs as they are and another for keys
* Add all URLs to the language files

andresmoschini:
Closing this PR, I have no time to fix it, but we should.

Let me know if you need help fixing it. Thanks!

fgchaio:
I will create a ticket for this.

andresmoschini:
>     * Translate it here in `control-panel-service.ts`

I think that this is the easiest solution, but we do not have a translation service out of React context. But we could inject it into the call to `getControlPanelSections()`

image

Maybe it is not so elegant, but it is easy and fixes the issue.

PR #2142, opened 2022-10-11
Fix local environment behavior with dummy session

andresmoschini:
I do the same that `index.js`: 

https://github.com/FromDoppler/doppler-webapp/blob/master/src/index.js#L39-L41


jcamposmk:
👍

PR #2144, opened 2022-10-13
DAT-1193 - Fix Shopify page

jcamposmk:
### **Fix Shopify page**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-1193

**Before:**
Captura de Pantalla 2022-10-13 a la(s) 11 40 13 a  m

**After:**
Captura de Pantalla 2022-10-13 a la(s) 11 40 41 a  m

### DIFF
--- a/src/components/Integrations/Shopify/Shopify.test.js
+++ b/src/components/Integrations/Shopify/Shopify.test.js
@@ -1,9 +1,15 @@
 import React from 'react';
-import { render, cleanup, waitFor } from '@testing-library/react';
+import {
+  render,
+  cleanup,
+  waitFor,
+  screen,
+  waitForElementToBeRemoved,
+} from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
 import DopplerIntlProvider from '../../../i18n/DopplerIntlProvider.double-with-ids-as-values';
 import { AppServicesProvider } from '../../../services/pure-di';
-import Shopify from './Shopify';
+import Shopify, { FETCH_SHOPIFY_DATA_INTERVAL } from './Shopify';
 import { SubscriberListState } from '../../../services/shopify-client';
 
 const oneShop = [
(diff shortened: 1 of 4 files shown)


andresmoschini:
Why the icons are broken?

src/utils.test.js
```diff
@@ -161,73 +159,6 @@ describe('utils', () => {
...
-        }),
-      );
-      expect(counter).toBe(1);
-
-      jest.advanceTimersByTime(200);
-      // Assert
-      expect(counter).toBe(3);
-    });
-  });
-
```
andresmoschini, line 230: Is possible to keep the tests? If I am not wrong, the behavior should be the same as before, right?

jcamposmk:
> Why the icons are broken?

That image is in my local environment. Right now they look good. 
Captura de Pantalla 2022-10-13 a la(s) 12 31 32 p  m

src/utils.test.js
jcamposmk: Yes, the behavior should be the same as before. 
But, now I changed it to use a different `useInterval` located in `src/hooks/useInterval/index.js`. 
Practically it is a bit like the previous one, but it is a little more flexible.

andresmoschini: > Practically it is a bit like the previous one, but it is a little more flexible.

But we loose tests :(

jcamposmk: I'm going to upload a commit testing this functionality in the `Shopify Component`

src/components/Integrations/Shopify/Shopify.test.js
```diff
@@ -205,4 +215,44 @@ describe('Shopify Component', () => {
...
+    // because it's executed to start
+    expect(getShopifyDataMock).toHaveBeenCalledTimes(1);
+
+    jest.advanceTimersByTime(FETCH_SHOPIFY_DATA_INTERVAL);
+    expect(getShopifyDataMock).toHaveBeenCalledTimes(2);
+
+    jest.advanceTimersByTime(FETCH_SHOPIFY_DATA_INTERVAL);
+    expect(getShopifyDataMock).toHaveBeenCalledTimes(3);
+  });
 });
```
jcamposmk, line 258: This was the missed test @andresmoschini

jhoffman-ms:
👍

src/components/Integrations/Shopify/Shopify.test.js
andresmoschini: 👏👏👏

PR #2149, opened 2022-10-17
[DOI-739] Refactor Control Panel URL translation

fgchaio:
The console was logging a lot of warnings of translation ids that couldn't resolve.
image

This way we only translate the URLs that need it.

Related to: https://github.com/FromDoppler/doppler-webapp/pull/2141


src/services/control-panel-service.ts
```diff
@@ -514,7 +514,7 @@ export class ControlPanelService implements ControlPanelService {
             targetBlank: true,
           },
           {
-            linkUrl: `control_panel.external_integrations.dynamics_link_url`,
+            linkUrl: _('control_panel.external_integrations.dynamics_link_url'),
             imgSrc: dynamics_icon,
             imgAlt: 'control_panel.external_integrations.dynamics_title',
             iconName: 'control_panel.external_integrations.dynamics_title',
```
andresmoschini, line 520: These fields are also translatable, what do you think about translating all translatable things in `getControlPanelSections`?

andresmoschini:
```suggestion
            imgAlt: _('control_panel.external_integrations.dynamics_title'),
            iconName: _('control_panel.external_integrations.dynamics_title'),
```

PR #2157, opened 2022-10-25
Dat 1201 get content activation

jcamposmk:
### **Get activation content**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-1201

**Description**

- The user enters to `Signup` page
- `getBannerData` is executed. Return a new `contentActivation` property
- The user creates an account (`contentActivation` is sent)
- The user is redirected to `/signup/confirmation `
- If there is `contentActivation` then is used as content instead of content by default


ms-fcabrera:
👍

PR #2168, opened 2022-11-03
fix: use environment variables for the generation of the Doppler reports url

ms-fcabrera:
use environment variables for the generation of the Doppler reports url instead of the hardcoded url


jhoffman-ms:
👍, excellent @facundo1cabrera

PR #2169, opened 2022-11-04
fix: fix not a number ui error on the purchase summary component

ms-fcabrera:
When the purchase summary component was loading it showed: "NaN" on the total tag. Now when is loading it shows 0

Before:
image
After:
image


jhoffman-ms:
👍

PR #2180, opened 2022-11-18
[DAT-1133] - Integrate app with provisory token

jhoffman-ms:
Integrate app with provisional token. 

The idea about this functionality is that when the user is blocked. The user can access the update payment information page with a provisional token.


ms-fcabrera:
👍

jcamposmk:
👍

PR #2181, opened 2022-11-23
feat: hide SMS option when the user hasn't sms enabled

jcamposmk:
### **Hide SMS option when the user hasn't sms enabled**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-1248


jhoffman-ms:
👍

PR #2192, opened 2022-12-01
[DOI-820] Move facebook card to native integrations

fgchaio:
The Facebook card was on the "external integrations" section but since it's actually an app of our own we are moving it to "native integrations".
image

image


jcamposmk:
👍

PR #2201, opened 2022-12-07
fix: use getInvoices and remove deprecated version

ms-fcabrera:
Use getInvoices instead of getDeclinedInvoices from the billing-user-api service


jcamposmk:
👍

PR #2208, opened 2022-12-13
[DOI-831] Add spinning icon when waiting for status fetch

fgchaio:
When fetching for the integrations status there is a delay between the rendering of the boxes and the re-ordering by status of them. This already happened in the control panel but since you had to scroll down to get to the integrations sections it wasn't visible.
If the request fails it should show the boxes without the connection icon.

**BEFORE:**
before

**AFTER**
after

### DIFF
--- a/src/components/Integrations/index.js
+++ b/src/components/Integrations/index.js
@@ -42,7 +42,7 @@ export const IntegrationsSection = InjectAppServices(
   ({ dependencies: { controlPanelService, dopplerUserApiClient } }) => {
     const intl = useIntl();
     const _ = useCallback((id, values) => intl.formatMessage({ id }, values), [intl]);
-    const [{ integrationSection, loading, loadingStatus }, dispatch] = useReducer(
+    const [{ integrationSection, loadingPage, loadingNativeIntegrations }, dispatch] = useReducer(
       IntegrationReducer,
       INITIAL_STATE,
     );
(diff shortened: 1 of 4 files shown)


src/components/Integrations/index.js
```diff
@@ -42,7 +42,7 @@ export const IntegrationsSection = InjectAppServices(
   ({ dependencies: { controlPanelService, dopplerUserApiClient } }) => {
     const intl = useIntl();
     const _ = useCallback((id, values) => intl.formatMessage({ id }, values), [intl]);
-    const [{ integrationSection, loading }, dispatch] = useReducer(
+    const [{ integrationSection, loading, loadingStatus }, dispatch] = useReducer(
```
andresmoschini, line 45: I do not understand the name `loadingStatus`, we already have a `loading` _status_, do not we?

Is it a kind of error?

fgchaio: hmm yes, we do have a loading already but that is beeing used for the page loading. 
This new loading I'm using it for the native integrations only, since there is a delay between the rendering and the boxes re-ordering.
Maybe there is a better way to handle it? or the name might be confusing?

andresmoschini: Maybe renaming the properties as `loadingPage` and `loadingNativeIntegrations`?

fgchaio: I like it! will do...

PR #2215, opened 2022-12-22
Dat 1306 add automatic scroll to error

jcamposmk:
### **Add automatic scroll and focus**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-1306

**Description**
Add automatic scroll and focus to first field with error in Signup form

* **BEFORE:**

Grabación de pantalla 2022-12-22 a la(s) 11 33 15 a

* **AFTER:**

Grabación de pantalla 2022-12-22 a la(s) 11 26 27 a


src/components/shared/ScrollToFieldError/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+    const element = document.querySelector(`input[name='${firstFieldErrorName}']`);
+    if (!element) {
+      return;
+    }
+
+    // unfocus the active element (in mobile, close keyboard)
+    if (document.activeElement) {
+      document.activeElement.blur();
+    }
```
andresmoschini, line 32: Thanks for the clarification comment!

src/components/shared/ScrollToFieldError/index.js
```diff
@@ -0,0 +1,47 @@
...
+    // Scroll to first known error into view
+    setTimeout(() => {
+      window.scroll({
+        top: element.getBoundingClientRect().top + window.scrollY - 30,
+        ...scrollBehavior,
+      });
+    }, 150);
+    setTimeout(() => {
+      element.focus();
+    }, 500);
```
andresmoschini, line 43: Why do you scroll first and then set focus?

And why focus is delayed?

src/components/shared/ScrollToFieldError/index.js
```diff
@@ -0,0 +1,47 @@
...
+  }, [submitCount]); // eslint-disable-line react-hooks/exhaustive-deps
```
andresmoschini, line 44: Do you really need to disable exhaustive-deps validation? is not possible to add the right dependencies?

src/components/shared/ScrollToFieldError/index.js
```diff
@@ -0,0 +1,47 @@
+import { useEffect } from 'react';
+
+export const ScrollToFieldError = ({
+  scrollBehavior = { left: 0, behavior: 'smooth' },
+  useFormikContext,
```
andresmoschini, line 5: Why do you need to inject `useFormikContext`? Can you not simply use the hook?

src/components/shared/ScrollToFieldError/index.js
```diff
@@ -0,0 +1,47 @@
...
+      fieldErrorNames.includes(fieldName),
+    );
+    if (!firstFieldErrorName) {
+      return;
+    }
+
+    const element = document.querySelector(`input[name='${firstFieldErrorName}']`);
+    if (!element) {
+      return;
+    }
```
andresmoschini, line 27: I am not sure if it is better or not, but I think that you could identify the first error from the DOM in place of using `fieldsOrder`? Am I missing something?

src/components/Signup/Signup.js
```diff
@@ -223,6 +225,10 @@ const Signup = function ({
             onSubmit={onSubmit}
             validate={validate}
           >
+            <ScrollToFieldError
+              useFormikContext={useFormikContext}
+              fieldsOrder={Object.values(fieldNames)}
+            />
```
andresmoschini, line 231: Nice API!

src/components/shared/ScrollToFieldError/index.js
jcamposmk: Yes, I need it. Because otherwise it would give warnings in console.

src/components/shared/ScrollToFieldError/index.js
jcamposmk: makes the tests much easier ...

src/components/shared/ScrollToFieldError/index.js
jcamposmk: I had not thought of that, but it's a good idea. I'm going to do that

src/components/shared/ScrollToFieldError/index.js
andresmoschini: Great! thanks for the explanation ❤️

src/components/shared/ScrollToFieldError/index.js
andresmoschini: Oh, I understand, that is the problem with the hooks!

maybe using a default value to simplify the code in the client component?

src/components/shared/ScrollToFieldError/index.js
andresmoschini: You will not get warnings if you add all the dependencies to the array, right?

ms-fcabrera:
👍

src/components/shared/ScrollToFieldError/index.js
jcamposmk: Yes, if I add all the dependencies I will not get warnings. But I just need that the useEffect it execute every time the user clicks the submit, for that I just add `submitCount` as dependency

src/components/shared/ScrollToFieldError/index.js
jcamposmk: yes, I'm agree. I'm going to do that ... thanks

src/components/shared/ScrollToFieldError/index.js
andresmoschini: Just for the record:

We have talked about it, the `errors` array changes more often than the `submitCount` and we do not want to run this code each time that the `errors` array has a change.

In my opinion, we should search for a way to include the dependency, for example with a _ref_, in order to make this code less error-prone. But @jcamposmk alternative has the advantage of simplicity: just adding the comment `// eslint-disable-line react-hooks/exhaustive-deps` is enough.

andresmoschini:
Great work @jcamposmk!

PR #2221, opened 2023-01-04
[DOI-840] Change breadcrumb links for Shopify and BigQuery

fgchaio:
Before:
image

After:
image

This pages aren't included in the control panel anymore. Users can access them through the integrations page.


src/components/Integrations/Shopify/Shopify.js
```diff
@@ -75,7 +75,7 @@ const FooterBox = ({ children }) => {
 
   return (
     <footer className="dp-integration__actions">
-      <a href={_('common.control_panel_url')} className="dp-button button-medium primary-grey">
+      <a href="/Integrations#native-integrations" className="dp-button button-medium primary-grey">
```
andresmoschini, line 78: If it is a link to the own webapp, you have to use `<Link>` component to avoid a _real navigation_. It also works better with the base URL.

fgchaio: cool, thanks!

andresmoschini: BTW, maybe you can use lowercase `integrations` in place of `Integrations` (we are trying to have more modern URLs now 😛)

fgchaio: ops! you are right, that should be lower case.

fgchaio: I was thinking to use the `<breadcrumb>` component instead of `<Link>`.
Also we could change the a tag in the [breadcrumb component](https://github.com/FromDoppler/doppler-webapp/blob/master/src/components/shared/Breadcrumb/Breadcrumb.js) for `<Link>`. I would do it in another PR since the scope will be much bigger.
 What do you think?

andresmoschini: Mmm, the component `<Breadcrumb>` is a little poor. In my opinion, reusing it as it is we lose more than we win.

Maybe we could improve it with a kind of _intelligent component_ that decides if using an `<a>` element or a `<Link>` based on the destination URL.

But I do have not a strong opinion. 

@jcamposmk what do you think?

jcamposmk: It could be, but I do have not a strong opinion, too.

fgchaio: switcheo a español para no perder ningun detalle.

Me queda una duda, si la redireccion es fuera de webapp el componente link tiene peor performance que un tag a?
Si es asi iria por la sugerencia de Andres, pero si no hay diferencia podriamos usar `<Link>` para todo?
Si dije cualquier cosa me dicen jeje.

andresmoschini: > Me queda una duda, si la redireccion es fuera de webapp el componente link tiene peor performance que un tag a?

Mmm, habría que verificarlo, pero si es fuera de la webapp creo que el componente Link funcionaría mal.

Buscando...

andresmoschini: Mmm, [documentation](https://reactrouter.com/en/main/components/link) does not clarify external URLs.

andresmoschini: Pareciera que para URLs externas hay que usar un `<a>`: https://github.com/remix-run/react-router/issues/8317#issuecomment-974577787

fgchaio: Joya, gracias Andres!

fgchaio: Me cree esta tarea [DOI-842](https://makingsense.atlassian.net/browse/DOI-842) para ajustar lo que hablamos en esta conversacion.
Gracias a ambos!!

PR #2269, opened 2023-03-23
[DOI-894] Change Shopify UI to display array of lists

fgchaio:
Before:
image

After:
image

**DO NOT MERGE YET.** 
This has to be deployed at the same time as the shopify plugin to avoid errors on the UI.


src/services/shopify-client.test.ts
```diff
@@ -30,48 +30,26 @@ describe('HttpShopifyClient', () => {
     console.error = consoleError; // Restore console error logs
   });
 
-  it('should set connected shop without list correctly', async () => {
```
fgchaio, line 33: This doesn't apply anymore

jcamposmk:
👍

PR #2277, opened 2023-04-05
feat: add taxes-excluded class to li and delete span

jcamposmk:
### **add taxes-excluded class to li and delete span**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-1429


jhoffman-ms:
👍

PR #2384, opened 2023-08-23
Dop 1445 get restriction time values from doppler api

leoslopez:
Obtain values from the Doppler api to complete the restriction time section:
image

@mmosquera the endpoint to be requested is: `https://apisqa.fromdoppler.net/contact-policies/accounts/{email}/settings`, 
and the new fields to be added in the response should be:
```
timeRestriction: {
	timeSlotEnabled: boolean,
        hourFrom: number,
        hourTo: number,
        weekdaysEnabled: boolean,
},
```

### DIFF
--- a/src/services/doppler-contact-policy-api-client.ts
+++ b/src/services/doppler-contact-policy-api-client.ts
@@ -78,6 +78,15 @@ export class HttpDopplerContactPolicyApiClient implements DopplerContactPolicyAp
     }));
   }
 
+  private mapTimeRestriction(data: any): TimeRestriction {
+    return {
+      timeSlotEnabled: data.timeSlotEnabled,
+      hourFrom: data.hourFrom || 0,
+      hourTo: data.hourTo || 0,
+      weekdaysEnabled: data.weekdaysEnabled,
+    };
+  }
+
   async getAccountSettings(): Promise<ResultWithoutExpectedErrors<AccountSettings>> {
     try {
       const { email, jwtToken } = this.getDopplerContactPolicyApiConnectionData();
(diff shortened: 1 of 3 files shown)


src/services/doppler-contact-policy-api-client.ts
```diff
@@ -19,13 +19,21 @@ export interface AccountSettings {
...
+  timeRestriction: TimeRestriction | null;
 }
 
 export interface SubscriberList {
   id: number;
   name: string;
 }
 
+export interface TimeRestriction {
+  timeSlotEnabled: boolean | null;
```
andresmoschini, line 31: why it is nullable?

src/services/doppler-contact-policy-api-client.ts
```diff
@@ -19,13 +19,21 @@ export interface AccountSettings {
...
+  hourFrom: number | null;
+  hourTo: number | null;
+  weekdaysEnabled: boolean | null;
+}
```
andresmoschini, line 35: Another more restrictive option could be

```suggestion
export interface TimeRestriction {
  timeSlotEnabled: false;
  hourFrom: number | null;
  hourTo: number | null;
  weekdaysEnabled: boolean | null;
} | {
  timeSlotEnabled: true;
  hourFrom: number;
  hourTo: number;
  weekdaysEnabled: boolean;
}
```

src/services/doppler-contact-policy-api-client.ts
leoslopez: The only reason was to have consistency with the definition for the [`active` field](https://github.com/FromDoppler/doppler-webapp/blob/master/src/services/doppler-contact-policy-api-client.ts#L18) in the AccountSettings definition, which corresponds to the configuration for the shipments quantity section:

image

PR #2385, opened 2023-08-28
Dop 1445 add contact policy time restriction in control panel

leoslopez:
Add a new sub-section **Time Restriction** into the **Contact Policy** section.

It's missing the proper validations to avoid saving incorrect data (for example, the hours valid are between 0 and 23). 
Anyway, this section is not being displayed by default, in case you want to display/test this section, you can run the following on the console: `localStorage.setItem('dopplerExperimental', JSON.stringify({timeRestrictionEnabled: true}));`.

chrome_rHyUjwIBEt
**Note:** the notifications/errors will be displayed at the bottom of the page, but will be treated with the validations in a next PR.


src/components/ContactPolicy/ContactPolicy.js
```diff
@@ -31,12 +31,24 @@ const fieldNames = {
...
+    dependencies: {
+      dopplerUserApiClient,
+      dopplerContactPolicyApiClient,
+      experimentalFeatures,
+      appSessionRef,
+    },
+  }) => {
     const navigate = useNavigate();
     const [enabled, setEnabled] = useState(false);
+    const [timeRestrictionEnabled, setTimeRestrictionEnabled] = useState(false);
```
andresmoschini, line 51: Since `timeRestrictionEnabled` is static (localStorage is only read on the first rendering), I think that you can simplify it in this way:

```suggestion
    // TODO: remove once the whole "time restriction" feature is implemented
    const timeRestrictionEnabled = experimentalFeatures.getFeature('timeRestrictionEnabled');
```

And remove the code in the _useEffect block_.

src/components/ContactPolicy/ContactPolicy.test.js
```diff
@@ -131,6 +171,50 @@ describe('ContactPolicy component', () => {
...
+      name: 'contact_policy.time_restriction.hour_from_aria_label',
+    });
+    expect(inputHourFrom).toHaveValue(10);
+
+    // Set "Hour To" = 20
+    let inputHourTo = screen.getByRole('spinbutton', {
+      name: 'contact_policy.time_restriction.hour_to_aria_label',
+    });
+    await act(() => user.clear(inputHourTo));
+    await act(() => user.type(inputHourTo, '20'));
```
andresmoschini, line 201: Is the act's result _awaitable_?

src/i18n/en.js
```diff
@@ -416,11 +417,24 @@ You can select **up to 10 Lists** to exclude them from the limit of Emails. Cont
...
     success_msg: 'Changes have been saved.',
+    time_restriction: {
+      hour_from_aria_label: '(TBD)From Hour',
+      hour_to_aria_label: '(TBD)To Hour',
+      legend: '(TBD)Configure and set a range of shipments for your Campaigns. Choose when you want the shipment to be executed:',
+      time_slot_hour_from_label: '(TBD)From',
+      time_slot_hour_to_label: '(TBD)To',
+      time_slot_toggle_text: "(TBD)Only in a specific time slot considering your account's time zone.",
+      title: '(TBD)Time restriction',
+      weekdays_toggle_text: '(TBD)Only from Monday to Friday',
```
andresmoschini, line 436: I would simply define acceptable content and use it (without the _TBD_). 

There is going to be time to improve it before the final release.

src/services/doppler-contact-policy-api-client.ts
```diff
@@ -107,7 +107,7 @@ export class HttpDopplerContactPolicyApiClient implements DopplerContactPolicyAp
             : [],
           timeRestriction: response.data.timeRestriction
             ? this.mapTimeRestriction(response.data.timeRestriction)
-            : null,
+            : this.mapTimeRestriction({}),
```
andresmoschini, line 110: Almost the same, but, in my opinion, it is simpler and makes it easier to understand why you are using an empty object.

```suggestion
          timeRestriction: this.mapTimeRestriction(response.data.timeRestriction ?? {}),
```

One alternative could be to accept falsy values as the parameter of `mapTimeRestriction`

src/components/ContactPolicy/ContactPolicy.test.js
leoslopez: Mm, I am not sure, but this is being treated in this way in all tests.

src/services/doppler-contact-policy-api-client.ts
leoslopez: Yes, I am going to fix it.

src/components/ContactPolicy/ContactPolicy.js
leoslopez: Done: https://github.com/FromDoppler/doppler-webapp/pull/2385/commits/4f9b8b0f61faba08b267cf68b5c258bf43e9af8f

src/services/doppler-contact-policy-api-client.ts
leoslopez: Done: https://github.com/FromDoppler/doppler-webapp/pull/2385/commits/ef6304825bd264cada7c8f5fb327154a7e97c694

PR #2395, opened 2023-09-07
feat: create use payment method data hook

jcamposmk:
### **Create use payment method data hook**
**Jira Ticket / Subtask:** https://makingsense.atlassian.net/browse/DAT-1687


jhoffman-ms:
👍

PR #2402, opened 2023-09-18
fix: add redirect when the user is not a free user

jcamposmk:
### **Add redirect when the user is not a free user**
**Jira Ticket:** https://makingsense.atlassian.net/browse/DAT-1707


jhoffman-ms:
👍

